PR #13, opened 2022-09-12
Add docs

mgoulao:
# Description

Add the documentation folder. I will add the workflow to deploy the docs website to gh-pages, and the versioning in a separate PR.


.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
     hooks:
       - id: codespell
         args:
-          - --ignore-words-list=nd,reacher,thist,ths, ure, referenc
+          - --ignore-words-list=nd,reacher,thist,ths, ure, referenc,Wile
```
pseudo-rnd-thoughts, line 12: I believe this needs to be lower case

pseudo-rnd-thoughts (changes requested):

pseudo-rnd-thoughts:
Do we need to modify the build process to not include the docs?

.pre-commit-config.yaml
mgoulao: ok, that makes sense

mgoulao:
> Do we need to modify the build process to not include the docs?

I don't think so, both `tests` and `docs` are ignored.

pseudo-rnd-thoughts:
Do we want to add some pre-commit for markdown? https://github.com/markdownlint/markdownlint

mgoulao:
> Do we want to add some pre-commit for markdown? https://github.com/markdownlint/markdownlint

We can, but I think it's better to leave that to a future PR.

pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts (approved):

PR #22, opened 2022-09-17
New wrappers that support complex observation / action spaces using jumpy

gianlucadecola:
# Description

This PR aims at adding new dev wrappers which use `jumpy` to allow high performance on higly parallelized environments (using Jax)
Compared to old wrappers they also fully support `Dict` and `Tuple` spaces and vectorized environments.
Given the huge amount of lines of code I will add soon to the PR description a short guide explaining the logic behind the new wrappers (some explanation is available on the `gym` PRs)

Code is already partially reviewed in the gym PRs:
https://github.com/openai/gym/pull/2985 
https://github.com/openai/gym/pull/3002
https://github.com/openai/gym/pull/3013

### Draft  
Opening in draft mode since I still need to check and possibly update all the docstring examples since they were written with the old API.

## TimeAwareObservationV0  
_ref_: https://github.com/openai/gym/pull/3013
TimeAwareObservationV0 is fully contained in:
`gymnasium/dev_wrappers/time_aware_observation.py`
`tests/dev_wrappers/test_time_aware_observation.py`  

By default, the wrapper returns a `Dict` observation space with the `obs` key being the original observation space and the `time` key being the remaining timesteps before the episode ends.
The wrapper accepts two parameters `flatten` and `normalize`.
Flattening returns a  `Box` observation space with time information at the end. 
Normalizing normalize time between [0, 1]  

## LambdaRewardsV0  
_ref_: https://github.com/openai/gym/pull/2985

## LambdaObservation and LambdaAction

_ref_: https://github.com/openai/gym/pull/3002


gymnasium/dev_wrappers/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Root __init__ of the gym dev_wrappers."""
+from typing import Dict, Sequence, TypeVar, Union
+
+ArgType = TypeVar("ArgType")
+FuncArgType = Union[ArgType, Dict[str, "FuncArgType"], Sequence["FuncArgType"]]
```
RedTachyon, line 5: Can you explain what is the purpose of this whole construction? My intuition is that it's way overengineered (or rather, too clever, which makes it harder to maintain), but maybe there's a valid reason for it. But it seems to support arbitrarily nested dicts/sequences of arbitrary objects?

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -0,0 +1,181 @@
...
+        """Constructor for the scale action wrapper.
+
+        Args:
+            env: The environment to wrap
+            args: The arguments for scaling the actions
+        """
+        action_space = self._transform_space(env, args)
+        args = self._make_scale_args(env, args)
+
+        def scale(action, args):
```
RedTachyon, line 167: I don't think we need to define the function inside of the method, it's better to keep it outside if it's even necessary

gymnasium/dev_wrappers/__init__.py
pseudo-rnd-thoughts: Yes, the plan is to support **all** observation / action spaces which includes Dict(Dict(...), ...) like structures. 
Therefore, we need to have functions that can run recursively over spaces, along with argument structures

RedTachyon:
I'll pull out the type hint discussion here because it seems to be at the root of my problems:
So my understanding of what this tries to do is: we specify an `ArgType` via a TypeVar (let's say, `bool`). Then `FuncArgType[bool]` would be either a `bool`, or a `Sequence[bool]`, or `Sequence[Sequence[bool]]` and so on (and the same with Dict interspersed). Essentially this leads to an arbitrary tree where all the leaves are `bool`

Here's a few comments/concerns (I will be omitting Dict for simplicity):
- From what I see, this is not what happens. Python understands `FuncArgType[bool]` as `Union[bool, Sequence[FuncArgType]`. Going one more recursive step, it seems to evolve into `Union[bool, Sequence[Union[Any, FuncArgType]]]` because the typevar is no longer specified. So in practice what it converges to is `Union[bool, Sequence[Any]]`, so either a bool or a sequence of anything at all.
- From what I understand, this is not what we want. In the code `FuncArgType` seems to denote optional arguments to the lambda function. The nesting can be argued to be useful if it's meant to correspond to the observations/actions (which can be arbitrarily nested). I'd still disagree, but this doesn't even seem to be the case
- In many places, the way that FuncArgType is used doesn't actually have anything specified, so it effectively means `Any`. Which might actually be a valid annotation, but is currently hidden under a pile of complexity


I will probably have a bunch more comments about the general PR later, but this should be the first thing to be fixed IMO, since the type system makes it extremely hard to understand. 
Some stubs of these future comments:
- The way these additional args are used seems inconsistent (there's `self.func_args` and there's passing it directly)
- There seems to be too much chained inheritance, or rather functions calling other functions (and not doing anything else). Or, in essence, just too much complexity. As an example, trace what `LambdaActionv0.action` does (it **should** be something roughly as complex as `return self.func(action)`)
- There's a few functions that take references to themselves as an argument??? (like `transform_space_bounds`



tl;dr I expect the functionality to be roughly correct, but it's waaay too clever right now. This should be an extremely simple piece of code, so we need to make it understandable enough to reflect that.

gianlucadecola:
Hey, thanks a lot for taking the time to review! So I think we can start to discuss on how to improve and make things easier to understand

> I'll pull out the type hint discussion here because it seems to be at the root of my problems: So my understanding of what this tries to do is: we specify an `ArgType` via a TypeVar (let's say, `bool`). Then `FuncArgType[bool]` would be either a `bool`, or a `Sequence[bool]`, or `Sequence[Sequence[bool]]` and so on (and the same with Dict interspersed). Essentially this leads to an arbitrary tree where all the leaves are `bool`
> 

Yeah that was actually the idea behind this type annotation

> Here's a few comments/concerns (I will be omitting Dict for simplicity):
> 
> * From what I see, this is not what happens. Python understands `FuncArgType[bool]` as `Union[bool, Sequence[FuncArgType]`. Going one more recursive step, it seems to evolve into `Union[bool, Sequence[Union[Any, FuncArgType]]]` because the typevar is no longer specified. So in practice what it converges to is `Union[bool, Sequence[Any]]`, so either a bool or a sequence of anything at all.

So, I'm trying to wrap my head around this, maybe a more correct type annotation could be this, but it's probably complicating things more

```python
ArgType = TypeVar("ArgType")
FuncArgType = Union[ArgType, Dict[str, "FuncArgType[ArgType]"], Sequence["FuncArgType[ArgType]"]]
```

> * In many places, the way that FuncArgType is used doesn't actually have anything specified, so it effectively means `Any`. Which might actually be a valid annotation, but is currently hidden under a pile of complexity

So looking back at the code I think there's some inconsistency in how I've used type annotation for function arguments; I'm going to double-check it.

> * From what I understand, this is not what we want. In the code `FuncArgType` seems to denote optional arguments to the lambda function. The nesting can be argued to be useful if it's meant to correspond to the observations/actions (which can be arbitrarily nested). I'd still disagree, but this doesn't even seem to be the case

Yes they denote optional arguments; they typically correspond to the observation space but not on a 1:1 basis, suppose we have this:

```python
env = ExampleEnv(
    action_space=Dict(
        body=Dict(head=Box(0.0, 10.0, (1,))), 
        left_arm=Discrete(4), 
        right_arm=Box(0.0, 5.0, (1,))
        )
    )
```

args could be in the form of:

```python
args = {"right_arm": (0, 2), "body": {"head": (0, 3)}}
```

So here for example they map to the corresponding action space but they are also missing the `left_arm` key.

> I will probably have a bunch more comments about the general PR later, but this should be the first thing to be fixed IMO, since the type system makes it extremely hard to understand. 

Yes, let's begin by working on one thing at a time, I think it's better

So these are my 2 cents, Besides the fact that there are probably (let's say 100%) mistakes in how I've used the type annotation, I'm fully open to work on that and get suggestions

pseudo-rnd-thoughts:
This is most likely too complex to add to this PR but would it be possible to add a `jit: bool = False` to these wrapper that will let you jit accelerate all of the functions are the observations are changeable but the arguments are not

setup.py
```diff
@@ -66,7 +64,8 @@
         "cloudpickle >= 1.2.0",
         "importlib_metadata >= 4.8.0; python_version < '3.10'",
         "gymnasium_notices >= 0.0.1",
-        "dataclasses == 0.8; python_version == '3.6'",
+        "brax-jumpy>=0.1.0",
```
pseudo-rnd-thoughts, line 67: We have updated the project name to jax-jumpy, https://pypi.org/project/jax-jumpy/

gianlucadecola: fixed :+1:

pseudo-rnd-thoughts:
@gianlucadecola This is a highly complex PR, what is the status of it? 
Are all of the lambda based wrappers done in your opinion? Do they just need reviewing?

To me, there are a couple of missing things that must go in separate PRs
1. Add non-lambda wrappers, frame stack, etc
2. Add website documentation, with the new wrappers page this should be minimal work
3. Deprecation warnings on all of the old wrappers with their respective new dev wrapper
4. Update all non-updated wrappers to their underscore version to have a similar implementation. 
5. Personally, I would like to consolidate the old wrappers, such that the compatibility wrappers are in the same file, render wrappers in the same file, etc, however, this might not be the right idea.

gianlucadecola:
> @gianlucadecola This is a highly complex PR, what is the status of it? Are all of the lambda based wrappers done in your opinion? Do they just need reviewing?

So I think we should start reviewing it, lambdaRewards, lambdaObservations, and lambdaActions are all in place as they were in the previous gym PR so I think we should work on reviewing them.  

Maybe you want me to split them into three separate PRs as they were on gym to make review easier? (since TimeAwareObsV0 was already reviewed and LambdaRewards is a lot easier than LambdaObservation and LambdaAction)


> To me, there are a couple of missing things that must go in separate PRs
> 
> 1. Add non-lambda wrappers, frame stack, etc

Yes they are missing and I agree they should go in a separate PR since some function defined in this PR could be easily reused for other wrappers

> 2. Add website documentation, with the new wrappers page this should be minimal work
> 3. Deprecation warnings on all of the old wrappers with their respective new dev wrapper
> 4. Update all non-updated wrappers to their underscore version to have a similar implementation.
> 5. Personally, I would like to consolidate the old wrappers, such that the compatibility wrappers are in the same file, render wrappers in the same file, etc, however, this might not be the right idea.

No problem with these other points :+1: , they require a little more work and we can discuss on that

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -0,0 +1,181 @@
...
+        >>> env.action_space
+        Dict(body: Dict(head: Box(0.0, 10.0, (1,), float32)), left_arm: Discrete(4), right_arm: Box(0.0, 5.0, (1,), float32))
+        >>> args = {"right_arm": (0, 2), "body": {"head": (0, 3)}}
+        >>> env = ClipActionsV0(env, args)
+        >>> env.action_space
+        Dict(body: Dict(head: Box(0.0, 3.0, (1,), float32)), left_arm: Discrete(4), right_arm: Box(0.0, 2.0, (1,), float32))
+    """
+
+    def __init__(
+        self, env: gymnasium.Env, args: FuncArgType[typing.Tuple[float, float]]
```
DavidSlayback, line 118: Could we clip with jp.ndarrays (e.g., different min/max for each element of Box space)?

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -0,0 +1,181 @@
...
+        >>> env = ExampleEnv(
+        ...    action_space=Dict(left_arm=Box(-2, 2, (1,)), right_arm=Box(-2, 2, (1,))
+        ... )
+        >>> env = ScaleActionsV0(env, {"left_arm": (-1,1), "right_arm": (-1,1)})
+        >>> env.action_space
+        Dict(left_arm: Box(-1, 1, (1,), float32), right_arm: Box(-1, 1, (1,), float32))
+    """
+
+    def __init__(
+        self, env: gymnasium.Env, args: FuncArgType[typing.Tuple[float, float]]
```
DavidSlayback, line 156: Same as the clipping, potentially array-based

gymnasium/dev_wrappers/lambda_reward.py
```diff
@@ -0,0 +1,97 @@
...
+            env (Env): The environment to apply the wrapper
+            min_reward (Union[float, jp.ndarray]): lower bound to apply
+            max_reward (Union[float, jp.ndarray]): higher bound to apply
+        """
+        if min_reward is None and max_reward is None:
+            raise InvalidBound("Both `min_reward` and `max_reward` cannot be None")
+
+        elif max_reward is not None and min_reward is not None:
+            array_bounds = isinstance(
+                min_reward, (np.ndarray, jnp.ndarray)
```
DavidSlayback, line 83: Minor nitpick: jp.ndarray (or chex.Array) instead of unions, since we're already leveraging jumpy

gymnasium/dev_wrappers/lambda_reward.py
```diff
@@ -0,0 +1,97 @@
...
+        Args:
+            env (Env): The environment to apply the wrapper
+            min_reward (Union[float, jp.ndarray]): lower bound to apply
+            max_reward (Union[float, jp.ndarray]): higher bound to apply
+        """
+        if min_reward is None and max_reward is None:
+            raise InvalidBound("Both `min_reward` and `max_reward` cannot be None")
+
+        elif max_reward is not None and min_reward is not None:
+            array_bounds = isinstance(
```
DavidSlayback, line 82: Could we replace all this with `invalid_bounds = jp.any(max_reward - min_reward < 0)`

gymnasium/dev_wrappers/time_aware_observation.py
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            action: The action to take
+
+        Returns:
+            The environment's step using the action.
+        """
+        self.timesteps += 1
+        observation, reward, terminated, truncated, info = super().step(action)
+
+        self.timesteps = jp.where(terminated, 0, self.timesteps)
```
DavidSlayback, line 100: I'm assuming it's not `terminated | truncated` so that the returned observation uses the final timesteps value?

gymnasium/dev_wrappers/time_aware_observation.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @staticmethod
+    def _get_max_timesteps(env: gymnasium.Env):
+        """Get the max episode steps before truncation occurs.
+
+        Args:
+            env: the provided environment
+        """
+        return (
+            env.get_attr("_max_episode_steps")[0]
```
DavidSlayback, line 137: Reminder for future PRs: Do we want different environments to have different time limits? If so, we probably need to change this logic.

gymnasium/dev_wrappers/time_aware_observation.py
```diff
@@ -0,0 +1,140 @@
...
+        self.time_aware_observation_space = Dict(
+            obs=env.observation_space, time=Box(0, self.max_timesteps, (self.num_envs,))
+        )
+
+        if self.flatten:
+            self.observation_space = spaces.flatten_space(
+                self.time_aware_observation_space
+            )
+        else:
+            self.observation_space = self.time_aware_observation_space
```
DavidSlayback, line 65: Could we define the observation and _get_time_observation functions within init to avoid using the control flow every timestep?

gymnasium/dev_wrappers/time_aware_observation.py
```diff
@@ -0,0 +1,140 @@
...
+                otherwise return time as remaining timesteps before truncation
+        """
+        super().__init__(env)
+        self.flatten = flatten
+        self.normalize_time = normalize_time
+        self.max_timesteps = self._get_max_timesteps(env)
+        self.num_envs = getattr(env, "num_envs", 1)
+
+        self.time_aware_observation_space = Dict(
+            obs=env.observation_space, time=Box(0, self.max_timesteps, (self.num_envs,))
```
DavidSlayback, line 57: Box(0, 1) if normalize_time

gymnasium/dev_wrappers/utils/grayscale_space.py
```diff
@@ -0,0 +1,48 @@
...
+    if args:
+        warnings.warn(
+            f"Grayscale conversion has no effect on spaces of type {type(space)}."
+        )
+    return space
+
+
+@grayscale_space.register(Box)
+def _grayscale_space_box(space, args: FuncArgType, fn: Callable):
+    if len(space.shape) != 3 and space.shape[-1] != 3:
```
DavidSlayback, line 37: Future PR may want to allow underlying environment to return a stack (NxWxHxC)

gymnasium/dev_wrappers/utils/reshape_space.py
```diff
@@ -0,0 +1,43 @@
...
+    Trying to reshape `Discrete`, `Multibinary` and `MultiDiscrete`
+    spaces has no effect.
+    """
+    if args:
+        raise InvalidSpaceOperation(f"Cannot reshape a space of type {type(space)}.")
+    return space
+
+
+@reshape_space.register(Box)
+def _reshape_space_box(space, args: FuncArgType[typing.Tuple[int, int]], fn: Callable):
```
DavidSlayback, line 34: `typing.Tuple[int,...]` for x-D Box shapes

gymnasium/dev_wrappers/utils/transform_space_bounds.py
```diff
@@ -0,0 +1,41 @@
...
+@transform_space_bounds.register(MultiBinary)
+@transform_space_bounds.register(MultiDiscrete)
+def _transform_space_discrete(
+    space, args: FuncArgType[typing.Tuple[int, int]], fn: Callable
+):
+    return space
+
+
+@transform_space_bounds.register(Box)
+def _transform_space_box(
```
DavidSlayback, line 28: `typing.Tuple[Union[int, jp.ndarray], ...]`

DavidSlayback (commented):
Overall looks good to me! Mostly minor suggestions. There are a few ideas for future PRs and performance tweaks to some of the utility functions, but no need to complicate this PR with them

gymnasium/spaces/utils.py
```diff
@@ -4,13 +4,17 @@
  to facilitate their usage in learning code.
 """
 import operator as op
+import typing
 from collections import OrderedDict
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: If you merge with main then run pre-commit this will run an automatic script for making all of these updates

gymnasium/dev_wrappers/lambda_action.py
gianlucadecola: Actually yes, if you mean this:

```python
>>> from gym.wrappers import ClipActionsV0
>>> from tests.dev_wrappers.utils import TestingEnv
>>> from gym.spaces import Box
i>>> mport numpy as np

>>> env = TestingEnv(action_space=Box(0, 5, (3,)))
>>> env.action_space
Box(0.0, 5.0, (3,), float32)

>>> wrapped_env = ClipActionsV0(env, (np.array([0,0,0]), np.array([2,3,1])))
>>> wrapped_env.action_space
Box(0.0, [2. 3. 1.], (3,), float32)

>>> wrapped_env.step(np.array([10,10,10]))
(array(-6.842204, dtype=float32), 0, False, False, {'action': array([2, 3, 1])})
```

gianlucadecola: There is still an open discussion about type hints, but considering also this way of clipping typing should be changed

gymnasium/dev_wrappers/lambda_action.py
gianlucadecola: Same as above

gymnasium/dev_wrappers/lambda_reward.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/7f98b19800369ad815e5e108454e52999cdd93c8

gymnasium/dev_wrappers/lambda_reward.py
gianlucadecola: Didn't thought about that!
https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/7b143cf40e1fe4fb32b22fa61657e21d99497667

gymnasium/dev_wrappers/time_aware_observation.py
gianlucadecola: Updated in https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/5202bf8b7c9e133b72a7eb2bbea8dea7b64ecb7a
This will avoid checking the control flow every time but maybe at the expense of some readability :thinking: are we good with that?

gianlucadecola: refactored a little bit in https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/e57935fe5bac47e51188e96b516eb0c110dd8d5c

gymnasium/dev_wrappers/time_aware_observation.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/e57935fe5bac47e51188e96b516eb0c110dd8d5c

gymnasium/dev_wrappers/utils/reshape_space.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/de9e43acdbe40a85cfdf6f0a39dce7a1326b1a03

gymnasium/dev_wrappers/utils/transform_space_bounds.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/22/commits/de9e43acdbe40a85cfdf6f0a39dce7a1326b1a03

RedTachyon:
Is this meant to be ready for a next review? I'm still seeing the whole `FuncArgType` thing, I don't have the time to look at it in detail right now, but it was supposed to be fixed/removed, no?

pseudo-rnd-thoughts (commented):
I think we are missing the following wrappers, Im happy to implement some 
- `FrameStackV0`
- `NormalizeRewardV0` - I think we should rename to `ScaleRewardV0` but need to check old implementations 
- `DelayObservationV0`
- `FrameSkipObervationV0`
- `StickyActionV0`
- `ActionNanV0`

PR #23, opened 2022-09-19
Update Docs with New Step API

arjun-kg:
# Description

Updates references of old step API (done) with new (terminated, truncated). Also added a `Handling Time Limits` section to explain the API change, and why time limits need to be correctly handled.

### DIFF
--- a/docs/content/basic_usage.md
+++ b/docs/content/basic_usage.md
@@ -37,10 +37,8 @@ to a specific point in space. If it succeeds in doing this (or makes some progre
 alongside the observation for this timestep. The reward may also be negative or 0, if the agent did not yet succeed (or did not make any progress). 
 The agent will then be trained to maximize the reward it accumulates over many timesteps.
 
-After some timesteps, the environment may enter a terminal state. For instance, the robot may have crashed! In that case,
-we want to reset the environment to a new initial state. The environment issues a done signal to the agent if it enters such a terminal state.
-Not all done signals must be triggered by a "catastrophic failure": Sometimes we also want to issue a done signal after
-a fixed number of timesteps, or if the agent has succeeded in completing some task in the environment.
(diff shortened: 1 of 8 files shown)


docs/content/basic_usage.md
```diff
@@ -73,6 +71,33 @@ the format of valid observations is specified by `env.observation_space`.
...
+`terminated` signal is set to `True` when the core environment terminates inherently because of task completion, failure etc. a condition defined in the MDP.   
+`truncated` signal is set to `True` when the episode ends specifically because of a time-limit or a condition not inherent to the environment (not defined in the MDP). 
+It is possible for `terminated=True` and `truncated=True` to occur at the same time when termination and truncation occur at the same step. 
+
+This is explained in detail in the `Handling Time Limits` section. 
+
+#### Backward compatibility
+Gym will retain support for the old API through compatibility wrappers. 
+
+Users can toggle the old API through `make` by setting `apply_api_compatibility=True`.
```
pseudo-rnd-thoughts, line 88: This section needs to be updated to `apply_env_compatibility`

docs/content/vectorising.md
```diff
@@ -139,7 +143,7 @@ Dict(fire:MultiDiscrete([2 2 2]), jump:MultiDiscrete([2 2 2]), acceleration:Box(
 ...     "jump": np.array([0, 1, 0]),
 ...     "acceleration": np.random.uniform(-1., 1., size=(3, 2))
 ... }
->>> observations, rewards, dones, infos = envs.step(actions)
+>>> observations, rewards, terminateds, truncateds, infos = envs.step(actions)
```
pseudo-rnd-thoughts, line 146: I think we should change `terminateds` and `truncateds` as this isn't good english to `terminations` and `truncations`. Repeat below

pseudo-rnd-thoughts (changes requested):
Other than the two requested changes, looks good to merge

docs/content/vectorising.md
arjun-kg: This is in line with "dones" which also isn't English. I think its common to have "name" + "s" to refer to array of "name"?

pseudo-rnd-thoughts: I agree that "dones" is not english but it is very close to it while "terminated" feels much more unenglish as it uses two suffixes (+ed and +s)

PR #25, opened 2022-09-24
Functional API and proof-of-concept jax classic-control envs

RedTachyon:
This is the gymnasium version of openai/gym#2958.

There was a lot of great ideas thrown around in various places, so we can try hashing them out here. But importantly, this is initially meant to be mainly for internal use, so we can do what makes sense for now, and later change it if necessary.

Some points to consider:
- Class name - FunctionalEnv or something like MarkovDecisionProcess
- What API to "officially" use - all methods being separate, or merging them into `reset` and `step`
- Which methods should accept RNG as an argument, and in some cases - what arguments should they accept
- - This needs extra attention to make sure `jax.jit` and `jax.vmap` will always work correctly


pseudo-rnd-thoughts:
My answer to the question you propose
1. The name should be `MarkovDecisionProcess` or `POMDP` as the class doesn't inherit from `gymnasium.Env`. Secondly, a minor point is that environment are currently called `CartpoleEnv` therefore, `FunctionalEnv` could be confused an environment called `Functional`
2. Originally I thought that separate `obs`, `reward`, `termination`, `truncation`, `info` functions was the best way however I now disagree and think we should have `initial` and `transition` as the state-based equivalent functions of `reset` and `step` then have `get_obs` (uncertain of the name) as the only other required functionality. 
```python
def initial(self, seed, options, rng) -> Tuple[StateType, ObsType, dict]
def transition(self, state, action, rng) -> Tuple[StateType, ObsType, float, bool, bool, dict]
def obs(self, state, rng) -> ObsType
```
Reasons for my change
- A. We would need to two info functions as the reset and step info might be different
- B. Function arguments, as discussed offline, the RL community uses a range of function definitions for reward, termination, 
etc depending on the system. To support all of these the reward would probably have to be `reward(self, state, action, next_state, rng) -> float`. This is ugly as a majority of environments wouldn't need to use half of the parameters. 
- C. Inefficient, as discussed with the function parameters, some environments will need to recompute (possibly complex) information done computing the next state. Therefore duplicating compute unnecessarily.
- D. Ease of change from `Env` to these new state-based environments. While, I agree that we should encourage users to separate their code out into their own reward, termination, etc functions. I think we shouldn't force people to do that and to make it easier to replicate code for these new environments, we should use the previous two function method.
3. My answer proposes type hints to question 2

balisujohn:
I support providing a class called `FunctionalEnv`, with core methods `functional_step`, and `functional_reset`, and `functional_obs`, otherwise exactly as specified by @pseudo-rnd-thoughts  above. I support also providing a class called `MarkovDecisionProcess`, with transition, reward, obs, initial as methods, and a conversion wrapper that can create an instance of `FunctionalEnv` from an instance of `MarkovDecisionProcess`.

Rationale:
I think it's good from a minimum complexity burden to users perspective to just have functional step and reset equivalents as a minimum requirement for interfacing with a conversion wrapper to the standard stateful gym API.  An issue is that it is misleading to call the functional step 'transition', because it is really a fusion of the transition and reward functions that make up an MDP. In total, this kind of pushes me towards using the name `FunctionalEnv`, and calling the functions functional_step, and functional_reset respectively. We could create a separate class called `MarkovDecisionProcess` which implements all the components of an MDP exactly as defined, and a conversion wrapper to create a `FunctionalEnv` from a `MarkovDecisionProcess`. This way, people can get the efficiency of just having to define functional step/reset to get a working Jax compatible env, and people can get the precision of being able to implement an MDP factorized exactly as mathematically defined quickly working with either the functional or stateful interface abstractions.

gymnasium/envs/phys2d/conversion.py
```diff
@@ -0,0 +1,77 @@
...
+
+import jax.random as jrng
+
+import gymnasium as gym
+from gymnasium import Space
+from gymnasium.envs.registration import EnvSpec
+from gymnasium.functional import ActType, FuncEnv, StateType
+
+
+class JaxEnv(gym.Env):
```
pseudo-rnd-thoughts, line 11: Add `close` function with `self.func_env.render_close()`

gymnasium/envs/phys2d/conversion.py
```diff
@@ -0,0 +1,77 @@
...
+        self.spec = spec
+
+        if self.render_mode == "rgb_array":
+            self.render_state = self.func_env.render_init()
+        else:
+            self.render_state = None
+
+    def reset(self, *, seed: Optional[int] = None, options: Optional[dict] = None):
+        super().reset(seed=seed)
+        if not hasattr(self, "rng") or seed is not None:
```
pseudo-rnd-thoughts, line 47: I think we should initialise "rng" in `__init__` with `seeding.np_random()` using the second value which is the random seed value. 
Then we only update seed when `seed is not None`

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def render_image(
+        self, state: StateType, render_state: RenderStateType
+    ) -> Tuple[RenderStateType, np.ndarray]:
+        """Show the state."""
+        raise NotImplementedError
+
+    def render_init(self, **kwargs) -> RenderStateType:
+        """Initialize the render state."""
+        raise NotImplementedError
```
pseudo-rnd-thoughts, line 86: Add `render_close_state`

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def observation(self, state: StateType) -> ObsType:
+        """Observation."""
+        raise NotImplementedError
+
+    def transition(self, state: StateType, action: ActType, rng: Any) -> StateType:
+        """Transition."""
+        raise NotImplementedError
+
+    def reward(
```
pseudo-rnd-thoughts, line 48: Optional change to `transition_reward` and add `rng` parameter?

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+
+    For the moment, this is predominantly for internal use. This API is likely to change, but in the future
+    we intend to flesh it out and officially expose it to end users.
+    """
+
+    def __init__(self, options: Optional[Dict[str, Any]] = None):
+        """Initialize the environment constants."""
+        self.__dict__.update(options or {})
+
+    def initial(self, rng: Any) -> StateType:
```
pseudo-rnd-thoughts, line 36: Optional change to `initial_state`

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+        """Initial state."""
+        raise NotImplementedError
+
+    def observation(self, state: StateType) -> ObsType:
```
pseudo-rnd-thoughts, line 40: Optional change to `state_observation`

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+        """Transition."""
+        raise NotImplementedError
+
+    def reward(
+        self, state: StateType, action: ActType, next_state: StateType
+    ) -> RewardType:
+        """Reward."""
+        raise NotImplementedError
+
+    def state_info(self, state: StateType) -> dict:
```
pseudo-rnd-thoughts, line 54: Optional change to `initial_state_info`

gymnasium/envs/phys2d/cartpole.py
```diff
@@ -0,0 +1,236 @@
...
+
+        return screen, clock
+
+
+class CartPoleJaxEnv(JaxEnv):
+    def __init__(self, render_mode: Optional[str] = None, **kwargs):
+        env = CartPoleF(**kwargs)
+        env.transform(jax.jit)
+        high = env.x_threshold
+        action_space = gym.spaces.Discrete(2)
```
pseudo-rnd-thoughts, line 227: Can this not be copied from the environment? And done automatically within JaxEnv if the parameters are None?

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+
+StateType = TypeVar("StateType")
+ActType = TypeVar("ActType")
+ObsType = TypeVar("ObsType")
+RewardType = TypeVar("RewardType")
+TerminalType = TypeVar("TerminalType")
+RenderStateType = TypeVar("RenderStateType")
+
+
+class FuncEnv(Generic[StateType, ObsType, ActType, RewardType, TerminalType]):
```
pseudo-rnd-thoughts, line 15: Add `RenderStateType` and add `abc.ABC`?

gymnasium/envs/phys2d/conversion.py
```diff
@@ -0,0 +1,77 @@
...
+        action_space: Space,
+        metadata: Optional[Dict[str, Any]] = None,
+        render_mode: Optional[str] = None,
+        reward_range: Tuple[float, float] = (-float("inf"), float("inf")),
+        spec: Optional[EnvSpec] = None,
+    ):
+        """Initialize the environment from a FuncEnv."""
+        if metadata is None:
+            metadata = {}
+        self.func_env = func_env
```
pseudo-rnd-thoughts, line 32: Should we automatically jit the functional environment? Probably with a parameter for `pmap` or `vmap`

pseudo-rnd-thoughts (changes requested):
Generally, I think this very close to have a ready experimental version

I have made a number of function name changes, these are suggestions to make the functions more descriptive and you can ignore them.

In the future, we should have an environment checker for jax functional environments to check they are correctly implemented. Using the chex.assert_max_compile function check that jitted functions are not being recompiled. I have done a bit of work on this.

Im still unsure of the function definitions and I think that `rng` should be added to `terminal`, `reward`, and `render_image`

Additionally, I think we should have some more tests, and check equivalence to the NumPy versions of the environments however these can be done later. 

Final comment, jax needs to be added setup (as a necessary install I believe) as all the tests are being ignored currently

gymnasium/envs/phys2d/conversion.py
RedTachyon: Added

gymnasium/functional.py
RedTachyon: This is the same as the first review comment, right?

gymnasium/envs/phys2d/conversion.py
RedTachyon: Added

gymnasium/functional.py
RedTachyon: I think I prefer the shorter names for now, we can reconsider it over time

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+    The class-based structure serves the purpose of allowing environment constants to be defined in the class,
+    and then using them by name in the code itself.
+
+    For the moment, this is predominantly for internal use. This API is likely to change, but in the future
+    we intend to flesh it out and officially expose it to end users.
+    """
+
+    def __init__(self, options: Optional[Dict[str, Any]] = None):
+        """Initialize the environment constants."""
+        self.__dict__.update(options or {})
```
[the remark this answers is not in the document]
RedTachyon: This is admittedly some "clever fuckery" going on. The idea here is that you can give any additional arguments, and those will then overwrite the "parameters" of the FuncEnv. So for example, in cartpole you can add `options={"gravity": 1, "x_init": 20}`. Then the instance of the environment will use those values instead of the default ones, allowing easy parametrization of the env. Something like we added in `reset` of the OOP API, but much more powerful

gymnasium/functional.py
```diff
@@ -0,0 +1,86 @@
...
+    def step_info(
+        self, state: StateType, action: ActType, next_state: StateType
+    ) -> dict:
+        """Info dict about a full transition."""
+        return {}
+
+    def terminal(self, state: StateType) -> TerminalType:
+        """Terminal state."""
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
RedTachyon: I'd keep it out for now, the state-dependent truncation in car racing is... weird. I think we should turn that into a normal termination more than anything else. We can add this if it's actually necessary, but just the idea of truncation depending on the (timeless) state seems very weird

gymnasium/envs/phys2d/cartpole.py
```diff
@@ -0,0 +1,236 @@
...
+        screen = pygame.Surface((screen_width, screen_height))
+        clock = pygame.time.Clock()
+
+        return screen, clock
+
+
+class CartPoleJaxEnv(JaxEnv):
+    def __init__(self, render_mode: Optional[str] = None, **kwargs):
+        env = CartPoleF(**kwargs)
+        env.transform(jax.jit)
```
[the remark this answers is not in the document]
RedTachyon: Not necessary (as checked in the tests). When you call the instantiated env's method `FuncEnv().initial(...)`, `.initial` doesn't actually take `self` as an argument

gymnasium/envs/phys2d/cartpole.py
RedTachyon: I'm not yet sure about putting it automatically in JaxEnv, there's a slightly tricky thing that the uninstantiated class doesn't necessarily have the spaces defined. But I did make it take the spaces directly from these two specific envs

gymnasium/envs/phys2d/conversion.py
RedTachyon: I'm not sure about automatic jitting, let's keep it out for now, it's one line per each environment. Later we can try make it automatic if everything works.

And especially vmap/pmap will have to wait for the vector API to be finalized

gymnasium/functional.py
RedTachyon: Added RenderStateType, I think I'll keep ABC out for now, I'd need to triple check if it wouldn't break anything (especially with jax stuff). It shouldn't, but...

RedTachyon:
@pseudo-rnd-thoughts This should be ready for a final-ish review. Some notable changes from the meantime:

- I changed one test which assumed that every env supports human rendering (only removed the assertion)
- I changed all the new envs to return numpy arrays as opposed to jax arrays. This is required by some tests, and we'll probably want to change that at some point if we want to allow envs which fully live in e.g. jax/torch
- I had to stop automatic jitting of OOP envs, because jitted envs can't be pickled, and some test requires every env to be picklable. Now that I think of it, it might be fixable with the whole EZPickle thing, but I'm not sure

pseudo-rnd-thoughts (approved):

PR #31, opened 2022-09-29
Faster EpisodeStatisticsRecorder, converted from gym PR

DavidSlayback:
# Description

(Re-PR for gymnasium, this was reviewed and approved on the old gym repo [PR #3101](https://github.com/openai/gym/pull/3101)) 

This addresses two issues I've run into with RecordEpisodeStatistics

1) Silent list conversion in vectorized environments. Currently, RecordEpisodeStatistics converts terminated and truncated into lists, but obs and reward may still be arrays, and the new vectorized `info` interface uses arrays. I particularly run into this in my code if I do something like `dones.any()`. This PR leaves the underlying details of terminated/truncated alone by using NumPy array functionality.

2) Speed. Instead of looping through a potentially large batch of environments, this PR uses NumPy functions with their underlying vectorization to achieve the same task.

I considered building this with Jumpy, but I'm not sure on its status in the project, and it would also eat the cost of creating new arrays wherever I'm doing in-place modifications. 

I tested to make sure that this replicates the exact functionality of the original wrapper. To prove my speedup, I need to come up with some decent micro benchmarks; a couple attempts on Colab were fairly noisy.

## Type of change

- [x] Existing feature enhancement (non-breaking change which adds functionality)

# Checklist:

- [N/A] I have commented my code, particularly in hard-to-understand areas
- [N/A] I have made corresponding changes to the documentation
- [N/A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gymnasium/wrappers/record_episode_statistics.py
+++ b/gymnasium/wrappers/record_episode_statistics.py
@@ -68,7 +68,7 @@ class RecordEpisodeStatistics(gym.Wrapper):
         """Resets the environment using kwargs and resets the episode returns and lengths."""
         observations = super().reset(**kwargs)
         self.episode_returns = np.zeros(self.num_envs, dtype=np.float32)
-        self.episode_lengths = np.zeros(self.num_envs, dtype=np.int32)
+        self.episode_lengths = np.zeros(self.num_envs, dtype=np.float32)
         return observations
 
     def step(self, action):
@@ -92,7 +92,7 @@ class RecordEpisodeStatistics(gym.Wrapper):
                 "episode": {
                     "r": np.where(dones, self.episode_returns, 0),
                     "l": np.where(dones, self.episode_lengths, 0),
-                    "t": round(time.perf_counter() - self.t0, 6),
+                    "t": np.full((self.num_envs,), round(time.perf_counter() - self.t0, 6), dtype=np.float32),
                 },
             }
             if self.is_vector_env:
--- a/gymnasium/wrappers/record_episode_statistics.py
+++ b/gymnasium/wrappers/record_episode_statistics.py
@@ -92,7 +92,9 @@ class RecordEpisodeStatistics(gym.Wrapper):
                 "episode": {
                     "r": np.where(dones, self.episode_returns, 0),
                     "l": np.where(dones, self.episode_lengths, 0),
-                    "t": np.full((self.num_envs,), round(time.perf_counter() - self.t0, 6)),
+                    "t": np.full(
+                        (self.num_envs,), round(time.perf_counter() - self.t0, 6)
+                    ),
                 },
             }
             if self.is_vector_env:
--- a/gymnasium/wrappers/record_episode_statistics.py
+++ b/gymnasium/wrappers/record_episode_statistics.py
@@ -23,7 +23,7 @@ class RecordEpisodeStatistics(gym.Wrapper):
         ...     "episode": {
         ...         "r": "<cumulative reward>",
         ...         "l": "<episode length>",
-        ...         "t": "<elapsed time since instantiation of wrapper>"
+        ...         "t": "<elapsed time since beginning of episode>"
         ...     },
         ... }
 
@@ -34,7 +34,7 @@ class RecordEpisodeStatistics(gym.Wrapper):
         ...     "episode": {
         ...         "r": "<array of cumulative reward>",
         ...         "l": "<array of episode length>",
-        ...         "t": "<array of elapsed time since instantiation of wrapper>"
+        ...         "t": "<array of elapsed time since beginning of episode>"
         ...     },
         ...     "_episode": "<boolean array of length num-envs>"
         ... }
@@ -56,8 +56,8 @@ class RecordEpisodeStatistics(gym.Wrapper):
         """
         super().__init__(env)
         self.num_envs = getattr(env, "num_envs", 1)
-        self.t0 = time.perf_counter()
         self.episode_count = 0
+        self.episode_start_times: np.ndarray = None
         self.episode_returns: Optional[np.ndarray] = None
         self.episode_lengths: Optional[np.ndarray] = None
         self.return_queue = deque(maxlen=deque_size)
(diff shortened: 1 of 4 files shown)


DavidSlayback:
Hmm...I'll take a look at the fail later today and fix it, it was a quick paste with a couple import changes.

DavidSlayback:
Fixed, needed to store time as a vector (which seems a bit silly, it's the same time for all environments in a batch) and use float64 arrays for everything (to pass vector_list_info)

DavidSlayback:
Weird, I never touched that file. Blacked it

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,30 @@ def step(self, action):
...
+        dones = truncateds | terminateds
+        num_dones = np.sum(dones)
+        if num_dones:
+            episode_info = {
+                "episode": {
+                    "r": np.where(dones, self.episode_returns, 0),
+                    "l": np.where(dones, self.episode_lengths, 0),
+                    "t": np.full(
+                        (self.num_envs,), round(time.perf_counter() - self.t0, 6)
+                    ),
```
gianlucadecola, line 97: Extremely minor, I think this could be the same as 
```python
np.full(self.num_envs, round(time.perf_counter() - self.t0, 6))
```
since self.num_envs is an integer it will be evaluated as the same shape without needing to wrap it in a tuple

gianlucadecola (approved):
Nice job! LGTM

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -96,8 +67,8 @@ def __init__(self, env: gym.Env, deque_size: int = 100):
     def reset(self, **kwargs):
         """Resets the environment using kwargs and resets the episode returns and lengths."""
         observations = super().reset(**kwargs)
-        self.episode_returns = np.zeros(self.num_envs, dtype=np.float32)
```
RedTachyon, line 99: Why this change? Pretty much all the floating point math in gymnasium works on float32, this change will either add an extra cast somewhere along the line, or require an explicit cast by user by breaking their code (numpy doesn't care, but pytorch does)

RedTachyon:
@pseudo-rnd-thoughts @gianlucadecola Just to be sure - did you go through the logic in both versions of the code to be sure that it's the same thing? It does seem cleaner which alone is sufficient reason for me to like this PR, but I'd need to sit down with it for a bit to make sure, and it might not be necessary. 

Also I see now that there's a message in the meantime about having to switch to float64 -- this is very weird and I'd be curious to see the reason for it, but I also hope that there's a different solution than using float64

gymnasium/wrappers/record_episode_statistics.py
DavidSlayback: Solely for passing the test for the VectorList wrapper, it checks each element with `assert isinstance(..., float)`, which only passed for float64 (not float32 or int). I agree I'd prefer to use float32, wasn't sure whether I could change pre-existing tests

RedTachyon: Can you point me to the specific test? File/line?

DavidSlayback: https://github.com/Farama-Foundation/Gymnasium/blob/9154d2db4e3955a99d81d1b1d982349cb78c80e6/tests/wrappers/test_vector_list_info.py#L56

RedTachyon: Is it the case that `list_info[i]["episode"][stats]` used to be a list of floats, and now is a numpy array? Or am I misunderstanding something here?

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,30 @@ def step(self, action):
...
-                self.episode_lengths[i] = 0
+        dones = truncateds | terminateds
+        num_dones = np.sum(dones)
+        if num_dones:
+            episode_info = {
+                "episode": {
+                    "r": np.where(dones, self.episode_returns, 0),
+                    "l": np.where(dones, self.episode_lengths, 0),
+                    "t": np.full(
+                        (self.num_envs,), round(time.perf_counter() - self.t0, 6)
```
pseudo-rnd-thoughts, line 96: I'm not sure this will matter much but should be use `np.round` rather than the python `round` function given the value gets passed directly to a numpy array.

Secondly, this "t" value doesn't seem much help as `t0` is never updated to my eyes. So "t" seems to be the number of milli (or nano) seconds since the wrapper was initialised. 

This would break backward compatibility but would we want to change "t" to be the number of milli (or nano) seconds that the episode has taken. This would require us to have a `perf_counter` for every sub-environment but this shouldn't be too bad

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,30 @@ def step(self, action):
...
+            self.length_queue.extend(self.episode_lengths[dones])
+            self.episode_count += num_dones
+            self.episode_lengths[dones] = 0
+            self.episode_returns[dones] = 0
         return (
             observations,
             rewards,
-            terminateds if self.is_vector_env else terminateds[0],
-            truncateds if self.is_vector_env else truncateds[0],
+            terminateds,
```
pseudo-rnd-thoughts, line 111: Could you update `terminateds` to `termnations` and `trucnateds` to `truncations` as this is better english

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,30 @@ def step(self, action):
...
-                        infos, episode_info["episode"], self.num_envs, i
-                    )
-                else:
-                    infos = {**infos, **episode_info}
-                self.return_queue.append(episode_return)
-                self.length_queue.append(episode_length)
-                self.episode_count += 1
-                self.episode_returns[i] = 0
-                self.episode_lengths[i] = 0
+        dones = truncateds | terminateds
```
pseudo-rnd-thoughts, line 88: For ease of reading, could you update to `dones = np.logical_or(termination, truncation)` as `|` is not clear what is means in this

gymnasium/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,30 @@ def step(self, action):
...
+                    "r": np.where(dones, self.episode_returns, 0),
+                    "l": np.where(dones, self.episode_lengths, 0),
+                    "t": np.full(
+                        (self.num_envs,), round(time.perf_counter() - self.t0, 6)
+                    ),
+                },
+            }
+            if self.is_vector_env:
+                episode_info["_episode"] = np.where(dones, True, False)
+            infos = {**infos, **episode_info}
```
pseudo-rnd-thoughts, line 102: Could we change this to
```python
if "episode" or "_episode" in infos:
    # raise warning
else:
    infos["episode"] = {...}
    if self.is_vector_env:
        infos["_episodes"] = ...
```
This seems better than the old way as it is a single element we are adding to the dictionary

pseudo-rnd-thoughts (changes requested):
@RedTachyon I believe that the implementation is the same. Could you check my suggestion on modifying "t" to be the episode length not time since initialisation. I suspect that very few people already use it so I think we can change it.
 
@DavidSlayback Looking at the code, given we are cleaning up the code, I have spotted a couple of other places that could be improved.

DavidSlayback:
Made the requested changes.
1) I changed the storage arrays to float32 and int32. I had to adjust the test for vector_list_info to use np.isscalar. Basically, python isn't a big fan of float32. Getting a single value from a float64 array returns a python float, but from a float32 array, you get a numpy scalar.

2) `t0` is renamed to `episode_start_times` and tracks the start time for each episode. Pyright didn't like it being an `Optional[np.ndarray]`, hence the hack in the `__init__`. Updated wrapper documentation to reflect the change

3) Changed some of the wording for readability

pseudo-rnd-thoughts (approved):
There will be another Gym release with some minor bug fixes, hopefully, released in the next few days. 
To keep Gymnasium equivalent to Gym until launch Im going to wait for merge this

Issue #11, opened 2022-09-11
[Bug] Allow Wrapper metadata to be different from environment metadata 


pseudo-rnd-thoughts:
The `HumanRendering` and `RenderCollection` wrappers enable additional `render_modes` which should be reflected in the `metadata`.
Currently, this is not true. 

I was originally thinking about of a complex method of overwriting the dict implementation such that the keys are linked to the environment. However, as the metadata is meant to be static, then I think we can just deepcopy the lower env's metadata and add the wrappers additional metadata

PR #35, opened 2022-10-03
Modify metadata in the human rendering wrapper

RedTachyon:
@pseudo-rnd-thoughts Does this solve #11? (as a proof-of-concept for now - I'll need to check other wrappers, and in some way handle duplicated render modes that might occur) I'm slightly confused because you were talking about overwriting the dict stuff on call, but I don't see any issues with this implementation (which you also seem to have proposed in the issue), so maybe I'm not seeing something.

### DIFF
--- a/gymnasium/wrappers/human_rendering.py
+++ b/gymnasium/wrappers/human_rendering.py
@@ -1,4 +1,6 @@
 """A wrapper that adds human-renering functionality to an environment."""
+import copy
+
 import numpy as np
 
 import gymnasium as gym
@@ -57,6 +59,9 @@ class HumanRendering(gym.Wrapper):
         self.screen_size = None
         self.window = None
         self.clock = None
+
+        self.metadata = copy.deepcopy(self.env.metadata)
+        self.metadata["render_modes"].append("human")
 
     @property
     def render_mode(self):


pseudo-rnd-thoughts:
Yes, Im not sure why I was thinking it would need to be more complex. 
We also need to update the `RenderCollection` wrapper, could you do that as well
I can't think of anymore wrappers that need to be updated

gymnasium/wrappers/render_collection.py
```diff
@@ -22,6 +24,10 @@ def __init__(self, env: gym.Env, pop_frames: bool = True, reset_clean: bool = Tr
         self.reset_clean = reset_clean
         self.pop_frames = pop_frames
 
+        self.metadata = copy.deepcopy(self.env.metadata)
+        if f"{self.env.render_mode}_list" not in self.metadata["render_modes"]:
```
pseudo-rnd-thoughts, line 28: Technically, we allow all render modes to be list versions except "human"
Could you modify it to add all valid list render modes

RedTachyon: I think I originally had this implemented, but then I realized that's not exactly the case - when you use this wrapper, it takes the current environment's render mode and adds "_list" to it. And in principle, you can't change the `render_mode` of an environment after it's initialized. So for example, if you have an env with `render_mode="ansi"` and use the wrapper, there is no practical notion in which it now has access to `render_mode="rgb_array_list"` unless you reinitialize the whole thing

Which also makes me realize now that updating the metadata doesn't actually make that much of a difference now that `render` doesn't take any arguments.

pseudo-rnd-thoughts: Good point, then I think this PR is good to merge

pseudo-rnd-thoughts (approved):
Im going to wait until gym v26.2 is released such that gymnasium is equivalent until launch

PR #37, opened 2022-10-06
Update the spaces for complete type hinting (and updates numpy to 1.21)

pseudo-rnd-thoughts:
A lot of the spaces in Gymnasium have partial type hints, are missing type hints or have incorrect type hints

This PR fixes and adds a number of type hints in Spaces using the np.NDArray type hints that allows the dtype of a numpy array to be specified. 

The aim of this PR is to enable the GeneralTypeHint in pyright for future PRs 

One of the requirements of this PR is `np.typing.NDArray` which was introduced in `1.21.0` to numpy.


pseudo-rnd-thoughts:
This waiting for the python 3.6 PR to be merged

requirements.txt
```diff
@@ -1,4 +1,4 @@
-numpy>=1.18.0
+numpy>=1.21.0
```
RedTachyon, line 1: This is a quiet but potentially important change. At the very least we will need to properly mention this in the changelog. And we definitely have to think about the potential implications, because if some other library on some version on some system as a requirement `numpy <= 1.20`, then everything breaks.

As I understand, this is the minimum version required for `np.typing.NDArray`?

pseudo-rnd-thoughts: Yes, you are correct. This is required for `np.typing.NDArray`.
I will add a comment on the PR explaining this

One of the limitations of < 1.21 is no python 3.6 version exists but if we are removing python 3.6 in the next release then this is fine. 
Do you know why 1.20 causes issues?

gymnasium/spaces/box.py
```diff
@@ -25,12 +27,12 @@ def _short_repr(arr: np.ndarray) -> str:
...
 
 
-def is_float_integer(var) -> bool:
+def is_float_integer(var: Any) -> bool:
     """Checks if a variable is an integer or float."""
     return np.issubdtype(type(var), np.integer) or np.issubdtype(type(var), np.floating)
 
 
-class Box(Space[np.ndarray]):
+class Box(Space[NDArray[Any]]):
```
RedTachyon, line 35: I'd omit the `Any` in `NDArray[Any]`, it doesn't add any information and makes it less readable

RedTachyon: (and in most other places where it's `[Any]`)

gymnasium/spaces/box.py
```diff
@@ -256,7 +257,7 @@ def __repr__(self) -> str:
         """
         return f"Box({self.low_repr}, {self.high_repr}, {self.shape}, {self.dtype})"
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
RedTachyon, line 260: This is probably technically correct (and redundant), but for consistency it should be `other: Any`

gymnasium/spaces/box.py
```diff
@@ -266,7 +267,7 @@ def __eq__(self, other) -> bool:
             and np.allclose(self.high, other.high)
         )
 
-    def __setstate__(self, state: Dict):
+    def __setstate__(self, state: Iterable[tuple[str, Any]] | Mapping[str, Any]):
```
RedTachyon, line 270: This is new. Is this actually a thing that can happen in `__setstate__`? As in, can it be an iterable or a non-dict mapping?

gymnasium/spaces/dict.py
```diff
@@ -1,18 +1,19 @@
...
+
 from collections import OrderedDict
-from collections.abc import Mapping, Sequence
+from collections.abc import Mapping as CollectionsMapping
 from typing import Any
 from typing import Dict as TypingDict
-from typing import List, Optional
-from typing import Sequence as TypingSequence
-from typing import Tuple, Union
+from typing import Mapping as TypingMapping
```
RedTachyon, line 8: I'd say
```python
import collections.abc
import typing
# : collections.abc.Mapping
# : typing.Mapping
```

I know it's already in the code, but `TypingDict` should also be turned into `typing.Dict` etc.

RedTachyon: Or actually `typing.Dict` isn't necessary anymore now that we have `from __future__ import annotations`

gymnasium/spaces/space.py
```diff
@@ -1,26 +1,19 @@
...
 
 import numpy as np
+import numpy.typing as npt
 
 from gymnasium.utils import seeding
 
 T_cov = TypeVar("T_cov", covariant=True)
 
 
+MASK_NDARRAY = npt.NDArray[np.int8]
```
RedTachyon, line 14: I don't think it should be in all caps, it's a type synonym and not really a constant (so `MaskNDArray` or something)


Also - please double-check that it should indeed always be a `np.int8` ndarray, I haven't really kept up with the sample mask feature

gymnasium/spaces/multi_binary.py
```diff
@@ -90,29 +93,33 @@ def sample(self, mask: Optional[np.ndarray] = None) -> np.ndarray:
...
+    def contains(self, x: Any) -> bool:
         """Return boolean specifying if x is a valid member of this space."""
         if isinstance(x, Sequence):
             x = np.array(x)  # Promote list to array for contains check
 
         return bool(
             isinstance(x, np.ndarray)
             and self.shape == x.shape
-            and np.all((x == 0) | (x == 1))
+            and np.all(np.logical_or(x == 0, x == 1))
```
RedTachyon, line 104: Is this required for typing, and exactly equivalent?

gymnasium/spaces/sequence.py
```diff
@@ -1,20 +1,25 @@
...
 
 import numpy as np
+import numpy.typing as npt
 
-import gymnasium as gym
 from gymnasium.spaces.space import Space
 
 
-class Sequence(Space[Tuple]):
+class Sequence(Space[Tuple[Any, ...]]):
```
RedTachyon, line 15: We can use the generic here, and omit the `[Any, ...]` (so `Tuple[Any, ...]` -> `tuple`)

gymnasium/spaces/sequence.py
```diff
@@ -54,8 +59,14 @@ def is_np_flattenable(self):
 
     def sample(
         self,
-        mask: Optional[Tuple[Optional[Union[np.ndarray, int]], Optional[Any]]] = None,
-    ) -> Tuple[Any]:
+        mask: None
+        | (
+            tuple[
+                None | (np.integer | npt.NDArray[np.integer]),
+                Any | None,
```
RedTachyon, line 66: `Any | None == Any`

gymnasium/spaces/space.py
```diff
@@ -71,23 +64,30 @@ def __init__(
...
+        """Lazily seed the PRNG since this is expensive and only needed if sampling from this space.
+
+        As :meth:`seed` is not guaranteed to set the `_np_random` for particular seeds. We add a
+        check after :meth:`seed` to set a new random number generator.
+        """
         if self._np_random is None:
             self.seed()
 
-        return self._np_random  # type: ignore  ## self.seed() call guarantees right type.
+            if self._np_random is None:
```
RedTachyon, line 75: This is new logic. Can't we just do an assertion here? if `self.seed()` doesn't create `self._np_random`, then there's something wrong with the space implementation

gymnasium/spaces/text.py
```diff
@@ -69,7 +70,9 @@ def __init__(
         super().__init__(dtype=str, seed=seed)
 
     def sample(
-        self, mask: Optional[Tuple[Optional[int], Optional[np.ndarray]]] = None
+        self,
+        mask: None
+        | (tuple[int | None, np.ndarray[Any, np.dtype[np.int8]] | None]) = None,
```
RedTachyon, line 75: We should be consistent with either `np.ndarray` or `NDArray` in typing. 

Are you sure about the generic arguments to `np.ndarray` here?

gymnasium/spaces/tuple.py
```diff
@@ -1,16 +1,17 @@
 """Implementation of a space that represents the cartesian product of other spaces."""
+from __future__ import annotations
+
 from collections.abc import Sequence as CollectionSequence
-from typing import Iterable, Optional
+from typing import Any, Iterable
 from typing import Sequence as TypingSequence
```
RedTachyon, line 6: `typing.Sequence`

gymnasium/spaces/tuple.py
```diff
@@ -1,16 +1,17 @@
...
 from typing import Tuple as TypingTuple
-from typing import Union
 
 import numpy as np
 
 from gymnasium.spaces.space import Space
 
 
-class Tuple(Space[tuple], CollectionSequence):
+class Tuple(Space[TypingTuple[Any, ...]], TypingSequence[Any]):
```
RedTachyon, line 14: This changes the logic, I'm not sure if this is intended.

By inheriting `collections.abc.Sequence`, we get a few automatic mixin methods: https://docs.python.org/3/library/collections.abc.html

As far as I know (though I could be wrong), the same isn't true of `typing.Sequence`, so this would remove e.g. `reversed(space)`

gymnasium/spaces/tuple.py
```diff
@@ -87,8 +86,8 @@ def seed(
         return seeds
 
     def sample(
-        self, mask: Optional[TypingTuple[Optional[np.ndarray], ...]] = None
-    ) -> tuple:
+        self, mask: TypingTuple[Any | None, ...] | None = None
+    ) -> TypingTuple[Any, ...]:
```
RedTachyon, line 90: `tuple` gives the same information but is simpler

gymnasium/spaces/utils.py
```diff
@@ -47,7 +52,7 @@ def flatdim(space: Space) -> int:
          NotImplementedError: if the space is not defined in ``gym.spaces``.
          ValueError: if the space cannot be flattened into a :class:`Box`
     """
-    if not space.is_np_flattenable:
+    if space.is_np_flattenable is False:
```
RedTachyon, line 55: I think the old version is more idiomatic. It's also technically not the same thing, if someone does `space.is_np_flattenable = 0`, the result will be different

gymnasium/spaces/utils.py
```diff
@@ -102,7 +107,9 @@ def _flatdim_text(space: Text) -> int:
 
 
 T = TypeVar("T")
-FlatType = Union[np.ndarray, TypingDict, tuple, GraphInstance]
+FlatType = Union[
```
RedTachyon, line 110: `NDArray | dict[str, Any] | tuple | GraphInstance`

I'm not 100% sure if this is the actual signature we want for `flatten`, but it will be yet another thing to look at later

gymnasium/spaces/utils.py
```diff
@@ -137,19 +157,21 @@ def flatten(space: Space[T], x: T) -> FlatType:
...
 @flatten.register(Box)
 @flatten.register(MultiBinary)
-def _flatten_box_multibinary(space, x) -> np.ndarray:
+def _flatten_box_multibinary(space: Box | MultiBinary, x: NDArray[Any]) -> NDArray[Any]:
     return np.asarray(x, dtype=space.dtype).flatten()
 
 
 @flatten.register(Discrete)
-def _flatten_discrete(space, x) -> np.ndarray:
+def _flatten_discrete(space: Discrete, x: int) -> NDArray[np.int64]:
```
RedTachyon, line 165: `NDArray[np.integer]`? Doesn't have to be `int64` afaik

gymnasium/spaces/utils.py
```diff
@@ -315,19 +354,21 @@ def _graph_unflatten(space, x):
...
-def _unflatten_sequence(space: Sequence, x: tuple) -> tuple:
+def _unflatten_sequence(
+    space: Sequence, x: TypingTuple[Any, ...]
+) -> TypingTuple[Any, ...]:
     return tuple(unflatten(space.feature_space, item) for item in x)
 
 
 @singledispatch
-def flatten_space(space: Space) -> Union[Dict, Sequence, Tuple, Graph]:
+def flatten_space(space: Space[Any]) -> Box | Dict | Sequence | Tuple | Graph:
```
RedTachyon, line 371: `Box` is new here, I assume it was a bug?

tests/spaces/test_dict.py
```diff
@@ -25,18 +27,18 @@ def test_dict_init():
     ):
         Dict(a=Discrete(2), b="Box")
 
-    with pytest.warns(None) as warnings:
+    with warnings.catch_warnings(record=True) as caught_warnings:
```
RedTachyon, line 30: Shouldn't we keep doing testing stuff with `pytest`? Is there a meaningful difference in functionality?

gymnasium/spaces/tuple.py
RedTachyon: Actually, nevermind all that, they are actually equivalent, except `typing.Sequence` is generic and the ABC version is not.

Except in python 3.9 `typing.Sequence` is deprecated, because the ABC version also became generic. Can you check if in other supported python versions, `from __future__ import annotations` enables using `collections.abc.Sequence[something]` as well?

If this works, we can consider switching from `typing.Sequence` to `collections.abc.Sequence` in general. If not, we should just stop using `collections.abc.Sequence` in favor of `typing.Sequence`

RedTachyon (changes requested):
There's a bunch of comments, a good chunk of them being rather pedantic (but many of the changes themselves are pedantic as well, so if we're doing it, we need to do it right)

Most notably I don't really like the pattern of putting `Something[Any]` everywhere, it doesn't add any information and is just additiona visual clutter.

Also many of the comments apply to multiple places in the code, I didn't comment each individual instance to avoid 100+ review comments

gymnasium/spaces/box.py
pseudo-rnd-thoughts: Good point, changed

gymnasium/spaces/box.py
pseudo-rnd-thoughts: As `Space.__setspace__` calls `dict(state)` so as `Iterable[tuple[str, Any]]` works then I have used the type however I agree it is weird

gymnasium/spaces/dict.py
pseudo-rnd-thoughts: Updated

gymnasium/spaces/space.py
pseudo-rnd-thoughts: Updated

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: True but `is_np_flattenable` should be a bool

tests/spaces/test_dict.py
pseudo-rnd-thoughts: The `pytest.warn(None)` is deprecated in pytest 7
This is the new recommended style

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: It was a bug in the old code, `Box -> Box` so I added it

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: `Discrete.dtype` is always `np.int64` so I think this is correct

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: Agreed

gymnasium/spaces/multi_binary.py
pseudo-rnd-thoughts: I believe this is equivalent and it makes it more clear for users

gymnasium/spaces/sequence.py
pseudo-rnd-thoughts: I think that `Tuple[Any, ...]` provides more detail than `tuple` and `tuple[Any, ...]` will cause error pyright reports

gymnasium/spaces/sequence.py
pseudo-rnd-thoughts: I find Any | None more descriptive and helpful but will change

gymnasium/spaces/space.py
pseudo-rnd-thoughts: For composite spaces, there are cases where you don't seed the super `np_random`, i.e. `Dict` space where only the subspaces are seeded not the super.
We have a couple of options
1. Keep it as is - the type will be sometimes incorrect
2. Change return type to optional `np.random.Generator` however this will force type asserts from pyright
3. Change the seed functions to always set the super `np_random` and raise error
4. Add this code to set a random np_random if the seed doesn't do it

I went to 4 as it requires minimal code change and doesn't require any code changes from custom spaces

gymnasium/spaces/text.py
pseudo-rnd-thoughts: Good spot, updating

gymnasium/spaces/tuple.py
pseudo-rnd-thoughts: I hadn't noticed that I used a different Sequence. I will change back to collections.abc

gymnasium/spaces/tuple.py
pseudo-rnd-thoughts: I find `Tuple[Any, ...]` is more obvious to the users what the type is compared to `tuple` which I agree is simpler

gymnasium/spaces/box.py
pseudo-rnd-thoughts: I think when I was testing with the strong pyright (including reportGeneralTypeIssues) then pyright was complaining saying that the Generic type was missing from `Space` and `npt.NDArray` so while it is obvious Im going to keep it for now

gymnasium/spaces/space.py
RedTachyon: In this case can you just unindent the second condition? I don't think the nesting is necessary

RedTachyon (approved):
One small stylistic comment, I'm fine with everything else

PR #39, opened 2022-10-06
Update the type hinting for core.py

pseudo-rnd-thoughts:
As core.py is the foundation for the whole of Gymnasium, it should have good type hinting to help users understand the expected types for functions.

This PR fills the gaps in the implementation where type hints were currently missing.

One of the things not done in this PR is to make this require pyright strict type hinting to enforce a number of type hinting specifications that are not applied currently to the project

Edit: I have added `ABC` and `abstractmethod` to `Env`, `Wrapper`, `Env.step`, `Env.reset`


gymnasium/core.py
```diff
@@ -340,18 +338,18 @@ def __repr__(self):
         return str(self)
 
     @property
-    def unwrapped(self) -> Env:
+    def unwrapped(self) -> Env[ObsType, ActType]:
```
RedTachyon, line 341: Not necessarily true - consider the wrappers that discretize actions, or add pixel observations

gymnasium/wrappers/monitoring/video_recorder.py
```diff
@@ -46,16 +50,12 @@ def __init__(
...
-        self.enabled = enabled
-        self._closed = False
-
         self.render_history = []
         self.env = env
 
         self.render_mode = env.render_mode
 
-        if "rgb_array_list" != self.render_mode and "rgb_array" != self.render_mode:
+        if "rgb_array" != self.render_mode and "rgb_array_list" != self.render_mode:
```
RedTachyon, line 58: Since we're changing this, can you flip the inequalities? Doing `<literal> != <variable>` makes me uneasy for a reason that stems either from some other language, of from some ancient evil hidden in the depths of python magic methods.

RedTachyon: We could also be even more clever and do `if self.render_mode not in ("rgb_array", "rgb_array_list")`

tests/test_core.py
```diff
@@ -1,139 +1,98 @@
...
 
-from gymnasium import core, spaces
-from gymnasium.wrappers import OrderEnforcing, TimeLimit
+from gymnasium import Env, ObservationWrapper, RewardWrapper, Wrapper
+from gymnasium.core import ActionWrapper, ActType, ObsType
+from gymnasium.spaces import Box
+from tests.generic_test_env import GenericTestEnv
 
 
-class ArgumentEnv(core.Env):
```
RedTachyon, line 10: Can you explain these changes? It seems that a bunch of tests were removed and replaced with new ones. Why? I'd be very, **very** careful about removing tests.

tests/wrappers/test_step_compatibility.py
```diff
@@ -1,35 +1,13 @@
...
-        self.observation_space = Discrete(2)
-
-    def step(self, action):
-        obs = self.observation_space.sample()
-        rew = 0
-        terminated = False
-        truncated = False
-        info = {}
-        return obs, rew, terminated, truncated, info
+OldStepEnv = partial(GenericTestEnv, step_fn=old_step_fn)
```
RedTachyon, line 9: What benefit does this give us? Instead of a self-contained, easy to understand implementation, now we have to go like "Okay, so it's a partial application, so we call the first argument, let's see what GenericTestEnv is... okay, and the argument is..."

I can definitely agree that it's more clever because of the whole partial application thing, but clear > clever

gymnasium/core.py
```diff
@@ -16,23 +16,18 @@
...
-        "Gymnasium minimally supports python 3.6 as the python foundation not longer supports the version, please update your version to 3.7+"
-    )
 
 ObsType = TypeVar("ObsType")
 ActType = TypeVar("ActType")
 RenderFrame = TypeVar("RenderFrame")
 
 
-class Env(Generic[ObsType, ActType]):
+class Env(Generic[ObsType, ActType], ABC):
```
RedTachyon, line 30: So this is something I've been contemplating for some time, whether Env should be an ABC, and I was always leaning more towards "No". It kind of is, and kind of isn't. Notice that, for example, `Env.reset` isn't in fact an abstract method because it does have some logic in it. In general, ABCs are meant to be empty skeletons, but we have a good chunk of built-in implementations. So I'm still kinda on the fence, and crucially I'm not 100% sure what runtime implications this has exactly.

gymnasium/core.py
```diff
@@ -16,23 +16,18 @@
 import numpy as np
 
 from gymnasium import spaces
-from gymnasium.logger import warn
 from gymnasium.utils import seeding
 
 if TYPE_CHECKING:
     from gymnasium.envs.registration import EnvSpec
 
-if sys.version_info[0:2] == (3, 6):
```
RedTachyon, line 25: Is this PR doubling for 3.6 removal?

RedTachyon (commented):
Added a bunch of extra comments.

And as a general note, please do keep the PRs a bit more atomic, or better described. A PR that only updates the type hints (as per the PR title) should have no reason to touch the tests whatsoever. If this were two separate PRs, one with types, and another with everything that affects tests, then we'd have one part that's safe to quickly review and merge, and then another part that needs careful review. Now, everything needs careful review.

gymnasium/core.py
pseudo-rnd-thoughts: Good point, removed

gymnasium/wrappers/monitoring/video_recorder.py
```diff
@@ -38,6 +38,10 @@ def __init__(
             Error: You can pass at most one of `path` or `base_path`
             Error: Invalid path given that must have a particular file extension
         """
+        self._async = env.metadata.get("semantics.async")
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I didn't have moviepy installed, so the try expect failed and the object was deleted, except `self.enable` didn't exist so a second error which originally confused me. So it is partially to remove a possible bug with users that haven't installed moviepy. I can change back as it is not related to the PR title

gymnasium/wrappers/monitoring/video_recorder.py
pseudo-rnd-thoughts: Changed

tests/test_core.py
pseudo-rnd-thoughts: I should have mentioned this in the top message, add `@abstractmethod` for `Env.step` and `Env.reset` caused a number of issues for testing environments where these functions were not filled out. See the CI in the initial commit. 

To simply the code, I made the environment inherit from `GenericTestEnv` that does have basic implementations of these functions. This simplifies the code and shouldn't have any effect on the tests

tests/wrappers/test_step_compatibility.py
pseudo-rnd-thoughts: Fair, I will change back

gymnasium/core.py
pseudo-rnd-thoughts: It wasn't intended to but I was doing both at the same time. Will revert

gymnasium/core.py
RedTachyon: Yep, I'd keep it non-abstract altogether. I'd also strongly consider moving towards asking env developers to call `super().step(action)` and similarly for all the other methods, but this would be a relatively complex procedure that I don't have sufficient justification and motivation to push for at the moment.

pseudo-rnd-thoughts: That makes sense and agree that `super().step(action)` could work but there isn't much good reasons for it other than we do it for `reset`

pseudo-rnd-thoughts:
@RedTachyon Removing the `ABC` and `abstractmethod` allowed me to revert most of the changes to other files. 

I have kept the changes to `test_core.py` as I thought that the tests before did not do a good job testing the base implementations (I actually found a weird bug). Let me know if you want it reverted back or more comments on the tests added.

The only major changes to before is two new TypeVars `WrapperObsType` and `WrapperActType`. These allow us to have  more complete types by differentiating between the wrappers environment types and the wrappers types. I believe this implementation is correct but due to Pyright having checks disabled we can't know for certain. 
This is probably the motivation needed to fix the pyright properties.

gymnasium/core.py
```diff
@@ -113,8 +115,8 @@ def reset(
         self,
         *,
         seed: Optional[int] = None,
-        options: Optional[dict] = None,
```
RedTachyon, line 116: I'd need to check, but I think that with removal of python 3.6 we can move (back) to using `dict[str, Any]` instead of `Dict[str, Any]`. It's not necessary in this PR, but might be a bit nicer long-term

gymnasium/core.py
```diff
@@ -194,14 +195,18 @@ def __enter__(self):
         """Support with-statement for the environment."""
         return self
 
-    def __exit__(self, *args):
+    def __exit__(self, *args: List[Any]):
```
RedTachyon, line 198: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

so it should be 
`def __exit__(self, *args: Any):`

(I'm not sure if Any wouldn't be automatically filled in any type checker, but doesn't hurt to include it)

gymnasium/core.py
```diff
@@ -212,55 +217,63 @@ class Wrapper(Env[ObsType, ActType]):
...
+        self._observation_space: Optional[spaces.Space[WrapperObsType]] = None
         self._reward_range: Optional[Tuple[SupportsFloat, SupportsFloat]] = None
-        self._metadata: Optional[dict] = None
+        self._metadata: Optional[Dict[str, Any]] = None
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str):
         """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
-        if name.startswith("_"):
+        if name == "_np_random":
```
RedTachyon, line 235: I'm not entirely sure about this, it's out of scope, and seems like a guardrail that's not necessarily needed.

gymnasium/core.py
```diff
@@ -393,28 +414,30 @@ class RewardWrapper(Wrapper):
     because it is intrinsic), we want to clip the reward to a range to gain some numerical stability.
     To do that, we could, for instance, implement the following wrapper::
 
-        class ClipReward(gym.RewardWrapper):
+        class ClipReward(gymnasium.RewardWrapper):
```
RedTachyon, line 417: Imo it's better to keep it as `gym.RewardWrapper` since that's the recommended usage, and the code snippet isn't (a) related to the gym-gymnasium compatibility, or (b) runnable by itself

tests/test_core.py
```diff
@@ -1,139 +1,143 @@
...
-    },
-]
+class ExampleObservationWrapper(ObservationWrapper):
+    def observation(self, observation: ObsType) -> ObsType:
+        return np.array([1])
 
 
-@pytest.mark.parametrize("class_", [UnittestEnv, UnknownSpacesEnv])
-@pytest.mark.parametrize("props", properties)
-def test_wrapper_property_forwarding(class_, props):
```
RedTachyon, line 116: I'm still a bit hesitant about removing the existing tests.

Are these tests redundant now? Do they fail with the new changes? (if so, that's a potential problem) Are they 100% replaced with the new tests?

I'd lean towards keeping them in, and just adding new tests instead. And if some tests turn out to be unnecessary, we can remove them separately with no other changes.

RedTachyon (changes requested):
Added a few, mostly minor comments. The biggest thing is probably the tests, since there's technically a chance that this introduces a bug/breaking behavior that sneaks past because the corresponding test is removed.

gymnasium/core.py
pseudo-rnd-thoughts: Wow impress check, will update

gymnasium/core.py
pseudo-rnd-thoughts: This was a guardrail that already existed, but it was bugged so didn't work. I had to modify this function to re-add it such that the correct error occurred.

We could replace it with a warning? I added it to gym a couple of months ago as I thought I had found a bug, but what was accurately happening was that I was using the wrapper `_np_random` not the environment `_np_random` which took me about an hour + to solve. Therefore, I proposed this solution. I guess the issue is that wrappers could modify the reproducibility of an environment using the environment rng not the wrapper rng however in this case, the wrapper could use its own rng with a different variable name. Thoughts?

gymnasium/core.py
pseudo-rnd-thoughts: I will change all of the example code to be consistent

gymnasium/core.py
RedTachyon: I think it's fine for now, we can reconsider this guardrail in general in the future

RedTachyon (commented):
Looks good overall. I'm wondering if we should maybe make `env.spec` mandatory so we can omit all the `assert self.spec is not None`, but within the scope of typing, this can be merged

RedTachyon (approved):

PR #42, opened 2022-10-08
fix: remove auto close feature in VideoRecorder

younik:
If user forgets to call `close`, and the VideoRecorder is deleted by the garbage collector, the class tries to save the video, but it raises an error (and it is not easy to understand the reason). With this PR, RecordVideo just warns the user that the last video is not saved and to call close. 

It also adds a parameter `disable_logger` to disable MoviePy messages.
Screenshot from 2022-10-08 13-33-51

### DIFF
--- a/gymnasium/wrappers/monitoring/video_recorder.py
+++ b/gymnasium/wrappers/monitoring/video_recorder.py
@@ -155,7 +155,7 @@ class VideoRecorder:
 
             logger.debug(f"Closing video encoder: path={self.path}")
             clip = ImageSequenceClip(self.recorded_frames, fps=self.frames_per_sec)
-            clip.write_videofile(self.path)
+            clip.write_videofile(self.path, logger=None)
         else:
             # No frames captured. Set metadata.
             if self.metadata is None:
--- a/gymnasium/wrappers/monitoring/video_recorder.py
+++ b/gymnasium/wrappers/monitoring/video_recorder.py
@@ -24,6 +24,7 @@ class VideoRecorder:
         metadata: Optional[dict] = None,
         enabled: bool = True,
         base_path: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts:
Do we want this a debug message when a video is saved as depending on the video length then can take some time I believe?

younik:
> Do we want this a debug message when a video is saved as depending on the video length then can take some time I believe?

Good point; on my laptop, it takes ~5 seconds for 1000 frames.

There is already line 156:
```python
logger.debug(f"Closing video encoder: path={self.path}")
```

I can add a `Done!` message after line 158.
Alternatively, we can set a threshold on the number of frames to turn on logger.

pseudo-rnd-thoughts:
We could do with a parameter `disable_logger: bool = False`? This should allow people to disable if they want but it is enabled by default

pseudo-rnd-thoughts (commented):
Is it possible to save multiple videos?

younik:
> Is it possible to save multiple videos?

The `close` method of VideoRecorder is called every time the recording of the current video has finished and the video is saved. This can happen multiple times (depending on the parameters of RecordVideo). Does this answer your question?

pseudo-rnd-thoughts:
Yes, but I dislike the answer in the sense that to me close should only be called once. But I guess users can now use the function version of the wrappers for cases like that if they care

younik:
> Yes, but I dislike the answer in the sense that to me close should only be called once. But I guess users can now use the function version of the wrappers for cases like that if they care

Yes, I agree; we decided to keep VideoRecorder after discussion here: https://github.com/openai/gym/pull/2894
It is also used by wandb

pseudo-rnd-thoughts (approved):

PR #46, opened 2022-10-09
Support gym env in Gymnasium's make

younik:
# Description

With this PR, Gymnasium can load environments that are registered in gym, but not in Gymnasium. 

The environment will still be a gym.Env instance.

## Type of change

New feature

# Checklist:

### DIFF
--- a/gymnasium/envs/registration.py
+++ b/gymnasium/envs/registration.py
@@ -312,8 +312,6 @@ def load_env_plugins(entry_point: str = "gymnasium.envs") -> None:
 # fmt: off
 @overload
 def make(id: str, **kwargs) -> Env: ...
-
-
 @overload
 def make(id: EnvSpec, **kwargs) -> Env: ...
 
@@ -322,21 +320,12 @@ def make(id: EnvSpec, **kwargs) -> Env: ...
 # ----------------------------------------
 @overload
 def make(id: Literal["CartPole-v0", "CartPole-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
-
-
 @overload
 def make(id: Literal["MountainCar-v0"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
-
-
 @overload
-def make(id: Literal["MountainCarContinuous-v0"], **kwargs) -> Env[
-    np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
-
-
+def make(id: Literal["MountainCarContinuous-v0"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
 @overload
 def make(id: Literal["Pendulum-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
-
-
 @overload
 def make(id: Literal["Acrobot-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
 
@@ -344,15 +333,9 @@ def make(id: Literal["Acrobot-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarra
 # Box2d
 # ----------------------------------------
 @overload
-def make(id: Literal["LunarLander-v2", "LunarLanderContinuous-v2"], **kwargs) -> Env[
-    np.ndarray, Union[np.ndarray, int]]: ...
-
-
+def make(id: Literal["LunarLander-v2", "LunarLanderContinuous-v2"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
 @overload
-def make(id: Literal["BipedalWalker-v3", "BipedalWalkerHardcore-v3"], **kwargs) -> Env[
-    np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
-
-
+def make(id: Literal["BipedalWalker-v3", "BipedalWalkerHardcore-v3"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
 @overload
 def make(id: Literal["CarRacing-v2"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, Sequence[SupportsFloat]]]: ...
 
@@ -361,16 +344,10 @@ def make(id: Literal["CarRacing-v2"], **kwargs) -> Env[np.ndarray, Union[np.ndar
 # ----------------------------------------
 @overload
 def make(id: Literal["Blackjack-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
-
-
 @overload
 def make(id: Literal["FrozenLake-v1", "FrozenLake8x8-v1"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
-
-
 @overload
 def make(id: Literal["CliffWalking-v0"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
-
-
 @overload
 def make(id: Literal["Taxi-v3"], **kwargs) -> Env[np.ndarray, Union[np.ndarray, int]]: ...
 
@@ -391,8 +368,6 @@ def make(id: Literal[
     "HumanoidStandup-v2", "HumanoidStandup-v4",
     "Humanoid-v2", "Humanoid-v3", "Humanoid-v4",
 ], **kwargs) -> Env[np.ndarray, np.ndarray]: ...
-
-
 # fmt: on
 
 
@@ -409,8 +384,8 @@ def _check_spec_register(spec: EnvSpec):
             spec_
             for spec_ in registry.values()
             if spec_.namespace == spec.namespace
-               and spec_.name == spec.name
-               and spec_.version is not None
+            and spec_.name == spec.name
+            and spec_.version is not None
         ),
         key=lambda spec_: int(spec_.version),  # type: ignore
         default=None,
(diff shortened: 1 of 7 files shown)


RedTachyon:
Are you able to give any examples where this would happen? Also I'm pretty sure it won't return a valid `gymnasium.Env` instance, it will be `gym.Env`. Right now it's effectively the same thing, sure, but this won't be the case in the future as we add/change features. And even now, I'm pretty sure it could result in situations where `isinstance(gymnasium.make(...), gymnasium.Env) == False`, which is very much not a good thing.

younik:
> Are you able to give any examples where this would happen? 

Yes, you can try with ALE, or commenting env registration in Gymnasium (in `envs.__init__.py`).

> Also I'm pretty sure it won't return a valid `gymnasium.Env` instance, it will be `gym.Env`. Right now it's effectively the same thing, sure, but this won't be the case in the future as we add/change features. And even now, I'm pretty sure it could result in situations where `isinstance(gymnasium.make(...), gymnasium.Env) == False`, which is very much not a good thing.

Yes, this is true. I don't think it is a good idea to keep this feature for a long time, especially if we will do changes in the env API. 
I added a dummy wrapper, that at least makes true `isinstance(env, gymnasium.Env) == False` (but false `isinstance(env.unwrapped, gymnasium.Env) == False`).

RedTachyon:
> I added a dummy wrapper

I don't think this is a good idea, all we're doing is pushing the problem one step down. I'm also wondering why the type checker isn't complaining about this, but it is what it is.

If we want this kind of compatibility, it can't be a wrapper per se, but a proper `GymEnv(gymnasium.Env)` which takes in a `gym.Env` in the constructor, and properly routs all the properties/methods.

Overall I'm pretty doubtful about the utility of this feature, but I guess it's been discussed somewhere outside?

pseudo-rnd-thoughts:
> I don't think this is a good idea, all we're doing is pushing the problem one step down. I'm also wondering why the type checker isn't complaining about this, but it is what it is.

I think we both misremembered. In the original passive environment checker we had this check that the environment has a gym.Env however we encouraged me to remove it to minimise error.
Therefore, surprisingly, this works fine without raising any error even with the passive environment checker active.

But I think we should apply the gymnasium wrappers not the gym wrappers. So you should be able to remove the `Wrapper` on the `gym.make` then have the returned environment as the env_creator stripped of the wrappers automatically applied to the environment, i.e., passive environment checker, order enforcing, etc. This should allow an environment have still have wrappers that are automatically used with the environment applied. I hope this makes sense
 Also could the warning be after the `gym.make` has passed to say, "Environment has been loaded from Gym, please register the environment in the Gymnasium as well."

pseudo-rnd-thoughts:
An alternative implementation
```python
try:
    import gym

    if env_id in gym.envs.regstration.registry:
        gym_env_spec = EnvSpec(**dataclasses.asdict(gym.spec(env_id)))

        return gymnasium.make(gym_env_spec)
```

younik:
> An alternative implementation

I did something in this direction, without losing the nice features in make (e.g. default version, module loading).

So, right now, the spec is extracted from the `gym.registry` and then the environment is created with Gymnasium wrappers and control.

>Also could the warning be after the gym.make has passed to say, "Environment has been loaded from Gym, please register the environment in the Gymnasium as well."

Done

> proper GymEnv(gymnasium.Env) which takes in a gym.Env in the constructor, and properly routs all the properties/methods.

I removed the base Wrapper; but I don't see the utility of adding this GymEnv. Yes, right now, `isinstance(env.unwrapped, gymnasium.Env)` is False, but the `make` function gives two warnings about it, thus this should not be a surprise. Where do you think this would be helpful?

> Overall I'm pretty doubtful about the utility of this feature, but I guess it's been discussed somewhere outside?

We discussed this on Discord; I think it can be useful for helping people migrating to Gymnasium while envs are not supporting it yet. My idea is to remove this feature in the future.

gymnasium/envs/registration.py
```diff
@@ -543,31 +587,21 @@ def make(
...
-            )
+        spec_ = _search_spec(id, ns, name, version, registry)
 
         if spec_ is None:
-            _check_version_exists(ns, name, version)
-            raise error.Error(f"No registered env with id: {id}")
+            spec_ = _search_in_gym(id, ns, name, version)
+
+            if spec_ is not None:
+                disable_env_checker = True
```
pseudo-rnd-thoughts, line 598: This should be `spec_.disable_env_checker=True`. 
Or we modify the passive environment checker such that it only checks that the observation and action spaces exist not the type isinstance Gymnasium.Space which I believe is the current issue.

gymnasium/envs/registration.py
```diff
@@ -543,31 +587,21 @@ def make(
...
+                logger.warn(
+                    "Environment has been loaded from Gym, please register the environment in the Gymnasium as well."
```
pseudo-rnd-thoughts, line 600: "the environment in Gymnasium as well"

pseudo-rnd-thoughts (commented):
This is probably too difficult to add, but is it possible to add test for this?

tests/envs/utils_envs.py
```diff
@@ -1,6 +1,13 @@
+import gym as legacy_gym
```
pseudo-rnd-thoughts, line 1: Does this not require gym as a testing dependency?

setup.py
```diff
@@ -25,9 +25,16 @@
...
-) + ["pytest==7.0.1"]
+extras["testing"] = (
+    list(
+        set(
+            itertools.chain.from_iterable(
+                map(lambda group: extras[group], testing_group)
+            )
+        )
+    )
+    + ["pytest==7.0.1", "gym>=0.26"]
```
younik, line 36: @pseudo-rnd-thoughts 
Yes, I thought I added the dependency in this line, but apparently no looking at the tests
Do you know why?

pseudo-rnd-thoughts: Python 3.6 has a custom build requirement, see py.Dockerfile line 29

younik: Oh, I see, thanks

RedTachyon (changes requested):
We discussed it a bit with @pseudo-rnd-thoughts and I still think that this is a bad idea overall, at least with the current implementation.

I want to start with one remark about how we need to maintain gymnasium at this point. In the weird migration period between the APIs, we were breaking compatibility, implementing weird hacks and so on, because we were in that weird spot where we needed that. Now we're past that point, and to be taken seriously, we must make sure that we don't add and remove features arbitrarily, don't rename things too often, stuff like that. Gymnasium has to be stable to be a viable option, and it needs to be robust (i.e. good, reliable code with no hacks)

Getting to the point, I have two issues: the implementation of the env itself, and the way it's created. 

1. With the current implementation, as previously mentioned, it is not necessarily the case that `isinstance(gymnasium.make(...).unwrapped, gymnasium.Env)`. The updated version kinda makes it so that `isinstance(gymnasium.make(...), gymnasium.Env)` because we're wrapping a `gym.Env` with `gymnasium.Wrapper`. This, however, is a hack - a wrapper should wrap around a valid `gymnasium.Env`, so we're just abusing python's duck typing. This will break the moment we do something new with `gymnasium.Env`, and even now is pretty iffy.
2. With the current implementation, the conversion is automatic/implicit. There might be a warning, but I don't think that changes anything. More importantly, if we add a feature "You can register an env in gym, and get it in gymnasium", we basically need to keep this feature for the foreseeable future. This is **very** bad. In one of the comments you actually state that you intend this to be removed in the future - we can't do things like this anymore.

As for the solutions - for (1) I still stand by what I said earlier, that we need a proper conversion class which explicitly converts `gym.Env` to `gymnasium.Env`. It could live here, or we could (later?) put it in Shimmy along with all other conversions. For (2), I wouldn't make it automatically integrated into `gymnasium.make`. It might be enough to keep the wrapper explicitly available and say "If you want to use a gym env with gymnasium, use the following class". If we want something more usable, we need to think how we want to handle other compatibilities. Do we want, in the future, `dm_env` envs to be usable via `gymnasium.make`? If so, we need to design it well in a cohesive way, before implementing an ad-hoc solution just for gym.


As I understand, the main problem is with ALE, which might need separate treatment. But we can't make problematic changes to the library just to handle a (hopefully temporary) issue.

pseudo-rnd-thoughts:
After more discussions with @RedTachyon, we have agreed that the best options is to use the namespace to indicate the environment should be loaded from gym, i.e., `gymnasium.make("gym/CartPole-v1")` that allows users to use original gym environments in gymnasium along with external third-party environments that makes it clear where the environment is being loaded from.

In addition, this makes loading environments simpler
```python
## in gymnasium.make
if ns == "gym":
   import gym as legacy_gym
   env = legacy_gym.make(id)
   return GymCompatibility(env)
```

Here an additional wrapper (`GymCompatibility`, name to probably change) is applied to the outside of the environment that converts the observation and action spaces to the gymnasium versions. Currently, as there is no difference between the gym and gymnasium spaces, we should be safely convert the spaces. 

The advantage of this approach is that it makes it clear where the environment is being loaded from (allowing future API to be supported as well, i.e., dm_env) and all of the gym wrapper magic is kept with a simple conversion wrapper that can be updated over time. 
The limitation of this approach is that custom spaces are not supported. However if we use the singledispatch to implement this, users can add their own functions for custom spaces to add compatibility.

RedTachyon:
@younik FYI Mark and I are planning to take a slightly different approach for this functionality, soon I'll either make a PR or do a more detailed write-up of what we want to do.

younik:
> we basically need to keep this feature for the foreseeable future

If this is true, then this feature should not be added. We cannot ensure for the long term to be able to translate a gym.Env to gymnasium.Env.

My points were: adding/deleting this feature doesn't change anything about the API, and will be transparent when all third-party envs will migrate to Gymnasium. If this doesn't happen, it is third-party libraries' "fault" that is not updated, and simply that environment will not be supported in the future.
About typing, again, my idea was to remove the feature as soon as we cannot ensure that gym.Env and gymansium.Env share the same API (and likely, far before this happens). I also don't like it conceptually, but pragmatically, as a user I need this functionality, otherwise I will simply not use Gymnasium because every well-known benchmark is in gym.

> gymnasium.make("gym/CartPole-v1")

This will not work for envs loaded during make, you will need to handle something like `gymnasium.make("gym/ALE/Breakout-v5")`, but anyway, the code will break when ALE will migrate to Gymnasium. 


Anyway, consider the above points and choose what looks best to you.

PR #54, opened 2022-10-13
Updated the utils.__init__ and documentation

pseudo-rnd-thoughts:
`utils.__init__` has not been updated to include all of the new utils functions that users will commonly used.
Adding these functions to the `__init__` should make the imports shorter and easier to remember for users.
In addition, we can update the documentation to use these shorter links

### DIFF
--- a/docs/api/utils.md
+++ b/docs/api/utils.md
@@ -4,34 +4,6 @@ title: Utils
 
 # Utils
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
It seems like the circular imports makes this too difficult to practically achieve

PR #56, opened 2022-10-13
Add Tutorials using sphinx-gallery and other template updates

mgoulao:
# Description

[Live example](https://mgoulao.github.io/Gymnasium/tutorials/demo/)

This PR adds
* tutorials using sphinx-gallery. The script used is available at [https://github.com/Farama-Foundation/Celshast/blob/main/src/furo/gen_tutorials.py](https://github.com/Farama-Foundation/Celshast/blob/main/src/furo/gen_tutorials.py). I added a demo tutorial (`tutorials/demo.py`) as an example.
* New template configs (site description, and image)
* update favicon
* Add link to the "compatibility with gym" page at the home page

I also change the number of environments in the display grid to 8.


pseudo-rnd-thoughts:
Is there an automatical way that gets added to the tutorials section on the website? 
Could the current tutorials to moved to this style?

mgoulao:
> Is there an automatical way that gets added to the tutorials section on the website?
 
Can you elaborate

> Could the current tutorials to moved to this style?

Do you want me to do it in this PR?

pseudo-rnd-thoughts:
Sorry, terrible English. 
If we have a folder with all of the tutorials, then could all of the tutorials be automatically added to a page / toctree?

How much work would it be? If it is too much work then don't worry and we can do it in a new PR

mgoulao:
> tutorials be automatically added to a page / toctree?

done

> How much work would it be?

a new PR is a good idea

pseudo-rnd-thoughts (approved):
The current tutorials are not added as part of this PR, this will be done in a future PR

PR #58, opened 2022-10-14
setup.py cleanup

jjshoots:
# Description

This updates the setup.py file to be consistent with that of PettingZoo.
The list of changes are as follows:
- Move version into gymnasium/__init__.py
- Updated license
- Edit setup to be in line with that of pettingzoo


pseudo-rnd-thoughts (commented):
Why the need to change the gymnasium setup.py or change version. This seemed like simple and effective method of doing it all

jjshoots:
Ready for review @pseudo-rnd-thoughts

setup.py
```diff
@@ -51,33 +55,23 @@ def get_version():
     set(itertools.chain.from_iterable(map(lambda group: extras[group], all_groups)))
 )
 
+version = get_version()
```
pseudo-rnd-thoughts, line 58: This function doesn't seemed to be used and the setup version=get_version

pseudo-rnd-thoughts (commented):
Generally looks good, just a comment on the version variable

setup.py
jjshoots: Fixed to do `version=version`.

pseudo-rnd-thoughts (approved):

PR #61, opened 2022-10-18
Gym-Gymnasium compatibility converter

RedTachyon:
Based on many external discussions, this implements a way to use gym environments inside gymnasium.

They can be created in two ways -- either pass it by name with optional arguments to create an env from the gym registry:
```python
import gymnasium as gym
gym.make("GymV26Environment-v0", env_id="CartPole-v0", make_kwargs={...})
```

or use an arbitrary instantiated gym env
```python
import gymnasium
import gym
env = gym.make(...)
gymnasium.make("GymV26Environment-v0", env=env)
```

At the moment it only supports the modern API used in gym 0.26, in the future we might add support for the old API, though at that point it might enter a dependency hell. We'll see.


gymnasium/envs/external/gym_env.py
```diff
@@ -0,0 +1,212 @@
...
+def _strip_default_wrappers(env: "gym.Env") -> "gym.Env":
+    """Strips builtin wrappers from the environment.
+
+    Args:
+        env: the environment to strip builtin wrappers from
+
+    Returns:
+        The environment without builtin wrappers
+    """
+    import gym.wrappers.compatibility  # Cheat because gym doesn't expose it in __init__
```
pseudo-rnd-thoughts, line 108: Remove

gymnasium/envs/external/gym_env.py
```diff
@@ -0,0 +1,212 @@
...
+    default_wrappers = (
+        gym.wrappers.RenderCollection,
+        gym.wrappers.HumanRendering,
+    )
+    while isinstance(env, default_wrappers):
+        env = env.env
+    return env
+
+
+def _convert_space(space: "gym.Space") -> gymnasium.Space:
```
pseudo-rnd-thoughts, line 120: Why are you not using the single dispatch?

pseudo-rnd-thoughts (commented):
A couple of initial comments

gymnasium/envs/external/gym_env.py
```diff
@@ -0,0 +1,212 @@
...
+        Args:
+            seed: the seed to reset the environment with
+            options: the options to reset the environment with
+
+        Returns:
+            (observation, info)
+        """
+        super().reset(seed=None)  # We don't need the seed inside gymnasium
+        # Options are ignored
+        return self.gym_env.reset()
```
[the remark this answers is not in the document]
RedTachyon: Just changed it

gymnasium/envs/external/gym_env.py
RedTachyon: Can't, we'd need to fix it in gym because `compatibility` and the other thing aren't imported in the `__init__.py`

(and we probably want to fix it in gymnasium too)

gymnasium/envs/external/gym_env.py
RedTachyon: Initially I had a version with single dispatch, but it crashes the code when running without gym installed - the functions are defined on the top level, so they are always created, and the registration decorator has to read the type annotations. Those type annotations are invalid if gym isn't installed, so single dispatch crashes.

RedTachyon: (there's probably a way to do some fancy conditional imports that would avoid it, but I figured this is the simpler way)

pseudo-rnd-thoughts: Agreed, this is the simpler way that there should be very few cases where users need to convert types. My only idea is if the function can defined in the environment constructor, does the decorator function still run?

RedTachyon: If the function was defined inside `__init__`, then I'm pretty sure it wouldn't run automatically, so it'd work. But this is also very cursed, and loses out on the potential benefit of doing single dispatch anyways (you can't register extra functions from the outside anyways), so probably not the best option.

gymnasium/envs/external/gym_env.py
pseudo-rnd-thoughts: But you are no longer using the compatibility or env_checker wrappers

gymnasium/envs/external/gym_env.py
pseudo-rnd-thoughts: Good point, I guess we need to keep the function version then for now until we move to shimmy then gym can be an optional install

pseudo-rnd-thoughts (approved):
Code is all good, could you update the PR top comment from "WiP" to be more descriptive and provide a short description of the PR rational over the prior ideas

pseudo-rnd-thoughts (commented):
Do we want a `__get_attr__` function to get attributes from the gym environment

RedTachyon:
> Do we want a `__get_attr__` function to get attributes from the gym environment

I'd rather not do that, I'm already slightly uneasy with the way getattr is handled in wrappers. If users need to access something from the backend environment, they can just do `env.gym_env.something` and it's clear where it comes from.

PR #70, opened 2022-10-23
added instructions for mujoco v3 and older

Kallinteris-Andreas:
# Description

added instructions for MuJoCo v3 and older

source:
https://discord.com/channels/961771112864313344/961772009770074142/1032587824626675722

Fixes # (issue)
did not provide instructions for MuJoCo v3 and older
`pip install gymnasium[mujoco]` does not work for mujoco-v3

note: it might be a good idea to add something like `pip install gymnasium[mujoco-old]` that would install `mujoco-py` instead of `mujoco`

## Type of change

Doc Change

### DIFF
--- a/docs/environments/mujoco.md
+++ b/docs/environments/mujoco.md
@@ -33,7 +33,9 @@ The unique dependencies for this set of environments can be installed via:
 pip install gymnasium[mujoco]
 ````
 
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR

I don't think we should do this but registering "fake" environments for mujoco-v2, -v3 or -v3 if the mujoco_py or mujoco is missing. Such that when users try and make these environments, an error is raised explaining the error. @RedTachyon Thoughts to make this less hacky. This will reduce the number of issues about atari in particular which with AutoROM always confuses people

PR #78, opened 2022-10-26
Added Atari environments to tests, removed dead code

Markus28:
- Adds some atari environments to tested environments (if gym and ale are available)
- Removed definition of `minimum_testing_env_specs`, which was dead code, also didn't make sense (compared specs to strings, I think)
- Atari environments are currently not being tested for `render_mode` because `GymEnvironment` doesn't support that kwarg

Tests are currently not passing locally, which seems to be due to an unrelated problem

### DIFF
--- a/tests/envs/test_envs.py
+++ b/tests/envs/test_envs.py
@@ -148,15 +148,17 @@ def check_rendered(rendered_frame, mode: str):
         )
 
 
-non_mujoco_py_env_specs = [
+# We do not check render_mode for some mujoco envs and any old Gym environment wrapped by `GymEnvironment`
+render_mode_env_specs = [
     spec
     for spec in all_testing_env_specs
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Looks good to me, could you just add a comment before the atari_id to explain why though particular environments were selected. 
We could replace the list with all of the environment ids in the registry that contains "Pong" just in case the number of environment expands

Markus28:
Yeah, I will add that comment :) 

I also considered fetching the ids from the (old) Gym registry, but I decided against it because that registry should not really change, given that Gym is no longer being maintained. Also, I would be somewhat worried that for some reason (e.g. ale not being installed) no Atari envs show up in the registry and the test is silently skipped.

Currently, neither this test, nor `test_gym_conversion` are in CI, because gym isn't being installed.

PR #85, opened 2022-10-27
[WiP] Rework the vector API

RedTachyon:
~~This is a **very early** draft PR to put the code out there.~~ Below I'm pasting a discord message where I tried to contain the important and difficult parts about doing this correctly.

We want to support three main types of env vectorization. The external API for each of them will be identical, but the difference lies in the internals.
1. (default, currently existing) Multiprocessing-based sync/async vector env -- this can stay more or less the way it is implemented right now
2. Envpool - it's fast, but only works with a very specific subset of environments that are implemented in envpool (in sync and async variants)
3. "Big Vec"/custom vectorization - an env can implement its own vector environment in whatever way it wants (which would usually be done via numpy broadcasting, jax vmapping, or something similar)

(we need a better name than big vec, but it is what it is for now)

There would be a new function, something like gym.make_vec("CartPole-v1", num_envs=100, vectorization="custom"), with the arguments doing what you can hopefully guess they're meant to do.

Now, the big problem here is wrappers. So far the only vectorization that officially exists in gymnasium is multiprocessing, which means that we have X fully-fledged copies of gym.Env living somewhere, either in the in the main thread, or in subprocesses. This means that each of them can be individually wrapped with whatever you need, and then aggregated.

This is not possible for the other two options. In envpool, individual environments exist as their own entities in the C++ code, which we don't mess with. With bigvec, individual environments might not even exist, the whole logic might be performed in a single batch.
So we need to do something about wrappers. Some of them are relatively unimportant, like the env checker or the order enforcing thing. Some of them are trivial to use/automatically convert to support vector envs, like the new (in development) lambda wrappers. Some of them are impossible to implement for a general vector env, like the atari preprocessing - which may step through the environment several times, but in a general vector env you can't step through each environment separately, or reset them selectively.

Currently, the relation between Env, VectorEnv and Wrapper is a lie. We pretend that a valid VectorEnv is a valid Env, which is not true because it doesn't return a float as a reward. This way, a Wrapper can accept a VectorEnv, and sometimes it will work, other times it will just crash.

We need to sort out the relationship between Env and VectorEnv. Either one inherits from the other, or they inherit from a "common ancestor" (more like implementing an interface than properly inheriting), or they are just two separate entities. We're leaning towards the last option right now.
This will have implications to what happens to Wrapper and VectorWrapper (which exists, but is never used in gymnasium). Right now we're pretending that a Wrapper can wrap a VectorEnv, but as I said, that's a lie. We might have to implement each wrapper in a single and vectorized way, which would be kinda annoying and duplicating the code. We might be able to automatically convert some wrappers, like the lambda wrappers, if they're written the right
In any case, we'll need to explicitly mark which wrappers work with which envs and with which vectorization methods. There needs to be a distinction between vectorizing wrapped environments, and wrapping vectorized environments. For example, I don't think there's a viable way for us to ensure the preprocessing wrapper working in envpool-vectorized Atari envs. But it can be definitely included with multiprocessing-based vectorization.

There's also a bunch of smaller considerations - the custom-implemented vectorized envs have to do autoreset, the custom vectorization class has to be registered, the supported vectorization modes have to be enumerated in the metadata, but all of this is pretty easy imo.


tests/vector/test_vector_env.py
```diff
@@ -53,16 +52,6 @@ def test_vector_env_equal(shared_memory):
     sync_env.close()
 
 
-def test_custom_space_vector_env():
```
RedTachyon, line 56: This was removed because trying to do `env = VectorEnv(...)` makes about as much sense as trying to do `env = gym.Env(...)`

tests/vector/test_vector_env_wrapper.py
```diff
@@ -8,8 +8,8 @@ def __init__(self, env):
         self.env = env
         self.counter = 0
 
-    def reset_async(self, **kwargs):
```
RedTachyon, line 11: SyncVectorEnv was refactored to not use `reset_async` (because why would it need to do async stuff if it's sync), so this was adapted

RedTachyon:
Closing this in favor of #93. We can reopen it if we'd rather merge directly to master

Issue #7, opened 2022-09-10
[Proposal] Migrate from `setup.py` to `pyproject.toml`


kir0ul:
### Proposal 

I'd like to propose to migrate from using `setup.py` to using `pyproject.toml`.

### Motivation

Probably not super urgent, but since `pyproject.toml` should be the new standard and `setup.py` is now considered legacy ([see `pip` documentation](https://pip.pypa.io/en/stable/reference/build-system/setup-py/#setup-py-legacy)), it could be good to migrate at some point.

### Pitch

- Migrate all the content of `setup.py` to `pyproject.toml`
- Have a single source of truth for all the dependencies in `pyproject.toml`, thus remove `requirements.txt` and `test_requirements.txt`

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

jkterry1:
I'd propose postponing this as long as we're able to without creating issues. Setup.py is crazy ubiquitously used right now and is useful for new contributors to be able to read.

kir0ul:
And how about removing the duplication of dependencies which are listed both in `setup.py` and `requirements.txt`?
1. Apparently `test_requirements.txt` is not used anywhere and there is already a `testing` entry in `setup.py`.
2. `requirements.txt` seems to be used only to build the docs, so we could make a `docs` entry in `setup.py`.

PR #86, opened 2022-10-27
Use poetry to do package management

vwxyzjn:
This PR uses poetry as the package manager. 

The legacy `setup.py` is removed in favor of the new `pyproject.yaml`. (`setup.py` is getting deprecated — see https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/)


RedTachyon:
Does this make poetry required to use gym? Or will people still be able to just do `pip install -e .`/`pip install gym` in a regular venv and use it normally?

vwxyzjn:
> Does this make poetry required to use gymnasium? Or will people still be able to just do pip install -e ./pip install gymnasium in a regular venv and use it normally?

Yes. 

Note that there is no way to specify `box2d-py` as a dependency. 

Poetry by default uses PEP 517 to install and build dependencies, but `box2d-py` breaks (e.g., try `pip install box2d-py --use-pep517` which will break).  We can not specify `box2d-py` needs to be installed by using non pep 517 in poetry (see https://github.com/python-poetry/poetry/issues/3433), so I currently set to install `box2d-py` manually via

```Dockerfile
RUN poetry run pip install swig==4.0.2
RUN poetry run pip install --no-use-pep517 box2d-py==2.3.8
```

Dmusulas:
Merging this PR would also accept and solve this issue https://github.com/Farama-Foundation/Gymnasium/issues/7

kir0ul:
I'd also remove `requirements.txt` and `test_requirements.txt`, see https://github.com/Farama-Foundation/Gymnasium/issues/7#issuecomment-1252574113.

pseudo-rnd-thoughts:
Could @vwxyzjn or @RedTachyon write a summary of the offline discussion about Poetry. I believe while adding poetry would be a good thing to add, there are several current issues that limit us from doing that

Issue #32, opened 2022-10-01
Vector API - design considerations


RedTachyon:
I started trying to implement a new, more sane vector API. And then I quickly realized that it is, indeed, as messy as I could have expected, so the code will have to wait for some time.

Here I want to dump my thoughts about how this whole thing should/could look so that we have a discussion going.

Main desired outcome: we can use the vector API to easily create envs vectorized through either simple vectorization, or jax vmapping (or any other fast mechanism). This can give us **huge** performance improvements for some envs without relying on additional external libraries. For other envs, we default to Sync/Async/EnvPool?

Current situation: vectorization is only possible via Sync/Async, which is slow af, but very general. EnvPool (not officially supported) only works with some envs, but is faster. Other existing options are generally similar to Sync/Async, with their own quirks (e.g. ray in rllib, or the custom implementation in SB3)

The main complication is wrappers. If an environment provides its own optimized vectorized version, then we can't apply single-env wrappers to it. A nice solution would be an automatic conversion from a `Wrapper` to a `VectorWrapper`, but that seems either very tricky or impossible to do in a general case. Fortunately, many actual wrappers don't need that "general case" treatment.

The hope I see for this is switching to lambda wrappers, at least for some of the existing wrappers. ActionWrappers, ObservationWrappers and RewardWrappers can in principle be stateful, which requires some workarounds to map them over vectorized envs. With lambda wrappers, we can literally just do a map.

An element that I think will be crucial is different levels of optimization - existing third-party environments and wrappers should work exactly the same way, with the clunky subprocess vecenv approach, unless they do a few extra things to opt-in for the improvements.

Another rough edge might be autoreset. Currently this concept is barely present in gym, it's an optional wrapper for single envs, and in that scope it works fine. In a vectorized case, it's more important and a bit more complicated. If we don't have some sort of autoreset by default in vector envs, that makes them borderline useless for many envs (consider cartpole where the first env instance happens to take 10 steps, and the second takes 100 steps - if we only reset after both are terminated, we just lost 45% of the data)

[...]

PR #113, opened 2022-11-04
Support list of options in VectorEnv.reset().

MarcCote:
# Description

Following a [discussion on Discord](https://discord.com/channels/961771112864313344/1029089576398114846/1037714004040237117), this PR changes `VectorEnv.reset(...)` to accept a list of options similar to seeds, i.e. one options dict per sub env. This is also mentioned in issue #32.

## Type of change

# Checklist:

### DIFF
--- a/tests/vector/test_vector_env.py
+++ b/tests/vector/test_vector_env.py
@@ -112,3 +112,62 @@ def test_final_obs_info(vectoriser):
     assert info["final_observation"] == np.array([0]) and info["final_info"] == {
         "action": 3
     }
+
+
+@pytest.mark.parametrize(
+    "vectoriser",
+    (
+        SyncVectorEnv,
+        partial(AsyncVectorEnv, shared_memory=True),
+        partial(AsyncVectorEnv, shared_memory=False),
+    ),
+    ids=["Sync", "Async with shared memory", "Async without shared memory"],
+)
+def test_reset_with_list_of_options(vectoriser):
+    """Tests that the vector environments can be reset with different options."""
+
+    def reset_fn(self, seed=None, options=None):
+        return [0], dict(options=options, **(options or {}))
+
+    def thunk():
+        return GenericTestEnv(reset_fn=reset_fn)
+
+    env = vectoriser([thunk, thunk, thunk])
+    _, infos = env.reset()
+    assert np.array_equal(infos["options"], (None, None, None))
+
+    # Test options broadcasting.
+    options = {"arg1": 123, "arg2": "abc", "arg3": True}
+    _, infos = env.reset(options=options)
+    assert np.array_equal(infos["options"], (options, options, options))
+    assert np.array_equal(infos["arg1"], (123, 123, 123))
+    assert np.array_equal(infos["arg2"], ("abc", "abc", "abc"))
+    assert np.array_equal(infos["arg3"], (True, True, True))
+
+    # When not all options are either list, tuple, or ndarray, default to options broadcasting.
+    options = {"arg1": "123", "arg2": ("a", "b", "c")}
+    _, infos = env.reset(options=options)
+    assert np.array_equal(infos["options"], (options, options, options))
+    assert np.array_equal(infos["arg1"], ("123", "123", "123"))
+    for value in infos["arg2"]:
+        assert value == ("a", "b", "c")
+
+    # When not all options have the same length, default to options broadcasting.
+    options = {"arg1": [1, 2], "arg2": [1, 2, 3]}
+    _, infos = env.reset(options=options)
+    assert np.array_equal(infos["options"], (options, options, options))
+
+    for dtype in (list, tuple, np.array):
+        options = {"arg1": dtype([1, 2, 3]), "arg2": [1, 2, 3]}
+        _, infos = env.reset(options=options)
+        assert np.array_equal(infos["arg1"], options["arg1"])
+        assert np.array_equal(infos["arg2"], options["arg2"])
+
+        # When options is omitted (or None), reuse cached options.
+        _, infos = env.reset()
+        assert np.array_equal(infos["arg1"], options["arg1"])
+        assert np.array_equal(infos["arg2"], options["arg2"])
+
+        _, infos = env.reset(options=None)
+        assert np.array_equal(infos["arg1"], options["arg1"])
+        assert np.array_equal(infos["arg2"], options["arg2"])
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from enum import Enum
 from multiprocessing.connection import Connection
 from multiprocessing.queues import Queue
-from typing import Any, List, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 
@@ -99,6 +99,7 @@ class AsyncVectorEnv(VectorEnv):
         self.num_envs = len(env_fns)
         self.shared_memory = shared_memory
         self.copy = copy
+        self.options = [None] * self.num_envs
 
         # This would be nice to get rid of, but without it there's a deadlock between shared memory and pipes
         dummy_env = env_fns[0]()
@@ -169,16 +170,16 @@ class AsyncVectorEnv(VectorEnv):
 
     def reset_async(
         self,
-        seed: Optional[Union[int, List[int]]] = None,
-        options: Optional[dict] = None,
+        seed: Optional[int] = None,
+        options: Optional[Union[Dict[str, Any], Dict[str, Sequence[Any]]]] = None,
     ):
         """Send calls to the :obj:`reset` methods of the sub-environments.
 
         To get the results of these calls, you may invoke :meth:`reset_wait`.
 
         Args:
-            seed: List of seeds for each environment
-            options: The reset option
+            seed: The reset seed for the first environment. The rest will be reset using [seed+1, seed+num_envs].
+            options: Option information for the environment reset.
 
         Raises:
             ClosedEnvironmentError: If the environment was closed (if :meth:`close` was previously called).
@@ -194,18 +195,47 @@ class AsyncVectorEnv(VectorEnv):
             seed = [seed + i for i in range(self.num_envs)]
         assert len(seed) == self.num_envs
 
+        if options is None:
+            options = self.options
+            pass  # Reuse cached self.options.
+        else:
+            # Check if options is dict[str, Sequence].
+            is_dict_of_sequences = True
+            for values in options.values():
+                is_sequence = isinstance(values, (list, tuple, np.ndarray))
+                if not is_sequence:
+                    is_dict_of_sequences = False
+                    break
+
+                correct_length = len(values) == self.num_envs
+                if not correct_length:
+                    is_dict_of_sequences = False
+                    break
+
+            if is_dict_of_sequences:
+                # Convert options from dict[str, Sequence[Any]] -> Sequence[dict[str, Any]]
+                options = [
+                    dict(zip(options.keys(), values))
+                    for values in zip(*options.values())
+                ]
+            else:
+                # Broadcasting the same options to all sub environments.
+                options = [dict(options) for _ in range(self.num_envs)]
+
+        self.options = options  # Cache options.
+
         if self._state != AsyncState.DEFAULT:
             raise AlreadyPendingCallError(
                 f"Calling `reset_async` while waiting for a pending call to `{self._state.value}` to complete",
                 str(self._state.value),
             )
 
-        for pipe, single_seed in zip(self.parent_pipes, seed):
+        for pipe, single_seed, option in zip(self.parent_pipes, seed, options):
             single_kwargs = {}
             if single_seed is not None:
                 single_kwargs["seed"] = single_seed
-            if options is not None:
-                single_kwargs["options"] = options
+            if option is not None:
+                single_kwargs["options"] = option
 
             pipe.send(("reset", single_kwargs))
         self._state = AsyncState.WAITING_RESET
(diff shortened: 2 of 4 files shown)


RedTachyon:
Two things:
1) I'm not sure if we actually want this feature in the first place. I can see the appeal, but if it's something that we can't reliably implement in all envs, we might not want to include it in the standard API.
2) If we do actually want this implemented, the PR should point at the `new-vector-api` branch since that's where development is happening.


I'm currently at a conference so I'll only be able to properly look at all the new vector stuff next week or so. But we'll have to carefully consider if we want this feature, as well as the option to use a list of seeds which was the motivation for this

MarcCote:
> * I'm not sure if we actually want this feature in the first place. I can see the appeal, but if it's something that we can't reliably implement in all envs, we might not want to include it in the standard API.
My main motivation is to have similar behavior for `seeds` and `options`. Also, it is backward compatible so it shouldn't impact existing envs, I think.

> * If we do actually want this implemented, the PR should point at the `new-vector-api` branch since that's where development is happening.
Sure, I'll make a PR to that branch then.

RedTachyon:
> Also, it is backward compatible so it shouldn't impact existing envs, I think.

There's a bit more nuance here. We intend for the VectorEnv API to be pretty much as generic as the normal Env API, which means that if you get some `VectorEnv` you know nothing else about, you know how you can interact with it. This means that ideally, **every** env implements all the components of the API.

If we include list of options as part of the API (and similarly, list of seeds - we need to reconsider this as well), this means that every env should support this functionality. If some envs support it and some don't, then it shouldn't be in the `VectorEnv` API.

pseudo-rnd-thoughts:
There are two types of vector environments that we are considering
1. Environments are independent from the Vector Environment, this is how the async and sync vector API work that you pass the vector environment functions for all of the environment and these environments run independently. For this case, adding list of seeds and list of options is very easy and I agree it is a feature, we should add.
2. Vectorised environments, custom environment implementations (a cartpole example is provided in the new-vector-api branch) that allows numpy (or jax) parallelised implementations using numpy broadcast functions. The issue is for these environments, these environments generally have a single random number generator used so a list of seeds doesn't make sense. Similarly for list of options however this is more possible to use vector arrays than scalar values. This would require a lot of checking.

Therefore, I view that we have two options
1. list of seed and list of options is included in the type definition of VectorEnv but type 2 vector implementation raise an error (or warning) if list versions are not allowed.
2. We don't include the list versions in VectorEnv however async and sync do include it in their type definition. 

Personally, I can't see a massive difference between the two options outlined above. My guess is that the second option is the one we should use

MarcCote:
That all makes sense and the second option you suggested would work for my use case.

Even for Vectorised environments, I do think having more control over the initialization of the different sub environments is useful. For instance, in cartpole environment, maybe I'd want all my sub-environments to have the same initial state. Currently, you can't do that with a single seed ([cartpole.py#L456](https://github.com/Farama-Foundation/Gymnasium/blob/new-vector-api/gymnasium/envs/classic_control/cartpole.py#L456)).

MarcCote:
Speaking of autoreset, I noticed the `SyncVectorEnv` still have the autoreset logic in its [step function](https://github.com/Farama-Foundation/Gymnasium/blob/new-vector-api/gymnasium/vector/sync_vector_env.py#L130-L134) while `AsyncVectorEnv` does not. Is the roadmap to use the `AutoResetWrapper` instead?

pseudo-rnd-thoughts:
Async does support autoreset just confusingly in the worker which sort of makes sense
https://github.com/Farama-Foundation/Gymnasium/blob/59740038696150523ab7396cf8082e4558199f30/gymnasium/vector/async_vector_env.py#L633

RedTachyon:
So one thing I'm thinking about now - let's say we allow lists of options in some configuration (I'm still not fully convinced either way what configuration that should be - whether it's mandatory, in what types of vector envs etc).

Should it be `dict | list[dict]` or should it be `dict[Value | Iterable[Value]]`? Which realistically means that it's either a list of dictionaries, or a dictionary of numpy arrays.

We do have to keep some performance considerations in mind here, since vector env is the use case where we actually care about things going reasonably fast.

pseudo-rnd-thoughts:
I think that a list of options (`list[dict]`) is both simpler to implement and for users to understand for the async and sync vectors.
For custom Jax environments a `dict[list]` will be the only option however if we follow opinion 2 explained below then I think this will be fine to have a different type than async and sync.

Even this, @MarcCote could you change your implementation to support both `int` and `list[int]` seeds in async and sync vector environments only and add support `dict` and `list[dict]` options in async and sync vector environments?

MarcCote:
> Async does support autoreset just confusingly in the worker which sort of makes sense
> 
> https://github.com/Farama-Foundation/Gymnasium/blob/59740038696150523ab7396cf8082e4558199f30/gymnasium/vector/async_vector_env.py#L633

Gotcha. I opened a proposal about removing the autoreset logic that is baked into the `*VectorEnv` (see https://github.com/Farama-Foundation/Gymnasium/issues/120).

> Should it be `dict | list[dict]` or should it be `dict[Value | Iterable[Value]]`? Which realistically means that it's either a list of dictionaries, or a dictionary of numpy arrays.

I think  `dict | list[dict]` is more flexible and allows for 
```python
    obs, infos = envs.reset(options=[{"task": "task-1-boil", "variation": 1},
                                     {"task": "task-1-freeze", "variation": 12},
                                     {"task": "task-1-melt", "variation": 7}])
```

@pseudo-rnd-thoughts sure, I'll edit the type hints.

MarcCote:
> Even this, @MarcCote could you change your implementation to support both `int` and `list[int]` seeds in async and sync vector environments only and add support `dict` and `list[dict]` options in async and sync vector environments?

Actually that's already the case for Sync/Async vector env. I changed the type hints of the `VectorEnv.reset`. Let me know if that's what you meant.

MarcCote:
@RedTachyon if you have a workaround for resetting sub envs with different configs let me know.

tests/vector/test_vector_env.py
```diff
@@ -112,3 +112,34 @@ def thunk():
...
+@pytest.mark.parametrize(
+    "vectoriser",
+    (
+        SyncVectorEnv,
+        partial(AsyncVectorEnv, shared_memory=True),
+        partial(AsyncVectorEnv, shared_memory=False),
+    ),
+    ids=["Sync", "Async with shared memory", "Async without shared memory"],
+)
+def test_reset_with_list_of_options(vectoriser):
```
pseudo-rnd-thoughts, line 126: Looks good, could you add another test that on autoreset, the options will be used again

pseudo-rnd-thoughts (changes requested):
Looks good to me, could you implement the test for autoreset using the option

tests/vector/test_vector_env.py
MarcCote: Hmm, at the moment, autoreset doesn't reused the provided `options` nor `seeds`.

RedTachyon:
Keep in mind that `dict[Sequence]` and `Sequence[dict]` are pretty much equivalent, but they can have a performance implication for jax envs. I'm leaning towards the former, if nothing else then for the sake of consistency, since that's how we report `info` in steps.

What is the most tricky part here is how we define the API. Ideally, there should be no ambiguity - if a certain single env accepts `options={"start": 5}`, then the corresponding vector env should accept `options={"start: [5, 2, None, 0]}`, where `None` indicates "we're not passing any value to this env".

Can we ensure that at least for the built-in envs with custom vectorization (i.e. numpy-based classic control, and jitted jax envs) this will work? If yes, I think I'd say it's best if proceed as follows:

1. The `VectorEnv` definition allows `seed: int | Sequence[int]` (check if Sequence or Iterable is more fitting)
2. The `VectorEnv` definition allows `options: dict[str, Value | Sequence[Value]]` where Value is a generic typevar (the type hint won't actually be very useful for type checkers, but it describes the interface). We could also restrict it to be `options: dict[str, Value] | dict[str, Sequence[Value]]`. In either case, there's a rough edge for situations where the option value is itself an iterable, but I'm not sure how to handle that yet - we can leave it up to env developers. 
3. Internally, if it's possible, we write our vector code to primarily handle the `Sequence` variant, and the `int`/`Value` variants get trivially promoted to sequences by making `n` copies. 
4. We write all our internal envs so that, to the extent that is possible, all these functionalities work and give the same results regardless of the vectorization (both for seeds and options)*
5. We provide extra env checkers thing which (a) checks if the VectorEnv fulfills general API requirments (like the current env checker), and (b) checks if it behaves identically to a standard Sync/Async vectorization of a given single environment


* I'm considering relaxing this requirement. For full reproducibility, should it be enough to state "We use Cartpole-v1", or should authors state "We use Cartpole-v1 with X vectorization"? If the latter is fine, it may be alright if there are small differences (e.g. numpy/jax-vectorized envs not accepting multiple seeds)


Does this make sense? Can anyone see any other edge cases that are not covered here? Our internal envs should be written with full complicance to all the standards we want, but for third-party envs we should probably treat users as adults, inform them if something is iffy (via the vectorenv checker), but let them do their thing if they want.

tests/vector/test_vector_env.py
RedTachyon: I don't think it should reuse seeds, that would lead to invalid trainings (with basically fixed initial states). Reusing options makes a lot of sense, we should probably cache this, either on the environment side, or inside `VectorEnv` by default

MarcCote:
> Keep in mind that dict[Sequence] and Sequence[dict] are pretty much equivalent, but they can have a performance implication for jax envs. I'm leaning towards the former, if nothing else then for the sake of consistency, since that's how we report info in steps.

Based on our discussion in #120, I agree with using `dict[Sequence]`. It should be trivial to conceptually convert it to a `Sequence[dict]` and loop through it to initialize each sub env in sync/async vectorization. I can implement that for this PR.

> I'm considering relaxing this requirement. For full reproducibility, should it be enough to state "We use Cartpole-v1", or should authors state "We use Cartpole-v1 with X vectorization"? If the latter is fine, it may be alright if there are small differences (e.g. numpy/jax-vectorized envs not accepting multiple seeds)

Correct me if I'm wrong but at the moment, the initial state from sync/async of Cartpole-v1 is different from the custom version. The former uses `num_envs` different RNGs while the custom version uses a single RNG. So, it seems users would need to mention which vectorization they used to be fully reproducible, right? One solution would be to keep `num_envs` RNGs in the custom vectorization and concat the vectors to build the state, but that has big impact on speed.

pseudo-rnd-thoughts:
This is true, but given the random number generators are random then you shouldn't be able to tell the difference between a single RNG and multiple RNG unless users provide the initial seed values. Therefore, I don't think this should be an issue

MarcCote:
> Based on our discussion in #120, I agree with using `dict[Sequence]`. It should be trivial to conceptually convert it to a `Sequence[dict]` and loop through it to initialize each sub env in sync/async vectorization. I can implement that for this PR.

What's the best way to differentiate between `Dict[str, Any]` and `Dict[str, Sequence[Any]]` ?

For instance, how to know if I need to broadcast the options dicts?

Given `{'arg1': "12", 'arg2': "ab"}`, should it be
```
envs[0].reset(options={'arg1': "12", 'arg2': "ab"})
envs[1].reset(options={'arg1': "12", 'arg2': "ab"})
```
or 
```
envs[0].reset(options={'arg1': "1", 'arg2': "a"})
envs[1].reset(options={'arg1': "2", 'arg2': "b"})
```

Given `{'arg1': [1, 2], 'arg2': ['a', 'b']}`, should it be
```python
envs[0].reset(options={'arg1': [1, 2], 'arg2': ['a', 'b']})
envs[1].reset(options={'arg1': [1, 2], 'arg2': ['a', 'b']})
```
or
```python
envs[0].reset(options={'arg1': 1, 'arg2': 'a'})
envs[1].reset(options={'arg1': 2, 'arg2': 'b'})
```

I think distinguishing between `List[Dict]` and `Dict[str, Sequence]` is easier.

pseudo-rnd-thoughts:
If we are using `dict[str, sequence]` which I am still unsure of doing for async and sync due to null values then I think we just enforce the use of `tuple`, `list` or `np.ndarray` as the sequence.

Personally, I think that `sequence[dict[str, Any]]` is easier for async and sync for users to understand and you don't need to consider null values while for custom environments they can choice the additional option style.

pseudo-rnd-thoughts:
@MarcCote How is adding the option autoreset + tests going?

MarcCote:
> If we are using `dict[str, sequence]` which I am still unsure of doing for async and sync due to null values then I think we just enforce the use of `tuple`, `list` or `np.ndarray` as the sequence.

Note that even if the type is enforced, there will be ambiguity.

Assuming, 
```python
envs = gym.make_vec('MyEnV-v1', num_envs=2)`
envs.reset(options={'arg1': [1, 2], 'arg2': ['a', 'b']}
```
should `Async/SyncVectorEnv.reset(...)` internally call

```python
# Same options dict to all sub envs.
self.envs[0].reset(options={'arg1': [1, 2], 'arg2': ['a', 'b']})
self.envs[1].reset(options={'arg1': [1, 2], 'arg2': ['a', 'b']})
```
or

```python
# Split options dict amongst the sub envs.
self.envs[0].reset(options={'arg1': 1, 'arg2': 'a'})
self.envs[1].reset(options={'arg1': 2, 'arg2': 'b'})
```

> Personally, I think that `sequence[dict[str, Any]]` is easier for async and sync for users to understand and you don't need to consider null values while for custom environments they can choice the additional option style.

I think this ^ is the viable option if we intend to do type checking. @RedTachyon

pseudo-rnd-thoughts:
It would be option two for the `dict[str, Sequence]`

MarcCote:
> It would be option two for the `dict[str, Sequence]`

But what if the user wanted to same options for all sub envs? How would they do that, if we split Sequence among subenvs?

MarcCote:
@RedTachyon @pseudo-rnd-thoughts I've updated the PR based on our discussion.

RedTachyon:
I'm gonna close this, I think the conclusion from discord discussions was that this isn't really feasible to implement in a sane way, at least for now

Issue #126, opened 2022-11-12
Calling play() from gymnasium.utils.play will error if render_mode is not in ["rgb_array", "rgb_array_list]"


axb2035:
### Describe the bug

I had made an assumption that from gymnasium.utils.play import play would support text rendering. Seems that was a bad assumption and that only pygame rendering is supported i.e. render_mode in ["rgb_array", "rgb_array_list]". 

However, if you pass a mode not in that list e.g. render_mode="human", it is captured and sends a message to the logger. However, it the script doesn't exit and it crashes shortly after when it tries to obtain the video size with `assert rendered is not None and isinstance(rendered, np.ndarray)`

### Code example

```shell
import gymnasium as gym
from gymnasium.utils.play import play

mapping = {"2": 1,  # Down.
           "4": 0,  # Left.
           "6": 2,  # Right.
           "8": 3,  # Up.
           }

play(gym.make('FrozenLake-v1', render_mode='human'), keys_to_action=mapping)
```

### System info

Gymnasium installed with pip
gymnasium.__version__ '0.26.3'
Python 3.10.6
Windows 10.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

Markus28:
I'll look into it

axb2035:
Thanks @RedTachyon and @Markus28.

PR #132, opened 2022-11-15
Fix `play` assertion error

Markus28:
This PR addresses https://github.com/Farama-Foundation/Gymnasium/issues/126
I just replaced the error message with an actual exception.
I guess `logger.error` can be a somewhat confusing name for a function that only does logging. Searching through the repo, https://github.com/Farama-Foundation/Gymnasium/blob/5674a52d4c4d7cca9f169db561622b8d3e49bd65/gymnasium/utils/save_video.py#L86 seems to have the same problem.

### DIFF
--- a/gymnasium/utils/play.py
+++ b/gymnasium/utils/play.py
@@ -51,7 +51,7 @@ class PlayableGame:
             zoom: If to zoom in on the environment render
         """
         if env.render_mode not in {"rgb_array", "rgb_array_list"}:
-            logger.error(
+            raise ValueError(
                 "PlayableGame wrapper works only with rgb_array and rgb_array_list render modes, "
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Can we have a short test that this works as expected?

axb2035:
BTW: logger does send a error to the console as well so it is providing some feedback on the error to the user running it. Also, if for some reason you were running it remotely you might not have a console, and want to see the output of logs if it fails. Though I appreciate this is probably an edge case.

pseudo-rnd-thoughts:
In `tests` we have a large collection of tests for the whole project using pytest. 
In `tests/utils/test_play.py`, add option 2 in there

Issue #127, opened 2022-11-13
[Proposal] Seed for `generate_random_map()` in FrozenLake?


kir0ul:
### Proposal

Could be useful to be able to pass a seed to the `generate_random_map()` function in FrozenLake.

### Motivation

Being able to get reproducible generated maps.

### Pitch

Being able to call `generate_random_map(size=8, seed=123)`.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
This seems like a critical issue, could you make a PR that solves this. 
Not sure, if we want the `__init__` to contain a `map_seed` parameter or for the map data to be set in `reset` and use the `seed` value

PR #139, opened 2022-11-17
[FrozenLake] Add seed in random map generation

kir0ul:
# Description

This PR adds a seed argument to the `generate_random_map()` function in FrozenLake.

Fixes https://github.com/Farama-Foundation/Gymnasium/issues/127.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/envs/toy_text/frozen_lake.py
+++ b/gymnasium/envs/toy_text/frozen_lake.py
@@ -9,6 +9,7 @@ import gymnasium as gym
 from gymnasium import Env, spaces, utils
 from gymnasium.envs.toy_text.utils import categorical_sample
 from gymnasium.error import DependencyNotInstalled
+from gymnasium.utils import seeding
 
 LEFT = 0
 DOWN = 1
@@ -51,7 +52,7 @@ def is_valid(board: List[List[str]], max_size: int) -> bool:
     return False
 
 
-def generate_random_map(size: int = 8, p: float = 0.8) -> List[str]:
+def generate_random_map(size: int = 8, p: float = 0.8, seed: int = None) -> List[str]:
     """Generates a random valid map (one that has a path from start to goal)
 
     Args:
@@ -64,9 +65,13 @@ def generate_random_map(size: int = 8, p: float = 0.8) -> List[str]:
     valid = False
     board = []  # initialize to make pyright happy
 
+    rng = np.random
+    if seed is not None:
+        rng, seed = seeding.np_random(seed)
+
     while not valid:
         p = min(1, p)
-        board = np.random.choice(["F", "H"], (size, size), p=[p, 1 - p])
+        board = rng.choice(["F", "H"], (size, size), p=[p, 1 - p])
         board[0][0] = "S"
         board[-1][-1] = "G"
         valid = is_valid(board, size)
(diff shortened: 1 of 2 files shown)


gymnasium/envs/toy_text/frozen_lake.py
```diff
@@ -64,9 +65,13 @@ def generate_random_map(size: int = 8, p: float = 0.8) -> List[str]:
     valid = False
     board = []  # initialize to make pyright happy
 
+    rng = np.random
```
pseudo-rnd-thoughts, line 68: Should this not be `np_random = seeding.np_random(seed)`

tests/envs/test_env_implementation.py
```diff
@@ -126,6 +126,16 @@ def test_frozenlake_dfs_map_generation(map_size: int):
     raise AssertionError("No path through the frozenlake was found.")
 
 
+@pytest.mark.parametrize("map_size, seed", [(5, 123), (10, 42), (16, 987)])
+def test_frozenlake_map_generation_with_seed(map_size: int, seed: int):
+    map1 = generate_random_map(size=map_size, seed=seed)
+    map2 = generate_random_map(size=map_size, seed=seed)
+    assert map1 == map2
+    map1 = generate_random_map(size=map_size)
```
pseudo-rnd-thoughts, line 134: I would add this as there are cases where this could fail. 
You could replace it with a test that `seed` and `seed+1` have different maps

gymnasium/envs/toy_text/frozen_lake.py
RedTachyon: I think this is meant to be essentially
```python
if seed is None:
    rng = np.random
else:
    rng, _ = seeding.np_random(seed)
```

The problem here is that `np_random` is annoyingly not an equivalent of `np.random` since `np.random` is essentially deprecated for this usage. Instead it should probably just be

```python
rng = np.default_rng(seed)
```

We might not need to use the whole `seeding.np_random` machinery, but if we want to, we can do `seeding.np_random(seed)` too, since it should handle `None` properly

pseudo-rnd-thoughts: Yes, this is what I meant
The if statement can be simplified to `np_random = seeding.np_random(seed)` then `np_random.choice()`

kir0ul: Updated as you suggested!

tests/envs/test_env_implementation.py
kir0ul: Indeed... updated as you suggested.

gymnasium/envs/toy_text/frozen_lake.py
```diff
@@ -51,22 +52,25 @@ def is_valid(board: List[List[str]], max_size: int) -> bool:
     return False
 
 
-def generate_random_map(size: int = 8, p: float = 0.8) -> List[str]:
+def generate_random_map(size: int = 8, p: float = 0.8, seed: int = None) -> List[str]:
```
pseudo-rnd-thoughts, line 55: `Optional[int]`

pseudo-rnd-thoughts (approved):
Could you fix the type hint then ready to merge

gymnasium/envs/toy_text/frozen_lake.py
kir0ul: Oops... fixed!
Strange Pyright didn't say anything :thinking:

PR #141, opened 2022-11-18
Change discrete dtype to np.int64

pseudo-rnd-thoughts:
The `Discrete` class definition return type is `int` while the space `dtype` is `np.int64`. 
This PR makes the `Discrete` type consistent with `np.int64`.

### DIFF
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -1,14 +1,14 @@
 """Implementation of a space consisting of finitely many elements."""
 from __future__ import annotations
 
-from typing import Any, Iterable, Mapping
+from typing import Any, Iterable, Mapping, Sequence
 
 import numpy as np
 
 from gymnasium.spaces.space import MaskNDArray, Space
 
 
-class Discrete(Space[int]):
+class Discrete(Space[np.int64]):
     r"""A space consisting of finitely many elements.
 
     This class represents a finite subset of integers, more specifically a set of the form :math:`\{ a, a+1, \dots, a+n-1 \}`.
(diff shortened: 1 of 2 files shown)


younik:
Side question: do we need `np.int64` instead of `np.int32`?

I don't have an example in mind where you need int64 for Discrete; while usually int32 is used

pseudo-rnd-thoughts:
Agree, this was my initial thoughts as well but the numpy random `integers`  and `choice` functions return `np.int64` and the `dtype` was `np.int64` already.

younik:
> Agree, this was my initial thoughts as well but the numpy random `integers` and `choice` functions return `np.int64` and the `dtype` was `np.int64` already.

If we want to do this, we can change it by passing `dtype` to `integers` and an explicit arange to `choice`

pseudo-rnd-thoughts:
@younik I have decided not to change it for three reasons
1. When you convert python int to numpy, its type is `np.int64`
2. numpy functions default `dtype` for integers is `int64`
3. `MultiDiscrete` uses `np.int64` 

Therefore, Im going to keep it to `np.int64` and it minimise the difference for backward compatibility.

gymnasium/spaces/discrete.py
```diff
@@ -21,9 +21,9 @@ class Discrete(Space[int]):
 
     def __init__(
         self,
-        n: int,
+        n: int | np.int64,
```
RedTachyon, line 24: Can't we use one of the general numpy dtypes to allow int/int64/int32/whatever other int? The annotation could be `int | np.integer`, then I think it would accept `int`, `int64`, `int32`, `uint8` and all of that stuff. While it does make sense to narrow down the return type, this would throw redundant errors (since all of those can be cast to `int64` anyways)

RedTachyon: Also in L24

pseudo-rnd-thoughts: Done

RedTachyon (approved):

PR #150, opened 2022-11-20
Versioned Action wrappers which supports jumpy

gianlucadecola:
# Description

This PR add support for jumpy to action wrappers `RescaleAction` and `ClipAction`.
Behaviour is the same of old wrappers.

### New features  
* Name versioning
* Jumpy

### TODOS:

- [x] moved the wrappers in experimental.wrappers
- [x] have all of the lambda actions wrappers in lambda_action.py not separate files
- [x] clip action and rescale action inherit from LambdaActionV0
- [ ] updated the docs in docs/api/experimental.md and docs/api/experimental/wrappers.md


gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -28,3 +35,152 @@ def __init__(
     def action(self, action: ActType) -> Any:
         """Apply function to action."""
         return self.func(action)
+
+
+class LambdaCompositeActionV0(LambdaActionV0):
```
pseudo-rnd-thoughts, line 40: @RedTachyon Are you sure that this is the best way? Will users use this / does it provide a helpful abstraction that we will actually use?

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -28,3 +35,152 @@ def __init__(
...
+        >>> env = ExampleEnv(action_space=Dict(body=Dict(head=Box(0.0, 10.0, (1,))), left_arm=Discrete(4), right_arm=Box(0.0, 5.0, (1,))))
+        >>> env.action_space
+        Dict(body: Dict(head: Box(0.0, 10.0, (1,), float32)), left_arm: Discrete(4), right_arm: Box(0.0, 5.0, (1,), float32))
+        >>> args = {"right_arm": (0, 2), "body": {"head": (0, 3)}}
+        >>> env = ClipActionV0(env, args)
+        >>> env.action_space
+        Dict(body: Dict(head: Box(0.0, 3.0, (1,), float32)), left_arm: Discrete(4), right_arm: Box(0.0, 2.0, (1,), float32))
+    """
+
+    def __init__(self, env: gym.Env, args: CompositeParameterType):
```
pseudo-rnd-thoughts, line 122: I feel there should a way of specifying the actual type of the `CompositeParameterType`

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -28,3 +35,152 @@ def __init__(
...
+        """Constructor for the clip action wrapper.
+
+        Args:
+            env (Env): The environment to wrap
+            args (TreeParameterType): The arguments for clipping the action space
+        """
+        action_space = self._transform_space(env, args)
+
+        super().__init__(
```
pseudo-rnd-thoughts, line 131: We could simplify this partially through inheriting from `ActionWrapper` and then just implementing the `action` function which to users to simpler and easier to debug. 
While I agree that we can use the do this in the end we have in terms of inheritance, `ClipAction -> LambdaCompositeAction -> LambdaAction -> ActionWrapper -> Wrapper`

RedTachyon (changes requested):
Not comprehensive review yet, but there's a bunch of issues

gymnasium/dev_wrappers/lambda_action.py
```diff
@@ -28,3 +35,152 @@ def __init__(
...
+            + (old_high - old_low) * ((action - new_low) / (new_high - new_low)),
+            old_low,
+            old_high,
+        )
+
+    @staticmethod
+    def _make_scale_args(
+        env: gym.Env, args: CompositeParameterType
+    ) -> CompositeParameterType:
+        return make_scale_args(env.action_space, args, make_scale_args)
```
RedTachyon, line 186: What's happening here and why? Why does the function need to take itself as an argument?

gianlucadecola:
As per discussion had this PR now only rewrite old action wrappers with jumpy support and changes naming convention by including versioning :+1:

gymnasium/wrappers/clip_action.py
```diff
@@ -25,6 +26,10 @@ def __init__(self, env: gym.Env):
         Args:
             env: The environment to apply the wrapper
         """
+        deprecation(
```
pseudo-rnd-thoughts, line 29: Sorry about this, I don't think we want deprecation warnings yet. The way that Im thinking about it, is v0.27 includes the new wrappers, v0.28 includes the new vector env and vector wrappers then v0.29 will include the deprecation warnings. See https://github.com/Farama-Foundation/Gymnasium/pull/159 for more details on how we could do it

setup.py
```diff
@@ -87,6 +87,7 @@ def get_version():
     include_package_data=True,
     install_requires=[
         "numpy >= 1.21.0",
+        "brax-jumpy >= 0.1.0",
```
pseudo-rnd-thoughts, line 90: We have moved the module to `jax-jumpy`

gymnasium/dev_wrappers/rescale_action.py
```diff
@@ -0,0 +1,80 @@
...
+            min_action (float, int or np.ndarray): The min values for each action. This may be a numpy array or a scalar.
+            max_action (float, int or np.ndarray): The max values for each action. This may be a numpy array or a scalar.
+        """
+        assert isinstance(
+            env.action_space, spaces.Box
+        ), f"expected Box action space, got {type(env.action_space)}"
+        assert np.less_equal(min_action, max_action).all(), (min_action, max_action)
+
+        super().__init__(env)
+        self.min_action = (
```
pseudo-rnd-thoughts, line 52: In my understanding, we don't need to modify the action space as the space specifies the valid input for inputs not outputs.

gymnasium/dev_wrappers/clip_action.py
```diff
@@ -0,0 +1,41 @@
+"""Wrapper for clipping actions within a valid bound."""
+import jumpy as jp
+
+import gymnasium as gym
+from gymnasium import spaces
+from gymnasium.core import ActType
+
+
+class ClipActionV0(gym.ActionWrapper):
```
pseudo-rnd-thoughts, line 9: @RedTachyon For the vectorising method, do you want wrappers to inherit from `Lambda` wrapper version?

gymnasium/dev_wrappers/clip_action.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    def __init__(self, env: gym.Env):
+        """A wrapper for clipping continuous actions within the valid bound.
+
+        Args:
+            env: The environment to apply the wrapper
+        """
+        assert isinstance(env.action_space, spaces.Box)
+        super().__init__(env)
```
pseudo-rnd-thoughts, line 30: Technically the action space can be expanded to `Box(-np.inf, npinf, env.action_space.shape)` as the action space is the valid inputs rather than output.

gymnasium/dev_wrappers/clip_action.py
```diff
@@ -0,0 +1,41 @@
...
+        # Executes the action np.array([1.0, 1.0, -1.0, 0]) in the base environment
+    """
+
+    def __init__(self, env: gym.Env):
+        """A wrapper for clipping continuous actions within the valid bound.
+
+        Args:
+            env: The environment to apply the wrapper
+        """
+        assert isinstance(env.action_space, spaces.Box)
```
pseudo-rnd-thoughts, line 29: This is optional testing addition, but the clip bounds should be within the `env.action_space`

gymnasium/dev_wrappers/rescale_action.py
```diff
@@ -0,0 +1,80 @@
...
+            np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + min_action
+        )
+        self.max_action = (
+            np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + max_action
```
pseudo-rnd-thoughts, line 56: Use `np.full`

gymnasium/dev_wrappers/rescale_action.py
```diff
@@ -0,0 +1,80 @@
+"""Wrapper for rescaling actions to within a max and min action."""
+from typing import Union
+
+import jumpy as jp
+import numpy as np
```
pseudo-rnd-thoughts, line 5: I'm not sure we need to bother now, but I think that we should look to only use jumpy where we add the necessary functions

gymnasium/wrappers/clip_action.py
gianlucadecola: No problem :wink: , removed in https://github.com/Farama-Foundation/Gymnasium/pull/150/commits/c869086369e7b95b068585f3ef9bec903d16b433

setup.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/150/commits/4a33977ee593feac1818d82e02721cd599a00eb9

gymnasium/dev_wrappers/rescale_action.py
gianlucadecola: https://github.com/Farama-Foundation/Gymnasium/pull/150/commits/194ef34365a0a8f8ca6a2617122c106609a3598a

gymnasium/dev_wrappers/rescale_action.py
gianlucadecola: It is used in https://github.com/Farama-Foundation/Gymnasium/blob/54b99ac8ac2194937e58e45efddf5d06922069e1/gymnasium/dev_wrappers/rescale_action.py#L80 maybe I am missing something?

gymnasium/dev_wrappers/clip_action.py
gianlucadecola: I see :+1: so no problem in changing the action space output with respect to the old `clip_action` wrapper?

gymnasium/dev_wrappers/rescale_action.py
gianlucadecola: Same question as here: https://github.com/Farama-Foundation/Gymnasium/pull/150#discussion_r1032955271

gymnasium/dev_wrappers/rescale_action.py
pseudo-rnd-thoughts: Good point, I was more looking at the code for np.less_equal in the constructor so I think this might be fine

gymnasium/dev_wrappers/clip_action.py
pseudo-rnd-thoughts: I don't think so

gymnasium/dev_wrappers/rescale_action.py
pseudo-rnd-thoughts: Yes no issue

gianlucadecola: Updated both for `rescale_action` and for `clip_action` here https://github.com/Farama-Foundation/Gymnasium/pull/150/commits/f616199bd40309e2d5eac1f276975bce04537c49

pseudo-rnd-thoughts:
With the experimental PR merged, could you 
* moved the wrappers in `experimental.wrappers`
* have all of the lambda actions wrappers in `lambda_action.py` not separate files
* clip action and rescale action inherit from `LambdaActionV0`
* updated the docs in `docs/api/experimental.md` and `docs/api/experimental/wrappers.md`

Then good to merge

pseudo-rnd-thoughts (approved):

Issue #92, opened 2022-10-28
[Proposal] Clean up the `gym.make` function


RedTachyon:
### Proposal

Clean up `registration.py` HumanRendering/RenderCollection automatic wrapping

https://github.com/Farama-Foundation/Gymnasium/blob/6f94a2a0bf46b73f809236a150c5387a439741bc/gymnasium/envs/registration.py#L588-L629

This chunk has 5 levels of if-else indentation, with pretty opaque logic and seems to conflate two wrappers together. I don't have the bandwidth to do it right now, but it desperately needs a cleanup.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
In addition, the env spec has a number of parameters that I don't believed are used by anyone, `reward_range`, etc. 
I think we should slim down a lot of this, the problem is backward compatibility.

PR #157, opened 2022-11-25
ENH: clean make function

younik:
Addressing https://github.com/Farama-Foundation/Gymnasium/issues/92
This also introduces a change: now `UnsupportedMode` is raised if 'render_mode` is invalid (previously we showed only a warning). 

Side question: are the make `overloading` useful? It seems they introduce a bit of noise in the file just for having a more specific typing; a general return type `Env` works fine too

### DIFF
--- a/gymnasium/envs/registration.py
+++ b/gymnasium/envs/registration.py
@@ -422,7 +422,9 @@ def _check_spec_register(spec: EnvSpec):
 
 def _check_metadata(metadata_: dict):
     if not isinstance(metadata_, dict):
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (commented):
Personally, I like this code a lot more than the old code to simplify the make function and can't see any major issues
However, as this is Ariel's PR, I will let him approve it

RedTachyon (approved):
This looks good, much better than the current version, thanks!

That being said, I'm not sure if it goes far enough, at least it's not super transparent to me how the render modes are changed in what situation. This might require some breaking changes, so we can merge this and see if we want to do more changes in the future.

PR #160, opened 2022-11-26
Update docstrings for improved documentation

axb2035:
# Description

Updates the toy_text docstrings to have better consitency and more information. As discussed on Discord. No dependcies required.

No related issue.

## Type of change

### Screenshots

n/a

# Checklist:

### DIFF
--- a/gymnasium/envs/toy_text/cliffwalking.py
+++ b/gymnasium/envs/toy_text/cliffwalking.py
@@ -18,39 +18,60 @@ LEFT = 3
 
 class CliffWalkingEnv(Env):
     """
-    This is a simple implementation of the Gridworld Cliff
-    reinforcement learning task.
+    Cliff walking involves crossing a gridworld from start to goal while avoiding falling off a cliff.
 
-    Adapted from Example 6.6 (page 106) from [Reinforcement Learning: An Introduction
-    by Sutton and Barto](http://incompleteideas.net/book/bookdraft2018jan1.pdf).
+    ## Description
+    The game starts with the player at location [3, 0] of the 4x12 grid world with the
+    goal located at [3, 11]. If the player reaches the goal the episode ends.
+
+    A cliff runs along [3, 1..10]. If the player moves to a cliff location it
+    returns to the start location.
+
+    The player makes moves until they reach the goal.
+
+    Adapted from Example 6.6 (page 132) from Reinforcement Learning: An Introduction
+    by Sutton and Barto [<a href="#ref">1</a>].
 
     With inspiration from:
     [https://github.com/dennybritz/reinforcement-learning/blob/master/lib/envs/cliff_walking.py](https://github.com/dennybritz/reinforcement-learning/blob/master/lib/envs/cliff_walking.py)
 
-    ## Description
-    The board is a 4x12 matrix, with (using NumPy matrix indexing):
-    - [3, 0] as the start at bottom-left
-    - [3, 11] as the goal at bottom-right
-    - [3, 1..10] as the cliff at bottom-center
+    ## Action Space
+    The action shape is `(1,)` in the range `{0, 3}` indicating
+    which direction to move the player.
 
-    If the agent steps on the cliff, it returns to the start.
-    An episode terminates when the agent reaches the goal.
+    - 0: Move up
+    - 1: Move right
+    - 2: Move down
+    - 3: Move left
 
-    ## Actions
-    There are 4 discrete deterministic actions:
-    - 0: move up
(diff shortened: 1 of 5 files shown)


gymnasium/envs/toy_text/cliffwalking.py
```diff
@@ -18,39 +18,60 @@
...
+    The action shape is `(1,)` in the range `{0, 3}` indicating
+    which direction to move the player.
+
+    - 0: Move up
+    - 1: Move right
+    - 2: Move down
+    - 3: Move left
+
+    ## Observation Space
+    There are 3 x 12 + 1 possible states. The player cannot be at the cliff, nor at
```
pseudo-rnd-thoughts, line 48: This is outside the scope of this pr, but should we be interested in adding an output with the coordinates of the agent. This could be an additional constructor argument and would change the output input, but would be easier to debug and as a demonstration to new users than an integer representing the state.

gymnasium/envs/toy_text/cliffwalking.py
```diff
@@ -18,39 +18,60 @@
...
+    Each time step incurs -1 reward, unless the player stepped into the cliff,
+    which incurs -100 reward.
+
+    ## Episode End
+    The episode terminates when the player enters state `[47]` (location [3, 11]).
+
+    ## Information
+
+    `step()` and `reset()` return a dict with the following keys:
+    - "p" - transition proability for the state.
```
pseudo-rnd-thoughts, line 72: Given that the transition probability is always 1.0, what information does the transition probability actually help with?

gymnasium/envs/toy_text/taxi.py
```diff
@@ -24,89 +24,113 @@
...
+    `step()` and `reset()` return a dict with the following keys:
+    - p - transition proability for the state.
+    - action_mask - if actions will cause a transition to a new state.
 
-    As Taxi's initial state is a stochastic, the "p" key represents the probability of the
-    transition however this value is currently bugged being 1.0, this will be fixed soon.
-    As the steps are deterministic, "p" represents the probability of the transition which is always 1.0
+    As taxi is not stochastic, the transition probability is always 1.0. Implementing
+    a transitional probability in line with the Dietterich paper ('The fickle taxi task')
+    is a TODO.
```
pseudo-rnd-thoughts, line 125: Should this TODO be here. Do we need someone to implement this?

gymnasium/envs/toy_text/cliffwalking.py
axb2035: I agree it would be easier to understand and remove a layer of complexity for new users. It took me a while to figure out how it works. FYI: taxi and frozen lake use the same approach.

gymnasium/envs/toy_text/taxi.py
axb2035: I've put my hand up to implement transitional probablities in taxi it once I finish the doc review. Replaces the "bugged will be fixed soon" comment :)

gymnasium/envs/toy_text/cliffwalking.py
axb2035: Agree, not a lot. It's there for completeness. Taxi, frozen lake and cliff walking have proability returned in their information, though only frozen lake actually has it implemented (see next comment regarding taxi).

For cliff walking, we could implement transitional probabilities similar to frozen lake / taxi. If we are not then suggest we change the info returned to `{}`. Though that may break some people's code...

pseudo-rnd-thoughts: I think it is more interesting to add the transition probability, we can do this in a future PR.
Could you make an issue outlining the problem and proposed solution. 
For the taxi environment, I would only apply randomness on the movement actions and not pick up and drop off

gymnasium/envs/toy_text/cliffwalking.py
pseudo-rnd-thoughts: Ok, I don't think, this can be done in a future PR, possibly in the same PR as the transition probability PR

gymnasium/envs/toy_text/cliffwalking.py
axb2035: Issue https://github.com/Farama-Foundation/Gymnasium/issues/161 created. It doesn't include the suggestion to output  coordinates of the agent for Cliff Walking, Frozen Lake and Taxi to keep the changes discrete. Will add the output co-ordinate enhancement proposal at a later date.

pseudo-rnd-thoughts (approved):

PR #164, opened 2022-11-28
Flatten space docstring note about unflattening samples

### LINKED ISSUE #102: [Question] Flatten Discrete box potentially problematic
### Question

Flattening Discrete space to Box space may be problematic. The flatten wrapper converts Discrete to Box as a one-hot encoding. Suppose the original space is Discrete(3), then:

```
0 maps to [1, 0, 0]
1 maps to [0, 1, 0]
3 maps to [0, 0, 1]
```

When we sample the action space for random actions, it samples the Box, which can produce any of the eight combination of 0s and 1s in a three-element array, namely:

```
[0, 0, 0],
[0, 0, 1], *
[0, 1, 0], *
[0, 1, 1],
[1, 0, 0], *
[1, 0, 1],
[1, 1, 0],
[1, 1, 1]
```

Only three of these eight that I’ve starred are useable in the strict sense of the mapping. The unflatten function for a Discrete space uses `np.nonzero(x)[0][0]`, and here’s at table of what the above arrays map to:

```
+ ------------------ + ---------------- + --------------------------------------------- +
| In Flattened Space | np.nonzero(x)[0] | np.nonzero(x)[0][0] (aka discrete equivalent) |
+ ------------------ + ---------------- + --------------------------------------------- +
| 0, 0, 0            | Error            | Error                                         |
| 0, 0, 1            | [2]              | 2                                             |
| 0, 1, 0            | [1]              | 1                                             |
| 0, 1, 1            | [1, 2]           | 1                                             |
| 1, 0, 0            | [0]              | 0                                             |
| 1, 0, 1            | [0, 2]           | 0                                             |
| 1, 1, 0            | [0, 1]           | 0                                             |
| 1, 1, 1            | [0, 1, 2]        | 0                                             |
+ ------------------ + ---------------- + --------------------------------------------- +
```

#### Implications
Obviously, [0, 0, 0] will fail because there is no nonzero.
Importantly, only one eighth of the random samples will map to 2. One fourth will map to 1, and one half will map to 0. This has some important implications on exploration, especially if action 2 is the “correct action” throughout much of the simulation. I'm very curious why I have not seen this come up before. This type of skewing in the random sampling can have major implications in the way the algorithm explores and learns, and the problem is exacerbated when `Discrete(n), n is large`. Am I missing something here?

#### Solution
This is unique to Discrete spaces. Instead of mapping to a one-hot encoding we could just map to a box of a single element with the appropriate range. Discrete(n) maps to Box(0, n-1, (1,), int) instead of Box(0, 1, (n,), int).

rusu24edward:
# Description

Just added a note about unflattening a sample from a flattened space.

Fixes #102

### DIFF
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -368,7 +368,10 @@ def flatten_space(space: Space[Any]) -> Box | Dict | Sequence | Tuple | Graph:
     with flat boundaries and `edge_space` being a `Box` with flat boundaries or
     `None`. The box has exactly :func:`flatdim` dimensions. Flattening a sample
     of the original space has the same effect as taking a sample of the flattenend
-    space.
+    space. However, sampling from the flattened space is not necessarily reversible.
+    For example, sampling from a flattened Discrete space is the same as sampling from
+    a Box, and the results may not be integers or one-hot encodings. This may result in
+    errors or non-uniform sampling.
 
     Example::
         >>> from gymnasium.spaces import Box
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -367,7 +367,7 @@ def flatten_space(space: Space[Any]) -> Box | Dict | Sequence | Tuple | Graph:
     the result for graph spaces is always a `Graph` with `node_space` being a `Box`
     with flat boundaries and `edge_space` being a `Box` with flat boundaries or
     `None`. The box has exactly :func:`flatdim` dimensions. Flattening a sample
-    of the original space has the same effect as taking a sample of the flattenend
+    of the original space has the same effect as taking a sample of the flattened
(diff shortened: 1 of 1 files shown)


gymnasium/spaces/utils.py
```diff
@@ -368,7 +368,10 @@ def flatten_space(space: Space[Any]) -> Box | Dict | Sequence | Tuple | Graph:
     with flat boundaries and `edge_space` being a `Box` with flat boundaries or
     `None`. The box has exactly :func:`flatdim` dimensions. Flattening a sample
     of the original space has the same effect as taking a sample of the flattenend
```
pseudo-rnd-thoughts, line 370: `flattenend` to `flattened`

pseudo-rnd-thoughts (approved):
Looks good, could you fix the line above while we are modifying this file

PR #173, opened 2022-12-02
Updated Wrapper docs

Markus28:
- Some descriptions of what wrappers do was moved from doctoring of superclass to module docstring
- Examples of wrapper implementations were moved from docstrings to a tutorial
- Removed the titles for each individual wrapper in the docs

### DIFF
--- a/docs/api/wrappers.md
+++ b/docs/api/wrappers.md
@@ -12,6 +12,11 @@ wrappers/observation_wrappers
 wrappers/reward_wrappers
 ```
 
+```{eval-rst}
+.. automodule:: gymnasium.wrappers
+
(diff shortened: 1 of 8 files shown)


pseudo-rnd-thoughts (approved):
Are there any links between the tutorials and api docs

Markus28:
Not yet, but I can add one from the tutorial to the docs :)
Should we also do one from the module doctoring to the tutorial? That would need to be an absolute URL then, right?

pseudo-rnd-thoughts:
I believe that sphinx allows both relative and absolution links

Markus28:
Done :)

PR #174, opened 2022-12-03
Fixed formatting of spaces docs

Markus28:
- Some bullet point lists weren't rendered correctly because an empty line was missing
- Some places used single backticks instead of double
- Added some links to other parts of the docs via :class: :mod: etc...
- We should put documentation of flattening functions first before talking about flatdim


pseudo-rnd-thoughts (approved):
Thanks looks good

PR #178, opened 2022-12-04
checking success before splitting result in step_wait

### LINKED ISSUE #177: [Bug Report] Error handling in AsyncVectorEnv.step_wait
### Describe the bug

There is a minor bug in `AsyncVectorEnv.step_wait`.  Here's what happens:
1. If an exception is thrown by the step function of an env inside the worker...
2. ...the worker catches the exception and sends `(None, False)` back through the pipe to the `AsyncVectorEnv`.
3. `AsyncVectorEnv.step_wait` receives the message from the pipe and unpacks it as `result, success = pipe.recv()`.  At this point `result = None` and `success = False`.
4. AsyncVectorEnv.step_wait attempts to unpack `result` further using `obs, rew, terminated, truncated, info = result`
5. But, because `result` is `None`, this immediately throws a `TypeError` and bypasses the normal error handling using `_raise_if_errors`.
6. This means exceptions thrown inside the step method of vectorized envs do not get caught correctly, do not generate the proper error messages and are not handled by `_raise_if_errors`.

### Code example

```shell
from gymnasium import Env
from gymnasium.vector import AsyncVectorEnv, SyncVectorEnv

class TestEnv(Env):
    def __init__(self):
        super().__init__()
        self.action_space = Discrete(4)
        self.observation_space = Discrete(20)

    def reset(self):
        self.x = 0
        return self.x, {}

    def step(self, action):
        self.x += action
        terminal = False
        reward = 0
        if self.x >= 20:
            self.x = 19
            terminal = True
            reward = 1

        # this error will not be caught properly and handled by _raise_if_errors
        blerg

        return self.x, reward, terminal, False, {}
    
g = AsyncVectorEnv([TestEnv for _ in range(16)])
o, i = g.reset()
o, r, t, u, i = g.step([3 for _ in range(16)])
```

### System info

This is not system-dependent

### Additional context

_No response_

### Checklist

aaronwalsman:
# Description

There is a minor bug in `AsyncVectorEnv.step_wait`.  Here's what happens:
1. If an exception is thrown by the step function of an env inside the worker...
2. ...the worker catches the exception and sends `(None, False)` back through the pipe to the `AsyncVectorEnv`.
3. `AsyncVectorEnv.step_wait` receives the message from the pipe and unpacks it as `result, success = pipe.recv()`.  At this point `result = None` and `success = False`.
4. AsyncVectorEnv.step_wait attempts to unpack `result` further using `obs, rew, terminated, truncated, info = result`
5. But, because `result` is `None`, this immediately throws a `TypeError` and bypasses the normal error handling using `_raise_if_errors`.
6. This means exceptions thrown inside the step method of vectorized envs do not get caught correctly, do not generate the proper error messages and are not handled by `_raise_if_errors`.

My change first checks the value of `success` before attempting to unpack the contents of `result`.  The values are only unpacked and appended to the intermediate lists (`observations_list`, `rewards`, `terminateds`, `truncateds` and `infos`) if `success`.

Note that this may result in temporarily strange but benign behavior.  If there are 8 workers and two of them fail inside `step`, then the intermediate lists will only have 6 elements, and you would have to look at the list of successes to figure out which corresponded to which worker.  However, this is irrelevant because as soon as the loop exits, `_raise_if_errors` is called, which will handle exceptions, raise further exceptions and nothing will be done with those arrays anyway.  The alternative would be to generate dummy variables (`obs, rew, terminated, truncated, info = None, 0, False, False, {}`) to place in those lists temporarily in the case when `success == False` but this seems unnecessary.

It looks like you guys are currently revamping the VectorEnv stuff, so hopefully this is still relevant.  Feel free to ignore this if this chunk of code is already on the chopping block.

Fixes #177 

## Type of change

# Checklist:
I have not done these things, sorry!  Excuses:
I can't run pre-commit because of a node dependency that won't run on my system
I haven't added comments because the `step_wait` method had no comments to begin with, and this is a very minor change.
I haven't done anything to do the documentation because this is a small bugfix.
I don't think these changes generate any new warnings, but again, I can't run pre-commit on my system.
I haven't added any tests for this.  Again, small bugfix.
I can't run the unit-tests locally.

Anyway, hopefully this is useful!  Sorry for not being able to run the checks!

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -318,14 +318,15 @@ class AsyncVectorEnv(VectorEnv):
         successes = []
         for i, pipe in enumerate(self.parent_pipes):
             result, success = pipe.recv()
-            obs, rew, terminated, truncated, info = result
-
             successes.append(success)
-            observations_list.append(obs)
-            rewards.append(rew)
-            terminateds.append(terminated)
-            truncateds.append(truncated)
-            infos = self._add_info(infos, info, i)
+            if success:
+                obs, rew, terminated, truncated, info = result
+
+                observations_list.append(obs)
+                rewards.append(rew)
+                terminateds.append(terminated)
+                truncateds.append(truncated)
+                infos = self._add_info(infos, info, i)
 
         self._raise_if_errors(successes)
         self._state = AsyncState.DEFAULT


pseudo-rnd-thoughts (approved):
Thanks for the PR and spotting the issue.
Would you be interested in contributor more to Gymnasium?

aaronwalsman:
Hey, thanks for quickly pulling this in!  I may be able to contribute a few small things here and there, but my time is a bit limited these days as I'm trying to wrap up my PhD :)

PR #197, opened 2022-12-08
Fix exception causes in video_recorder.py

cool-RR:
More context for this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with


pseudo-rnd-thoughts:
Thanks for the PR, the blog post is very cool.
Has this issue been raised with flake8 as we use it as our code analyse tool?

Would you be interested in finding all the other places in the codebase where this is used and applying the style there, otherwise, I can ask others

pseudo-rnd-thoughts (approved):

Issue #181, opened 2022-12-04
[Proposal] Add wrappers used to EnvSpec


WillDudley:
### Proposal

Add wrapper info to an environment's spec to achieve full reproducibility in a readable manner.

### Motivation

I'm no expert on the registry, but from my understanding [EnvSpec](https://github.com/Farama-Foundation/Gymnasium/blob/7e885597633bd7d9730ddf7c9524fda5d960ecb9/gymnasium/envs/registration.py#L122) is meant to aid in maximal reproducibility of Gymnasium environments. I trust that it does this at the base level, but I don't think it includes wrappers.

Maximal, complete reproducibility of RL environments is of vital importance for reproducibility and versioning of both models and datasets. We need a system that provides a bijection (two-way conversion) between structured, readable data and a Gymnasium environment.

### Pitch

We need a system that provides a bijection (two-way conversion) between structured, readable data and a Gymnasium environment. This involves:

- Extending env.spec to include wrappers,
- creation of a function to generate a gymnasium environment from env.spec,
- writing of tests to assert this complete reproducibility.

Something of this ilk:

```py
env = gym.make("halfcheetah-v4", <kwargs>)
env = wrapper_1(env, <kwargs>)
env = wrapper_2(env, <kwargs>)

complete_env_spec = env.spec

recreated_env = gym.make_from_spec(complete_env_spec)

assert env == recreated_env
```

### Alternatives

EzPickle was/is an answer to this which some of us have contributed to. Here are some disadvantages of it:

- It's unreadable data,
- it relies on pickle, making it unsafe to share,
- there's little documentation surrounding it.

### Additional context

Note that the order of wrappers applied may matter, so this information should be kept.

This feature is necessary for Minari to fully shine, as datasets need to be fully reproducible.

This feature should also be added to PettingZoo 2.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

WillDudley:
This could be achieved by having wrappers inherit from a superclass which modifies/extends env.spec with important wrapper attributes.

WillDudley:
I'm not committed to this proposal but I do think there needs to be a standardised env stack spec. We're wanting to generate datasets for gymnasium robotics soon, and it'd be good to include the env stack in the metadata for reproducibility puroposes.

[...]

PR #199, opened 2022-12-08
Spec stack

WillDudley:
# Description

Solution to https://github.com/Farama-Foundation/Gymnasium/issues/181

```
     env.spec_stack   <->      json          <->            str
           ^                    ^                            ^
       wrapped env       readable encoding         saveable in MinariDataset
```

### log specstack with MinariDataset
```py
import json
import gymnasium as gym

from gymnasium.utils.serialize_spec_stack import serialise_spec_stack

env_name = "FetchReach-v3"
env = gym.make(env_name)
environment_stack = json.dumps(serialise_spec_stack(env.spec_stack))

dataset_name = "FetchReach_v3_my-dataset-description"
dataset = MinariDataset(
    dataset_name=dataset_name
    ....
    environment_stack=environment_stack,
    ...
)
dataset.save()
minari.upload_dataset(dataset_name)
```

### recreate env from MinariDataset
```py
import json
import minari
import gymnasium as gym

dataset = minari.download_dataset("FetchReach_v3_my-dataset-description")
env = gym.make(json.loads(dataset.environment_stack))
```

## todo

- [x] encode/decode lambda functions
- [x] complete docstrings
- [x] ensure gym.make() works / update gym.make()
- [x] update tutorial


gymnasium/envs/registration.py
```diff
@@ -526,6 +526,7 @@ def make(
     autoreset: bool = False,
     apply_api_compatibility: Optional[bool] = None,
     disable_env_checker: Optional[bool] = None,
+    allow_default_wrappers: Optional[bool] = True,
```
pseudo-rnd-thoughts, line 529: I personally dislike this, `make` already has enough parameters
If a `spec_stack` is passed to `make`, we should bypass the standard wrappers applied and only use the `WrapperSpec`, imo

gymnasium/core.py
```diff
@@ -268,8 +268,8 @@ def __getattr__(self, name: str):
             raise AttributeError(
                 "Can't access `_np_random` of a wrapper, use `self.unwrapped._np_random` or `self.np_random`."
             )
-        elif name.startswith("_"):
-            raise AttributeError(f"accessing private attribute '{name}' is prohibited")
+        # elif name.startswith("_"):
```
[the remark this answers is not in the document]
WillDudley: temporary debugging purposes, I thought it may need to be permanently removed but it's fine now - added it back

gymnasium/envs/registration.py
WillDudley: "If a spec_stack is passed to make, we should bypass the standard wrappers applied and only use the WrapperSpec, imo"

Yes, this is my motivation for `allow_default_parameters`, otherwise `gym.make(envspec)` won't bypass the standard wrappers. What alternative solution do you propose? I'm struggling to think of any.

gymnasium/envs/registration.py
```diff
@@ -656,37 +659,38 @@ def make(
     spec_.kwargs = _kwargs
     env.unwrapped.spec = spec_
 
-    # Add step API wrapper
-    if apply_api_compatibility is True or (
-        apply_api_compatibility is None and spec_.apply_api_compatibility is True
-    ):
-        env = EnvCompatibility(env, render_mode)
+    if allow_default_wrappers:
```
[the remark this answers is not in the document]
WillDudley: See other comment (re: registration.py)

gymnasium/envs/registration.py
pseudo-rnd-thoughts: Should not a simple type check work? 
If string, look up Env spec
If Env spec, use Env spec
If sequence, extract wrappers specs and Env spec

WillDudley: Env spec includes kwargs like max_steps etc which automatically construct the standard wrappers. Note that the entire spec_stack isn't passed to make(), only the EnvSpec element in the spec_stack.

pseudo-rnd-thoughts: > Env spec includes kwargs like max_steps etc which automatically construct the standard wrappers. Note that the entire spec_stack isn't passed to make(), only the EnvSpec element in the spec_stack.

I don't understand? I thought the purpose of the spec stack was to have a list of wrapper / env specifications that can be saved and from the full environment stack could be recreated. Therefore, should we not provide capacity with `make` to pass the whole spec stack

WillDudley: I'm with you. Originally I had a separate `recreate_env()` function but I've now updated `gym.make()` to work with `SpecStack`s

WillDudley:
@pseudo-rnd-thoughts 

Gonna take a break from this for today and maybe tomorrow, but I'm willing to discuss anytime. 

Personally I think there should be an easier way to stringify callables, and that ezpickle_args should be depreciated.

WillDudley:
This is ready for review. There are two cases of recursion that I can't think how to recursify (see the TODOs in registration.py), but if it isn't easy then it shouldn't be much of an issue to assume limited depth.

I think that, like with EnvSpec, WrapperSpec should not support `args` and all wrapper arguments should forcibly be `kwargs` for interpretability reasons.

Please may someone more familiar with `registration.py` resolve the conflicts.

WillDudley:
pre-commits I need assistance with:
image
image

WillDudley:
This has now been implemented in MinariDataset (https://github.com/Farama-Foundation/Minari/pull/14).

The whole process now looks like this:
```
          env            <->             SpecStack              <->          json              <->                    str
            ^                                                                                 ^                                            ^
   wrapped env                                                          readable encoding            saveable in MinariDataset
```

gymnasium/envs/registration.py
```diff
@@ -521,7 +715,7 @@ def register(
 
 
 def make(
```
pseudo-rnd-thoughts, line 717: Personally and I think Ariel agrees, `make` is already too complex and this adds more weird complex code. 
Could we split this into other functions to simplify the overall function.

docs/tutorials/environment_creation.py
```diff
@@ -508,3 +508,26 @@ def close(self):
...
+# To remake the environment, you can use ``gymnasium.make(SpecStack(json.load(open("my_environment_config.json", "r"))))``.
+#
+# .. code:: python
+#
+#    import gym_examples
+#    from gym_examples.wrappers import RelativePosition
+#
+#    env = gymnasium.make('gym_examples/GridWorld-v0')
+#    wrapped_env = RelativePosition(env)
+#    spec_stack = gymnasium.SpecStack(wrapped_env)
```
pseudo-rnd-thoughts, line 527: I dislike this, personally, I think this should be `env.spec_stack` as a comparable function to `env.spec` but this probably need more discussion

pseudo-rnd-thoughts (changes requested):
I think we need a larger discussion on the implementation.
The registration function is very large and complex, personally, I would like to split it up into several classes, one for `EnvSpec` etc but I wasn't able to find an easy way of doing that

gymnasium/envs/registration.py
WillDudley: PR is incomplete but I gave this a go (check recent changes)

gymnasium/core.py
```diff
@@ -52,6 +52,8 @@ class Env(Generic[ObsType, ActType]):
     render_mode: str | None = None
     reward_range = (-float("inf"), float("inf"))
     spec: EnvSpec | None = None
+    spec_stack: SpecStack | None = None
```
pseudo-rnd-thoughts, line 55: I think this should be a property rather than an attribute. 
If you use the function proposed in the issue then you don't need to include any of the `self.spec_stack.append_wrapper(self)`

WillDudley: In order to avoid having to explicitly define whether a wrapper is an outer layer or not, there would need to be a callback in every wrapper like how `gym.utils.EzPickle.__init__(self)` is used. This is why we need `append_wrapper()`. `spec_stack()` (the function proposed in the issue), on the other hand, does not call wrappers at all; it recursively collects metadata from the unwrapping of wrappers.

pseudo-rnd-thoughts: What do you mean by an outer layer? In my head, I would have thought, you receive a sequence of wrapper specs + a env spec and you just generate each of the wrappers in order. Therefore, the `env.spec_stack` and `EzPickle` as part of the function definition is all that is needed.
Could we move the json based code for loading and saving info to minari not gymnasium? 

Personally, I think all gymnasium should do in this PR is add `EzPickle` to wrappers, add a `spec_stack` function and add functional to make for a sequence of wrapper and env spec. I would like minari to be as minimum as possible in gymnasium.

pseudo-rnd-thoughts: I think you have misread me as I wish `spec_stack` to be a function (technically a property) not an attribute. 
For `spec` this is an attribute in `Env` but a function (technically a property) in `Wrapper`.

This is my proposal from the issue, modified for `Env` and `Wrapper`
```python
class Env:
    @property
    def spec_stack(self) -> tuple[Union[EnvSpec, WrapperSpec]]:
         assert self.spec is not None
         return (self.spec,)
     
class Wrapper:
    @property
    def spec_stack(self) -> tuple[Union[EnvSpec, WrapperSpec]]:
          assert hasattr(self, "_ezpickle_kwargs") and hasattr(self, "_ezpickle_args")
          wrapper_spec = WrapperSpec(self.__name__, self.__module__ + "." + self.__name__, self._ezpickle_args, self.ezpickle_kwargs)
          return (wrapper_spec,) + self.env.spec_stack
```

Therefore, this is no need for the `append_wrapper`, just that the wrapper class inherit from `EzPickle` and the environment has a `spec`. It might be possible to wrap the into the `SpecStack` class but this would be for simplicity rather than 2/3 separate functions.

On where the code goes minari or gymnasium. It makes more sense now why we would keep it in gymnasium. I would just have the code in `utils` rather than `registration.py`.

WillDudley: Wow, I overcomplicated this.

Ok, I get you now. I've modified this PR accordingly (still need to update the tutorial, so ignore that). Please lmk your thoughts.

gymnasium/dataclasses.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass
```
WillDudley, line 1: This file was needed (I think) because putting it in `registration.py` or `core.py` both cause a circular import error for some reason - will need to think on this.

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,123 @@
...
+        if callable(v):
+            try:
+                str_repr = (
+                    str(inspect.getsourcelines(v)[0]).strip("['\\n']").split(" = ")[1]
+                )  # https://stackoverflow.com/a/30984012
+            except OSError:
+                raise error.Error(
+                    "The attempted seialisation of a callable failed. This is likely due to env.spec_stack being called twice, which for technical reasons doesn't work. Please modify your code to only call env.spec_stack once (you can save that to a variable and repeatedly use that)."
+                )
+            str_repr = re.search(r", (.*)\)$", str_repr).group(1)
```
WillDudley, line 54: Raises the precommit error " error: "group" is not a known member of "None" (reportOptionalMemberAccess"

but idk why

WillDudley:
Would this PR be fine without adding tests as it's a util, I can't think of any obvious failure cases

gymnasium/dataclasses.py
WillDudley: Still unsure on any better alternative

gymnasium/utils/serialize_spec_stack.py
pseudo-rnd-thoughts: I suspect that it is because, `re.search` has an optional return type. Therefore you just need to split up the statement if to two, if a check in the middle that the result of `search` is not None

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,123 @@
...
+            spec.kwargs[k] = str_repr
+    return spec
+
+
+def deserialise_spec_stack(
+    stack_json: str, eval_ok: bool = False
```
pseudo-rnd-thoughts, line 60: stack_json type hint is a string but the first line is `stack_json.items()` meaning that it is a dict

pseudo-rnd-thoughts (changes requested):
I think that the general idea outlined in `environment_creation`.
Before merging, I need to go through the `make` implementation in more detail

gymnasium/utils/ezpickle.py
```diff
@@ -19,17 +19,15 @@ class EzPickle:
 
     def __init__(self, *args, **kwargs):
         """Uses the ``args`` and ``kwargs`` from the object's constructor for pickling."""
-        self._ezpickle_args = args
```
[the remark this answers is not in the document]
WillDudley: I removed them with reasons outlined in https://github.com/WillDudley/Gymnasium/pull/2

But perhaps I'm missing something regarding their importance/usage.

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,123 @@
...
+    """Serialises the specification stack into a JSON string.
+
+    Args:
+        stack: Tuple of environment and wrapper specifications, known as the specification stack. Generated via `env.spec_stack`.
+
+    Returns:
+        A JSON string representing the specification stack.
+    """
+    num_layers = len(stack)
+    stack_json = {}
```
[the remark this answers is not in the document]
WillDudley: To me a dict seems like a more "json-like" structure than a list. The idea of stack_json is that it can be saved, interpreted and/or shared via `json.dumps()` or otherwise. Otherwise we have a list of dicts.

The added interpretability of the key-value-pair nature of stack_json is also a bonus.

gymnasium/utils/ezpickle.py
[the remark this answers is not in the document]
WillDudley: Following discussions on Discord, I've added back ezpickle_args for legacy purposes - they will not work with spec_stack functionality but will raise a warning  when detected and the "old way" of using ezpickle will still work (nothing in this PR would break it).

gymnasium/dataclasses.py
WillDudley: sticking with this (renamed file though)

gymnasium/utils/serialize_spec_stack.py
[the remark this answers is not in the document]
RedTachyon: Bad idea imo. I understand the sentiment behind "json-likeness", but what counts is that it's not json-like -- it actually has to be valid JSON.

JSON objects don't generally preserve order. In this specific case we're kinda lucky because objects map to python dictionaries which do preserve order, but we're being very not-future-proof, I can come up with several scenarios where the JSON would be stored somewhere via some other language, maybe even modified, and then sent back to python (with the order mangled)


What you probably want to do is something like
```json
{
"spec":
[
    "someSpec",
    "someOtherSpec",
]
}
```

because afaik JSON does need to be an object on the top-level.

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,139 @@
...
+        spec: An environment or wrapper specification.
+
+    Returns: The specification with lambda functions converted to strings.
+
+    """
+    for k, v in spec.kwargs.items():
+        if callable(v):
+            try:
+                str_repr = (
+                    str(inspect.getsourcelines(v)[0]).strip("['\\n']").split(" = ")[1]
```
RedTachyon, line 58: uhhhhhhhhhhhhhhhhhhh

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,139 @@
...
+                )  # https://stackoverflow.com/a/30984012
+            except OSError:
+                raise error.Error(
+                    "The attempted seialisation of a callable failed. This is likely due to env.spec_stack being called twice, which for technical reasons doesn't work. Please modify your code to only call env.spec_stack once (you can save that to a variable and repeatedly use that)."
```
RedTachyon, line 62: u(h)*100

Is there like a deep technical reason why this happens? I'm guessing it's related to the mildly cursed parsing of the lambda function code?

RedTachyon: This also fails in an interactive context which is a huge use-case of python

RedTachyon: This also also fails with an input like
```python
env = gym.experimental.wrappers.LambdaActionV0(env, lambda x: 1-x, None)
```
where the lambda function is parsed as 
```python
lambda x: 1-x, None
```

I didn't actually check deserialization, but I can't see this working properly

RedTachyon: Also it completely fails (with an incomprehensible error) when the lambda function is defined with the usual
```python
def switch(action):
    return 1 - action
```

AFAIK (or I definitely hope so) the lambda wrappers aren't limited to the actual python lambda statements, which are... bad. 
At least my understanding and intention is that the "lambda" refers to the general idea of pure functions (going back to **lambda** calculus), and overfitting to the specific way of defining pure functions calls for huge trouble.

RedTachyon (changes requested):
I left a few comments, but there's a larger discussion to be had about the high-level design - see discord.

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,139 @@
...
+                raise error.Error(
+                    f"Expected spec to be a WrapperSpec, but got {type(spec)} instead."
+                )
+            layer = f"wrapper_{num_layers - i - 2}"
+        spec_json = json.dumps(dataclasses.asdict(spec))
+        stack_json[layer] = spec_json
+    return stack_json
+
+
+def _serialise_callable(spec: Union[WrapperSpec, EnvSpec]) -> dict:
```
RedTachyon, line 45: (forgot to finish the comment in the review)

As I understand, the goal here is to serialize a (more or less arbitrary) python function into a string? 

This is, in general, borderline impossible in a sane way. There are some approaches and some libraries which do that, but there are generally some limitations. Importantly, I don't think we can hope to find a way to do it robustly, so I'd strongly suggest reconsidering the design that requires this functionality. (unless we specify the exact limitations and support everything within those bounds) If we do try to do it, we almost definitely don't want to try and make our own implementation 

To eliminate a few tempting options:
- The current implementation doesn't work because it sometimes parses incorrectly, and it only could work with `lambda x: y` functions, which is a very bad limitation (it also has other important limitations)
- Getting the full function code doesn't work because it might use externally defined functions
- Saving and executing the full file doesn't work because it might have side effects, also there could be external imports
- Saving and executing the full file and all files that it imports doesn't work because that might include huge external libraries (e.g. numpy)
- Saving and executing the full file and all files that it imports, checking if the other module is defined by the user, and otherwise just copying the import so that the appropriate library is imported after deserialization is actually Einsten's definition of insanity (the whole repetition thing is a secondary symptom).

Source: I tried to do the same thing some years ago. Don't recommend.

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -0,0 +1,139 @@
...
+import json
+import re
+from typing import Union
+
+from gymnasium import error
+from gymnasium.envs.registration import EnvSpec
+from gymnasium.wrapperspec import WrapperSpec
+
+
+def serialise_spec_stack(stack: "tuple[Union[WrapperSpec, EnvSpec]]") -> str:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Add `from __future__ import annotations`
And the type hint to `tuple[WrapperSpec | EnvSpec, ...]`

gymnasium/utils/serialize_spec_stack.py
```diff
@@ -53,22 +54,8 @@ def _serialise_callable(spec: Union[WrapperSpec, EnvSpec]) -> dict:
...
-                )
-            found_callable = re.search(r", (.*)\)$", str_repr)
-            if found_callable is not None:
-                str_repr = found_callable.group(1)
-            else:
-                raise error.Error(
-                    "The attempted seialisation of a callable failed. This is likely due to env.spec_stack being called twice, which for technical reasons doesn't work. Please modify your code to only call env.spec_stack once (you can save that to a variable and repeatedly use that)."
-                )
-            spec.kwargs[k] = str_repr
+            warn("Callable found in environment specification stack. This likely comes from a wrapper that inputs a function. "
```
pseudo-rnd-thoughts, line 57: I would add wrapper / env information to make it easier for users to debug

pseudo-rnd-thoughts:
I just realised a possible issue, as most of the experimental wrappers now inherit from `LambdaAction` or `LambdaObservation` and pass the function being called, i.e., `ClipAction = LambdaAction(func=lambda: np.clip())`.

As both `ClipAction` and `LambdaAction` need to inherit from `EzPickle` then `LambdaAction` will try to `EzPickle` the function which we can't do + `ClipAction` doesn't need this info to reconstruct + can mess up the reconstruction process when `LambdaAction` contains args not in `ClipAction`

I think the solution is two-fold, we require that the `EzPickle` super `__init__` call is before the `Wrapper` super `__init__` call. Then have a check in `EzPickle` to ignore adding **all** args and kwargs if `__ezpickle_args` already exists in the object. 

There might be an issue with multi-inheritance but I haven't been able to work out how

pseudo-rnd-thoughts:
Closing in favor of #292 which I believe has fixed all of the requested changes

Issue #202, opened 2022-12-09
[Proposal] `play` callback for Minari


Markus28:
### Proposal

Add a class that implements a callback function to record transitions that are generated in `play` to a Minari dataset.

### Motivation

This might be convenient for anyone who wants to train agents from human demonstrations

### Alternatives

Implement it in the Minari repository, or somewhere else.

### Additional context

We might want to add a cleanup callback to the play function that is executed once the window is closed. This cleanup callback would be used to actually store the data to disk. 

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

Kallinteris-Andreas:
@younik do you want such a function in minari?
If so, transfer the issue there, if not feel free to close this issue

pseudo-rnd-thoughts:
#203 specified that this change would happen in Minari

PR #203, opened 2022-12-09
Implementation of Minari callback for `play` 

Markus28:
Implements #202

Still need to check:
- Do reward, terminated, truncated etc. have to have a singleton dimension as in tutorials
- How to handle non-numpy observations, actions
- Are non-flat observations/actions handled correctly by Minari?
- Can we save intermediate results to disk if the replay buffer grows too large?

### DIFF
--- a/gymnasium/utils/play.py
+++ b/gymnasium/utils/play.py
@@ -154,6 +154,7 @@ def play(
     keys_to_action: Optional[Dict[Union[Tuple[Union[str, int]], str], ActType]] = None,
     seed: Optional[int] = None,
     noop: ActType = 0,
+    cleanup_callback: Optional[Callable] = None,
 ):
     """Allows one to play the game using keyboard.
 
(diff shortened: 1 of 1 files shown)


WillDudley:
"Can we save intermediate results to disk if the replay buffer grows too large?"

This is a feature I'd like to add to Minari - appending to MinariDatasets in batches. I don't believe this was supported in d3rlpy (which Minari is based off), so it would need to be added.

One potential issue with a replay buffer not fitting into memory is now it's more difficult to sample from earlier transitions (in case you'd want to do some wide-range off-policy learning)

pseudo-rnd-thoughts:
Would it be possible that if we wrap the environment with the minari dataset collector then call `play` on it, this should record all of the actions, observations, etc from the environment?

rodrigodelazcano:
@pseudo-rnd-thoughts that should be the case. We'll have to specify an event to call `minari.create_dataset_from_collector_env()` so that the data can be stored in a permanent location. This can be done with a custom `callback` in `play` and also adding the environment as one of the arguments.

pseudo-rnd-thoughts:
@rodrigodelazcano Could we not make that call before or after calling `play`. My point is more that I don't think it is necessary to modify `play` to saving the data.

pseudo-rnd-thoughts:
We are going to make this change in minari

PR #208, opened 2022-12-10
Fix `MuJoCo.Ant` action order in DOC

Kallinteris-Andreas:
# Description

source: https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/envs/mujoco/assets/ant.xml#L71


pseudo-rnd-thoughts (approved):
Thanks for reading the xml

PR #216, opened 2022-12-13
Revert `make` error when render mode is used without metadata render modes

pseudo-rnd-thoughts:
If an environment doesn't have static metadata containing the render modes but a user tries to create an environment then an error will be raised. This error is replaced by a warning.
A test has been added to check this


jjshoots (commented):
LGTM

jjshoots (approved):

PR #223, opened 2022-12-20
Update vector space utility functions for all spaces

pseudo-rnd-thoughts:
I am improving some of the wrappers and I found that we can use several of the built-in vector functions for generating empty arrays, stacking observations, etc

In this PR, I have copied the original space utility functions from vector which were in two files, `spaces` and `numpy_utils` into a single file, `space_utils`, the functions are `batch_space`, `concatenate`, `iterate` and `create_empty_array`

I have not changed the implementation of the functions other than adding new functions for the `Sequence`, `Text`, `Graph` spaces

For testing, I have not included expected outputs for each of the functions. This can be added, rather I have opted for functionality testing of the functions, in particular the interface between the four functions


RedTachyon:
I'm slightly confused by this, as I understand this is just refactoring of existing code, except it's put into the experimental directory, so it kinda looks like new code in the diff? Wouldn't it be better to just change it in the live version of the repo, and then it will get merged with the experimental stuff when that becomes stable?

pseudo-rnd-thoughts (commented):
With this implementation, the functions are in a single file where before the functions were in two functions (numpy_utils.py and spaces.py)

Im happy to add the changes to root vector

pseudo-rnd-thoughts (commented):
To help reviewing and such that the standard vector utils functions are updated, I have applied the changes to `gymnasium.vector.utils.X` functions

PR #225, opened 2022-12-21
Improve the warning when error is raised from plugin

pseudo-rnd-thoughts:
Within the plugin system, it is possible that the plugin raises an error, this is logged as a warning however does not give context on the plugin that caused the error.


RedTachyon:
Do you have a before and after of how the resulting warning looks?

RedTachyon (approved):

PR #227, opened 2022-12-24
Jax Functional BlackJack Implementation

balisujohn:
# Description

This Pull request adds a Jax-Based functional BlackJack Environment to Gymnasium. 

### Some Design Notes

So this adds a new environments folder called `jax_toy_text` which is intended to house all the new jax toytext environments. The existing jax cartpole inherits from `FunctionalJaxEnv` which is in the `experimental` folder, but `BlackJackJaxEnv` inherits from `JaxEnv` instead which is custom defined inside `conversion.py` inside the `jax_toy_text` folder. Before merging probably it is ideal to generalize `FunctionalJaxEnv` so it can support Jax BlackJack Env.

Right now the assets used for rendering in `BlackJackJaxEnv` are in the toy_text asset directories, to avoid duplication of assets. 

A human rendering test is included in the main function of `blackjack.py` in `jax_toy_text`. This is intended for testing out the environment/rendering, and will be removed before this is merged. 

The tests are in their owm new file called `test_jax_blackjack.py` due to being somewhat different than the cartpole and pendulum tests. These will likely need to be merged, and I think more tests for the actually blackjack functions would be good. 

image

Human rendering works, but it does seem like a window appears and instantly is closed before the first persistent window opens on calling the first render, so maybe there is a subtle bug there.

This is a draft PR, and while the environment works for testing purposes and the tests pass, I think more thought is needed before merging this.

## Type of change

# Checklist:


balisujohn:
So I generalized  the `FunctionalJaxEnv` class in `func_jax_env.py` to support wrapping `BlackjackFunctional`.

I think the remaining checklist is as follows:

Fix pydocstyle 
Add test to make sure blackjack transitions are correct
Think about how to combine tests
Figure out why a pygame window appears for a second in the human rendering `__main__`  function in blackjack.py, and remove it once rendering seems to be working seamlessly.

PR #230, opened 2022-12-31
ResizeObservation: fix silent bug on 2-dimensional observations.

ianyfan:
# Description

The `ResizeObservation` wrapper silently becomes buggy when passed an environment that produces two-dimensional observations, such as the output of `GrayscaleObservation` with `keep_dim=False` (the default), since it sets the `observation_space` to be two-dimensional (WxH), but actually produces three-dimensional output (WxHx1).

I changed the `observation` method to keep the shape, whether it be two- or three-dimensional. I also added extra checks in the constructor to make sure that the input is either two- or three-dimensional, since `cv2.resize` does not work otherwise.

Here is some code that demonstrates the bug:
```python
import gymnasium as gym
from gymnasium.wrappers import GrayScaleObservation, ResizeObservation


def assert_space_is_consistent(env):
    obs_shape = env.reset()[0].shape
    expected_shape = env.observation_space.shape
    if obs_shape != expected_shape:
        raise ValueError(f'Expected {expected_shape}, got {obs_shape}')


env = gym.make('CarRacing-v2', render_mode='rgb_array')

# Sanity check: works when keeping the dimension, i.e. WxHx1
gray_env = GrayScaleObservation(env, keep_dim=True)
resized_env = ResizeObservation(gray_env, 100)
assert_space_is_consistent(resized_env)  # this always passes

# Fails when discarding the dimension (the default behaviour), i.e. WxH
gray_env_flattened = GrayScaleObservation(env, keep_dim=False)
resized_env_flattened = ResizeObservation(gray_env_flattened, 100)
assert_space_is_consistent(resized_env_flattened)
```

In the current main branch, the last assert fails with the error:
```
ValueError: Expected (100, 100), got (100, 100, 1)
```
but passes with this fix applied.

Let me know if this change is desirable, and then I will go through the checklist before asking for a review. :)

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/resize_observation.py
+++ b/gymnasium/wrappers/resize_observation.py
@@ -25,23 +25,29 @@ class ResizeObservation(gym.ObservationWrapper):
         (64, 64, 3)
     """
 
-    def __init__(self, env: gym.Env, shape: Union[tuple, int]):
+    def __init__(self, env: gym.Env, shape: Union[tuple[int, int], int]) -> None:
         """Resizes image observations to shape given by :attr:`shape`.
 
         Args:
             env: The environment to apply the wrapper
             shape: The shape of the resized observations
         """
-        super().__init__(env)
-        if isinstance(shape, int):
-            shape = (shape, shape)
-        assert all(x > 0 for x in shape), shape
+        if not isinstance(env.observation_space, Box):
+            raise ValueError(f"Expected the observation space to be Box, actual type: {type(env.observation_space)}")
+        dims = len(env.observation_space.shape)
+        if not 2 <= dims <= 3:
+            raise ValueError(f"Expected the observation space to have 2 or 3 dimensions, got: {dims}")
 
+        try:
+            shape = tuple(shape)
+        except TypeError:
+            shape = (shape, shape)
+        if len(shape) != 2 or not all(isinstance(x, int) and x > 0 for x in shape):
+            raise ValueError(f"Expected shape to be a 2-tuple of positive integers, got: {shape}")
+
+        super().__init__(env)
         self.shape = tuple(shape)
 
-        assert isinstance(
-            env.observation_space, Box
-        ), f"Expected the observation space to be Box, actual type: {type(env.observation_space)}"
         obs_shape = self.shape + env.observation_space.shape[2:]
         self.observation_space = Box(low=0, high=255, shape=obs_shape, dtype=np.uint8)
 
@@ -67,6 +73,4 @@ class ResizeObservation(gym.ObservationWrapper):
         observation = cv2.resize(
             observation, self.shape[::-1], interpolation=cv2.INTER_AREA
         )
-        if observation.ndim == 2:
-            observation = np.expand_dims(observation, -1)
-        return observation
+        return observation.reshape(self.observation_space.shape)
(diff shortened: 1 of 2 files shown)


gymnasium/wrappers/resize_observation.py
```diff
@@ -25,23 +25,29 @@ class ResizeObservation(gym.ObservationWrapper):
...
         Args:
             env: The environment to apply the wrapper
             shape: The shape of the resized observations
         """
-        super().__init__(env)
-        if isinstance(shape, int):
+        if not isinstance(env.observation_space, Box):
+            raise ValueError(f"Expected the observation space to be Box, actual type: {type(env.observation_space)}")
+        dims = len(env.observation_space.shape)
+        if not 2 <= dims <= 3:
```
RedTachyon, line 38: Rewrite to `assert dims in (2, 3), "Expected the observation space [...]"`

gymnasium/wrappers/resize_observation.py
```diff
@@ -25,23 +25,29 @@ class ResizeObservation(gym.ObservationWrapper):
...
+
+        try:
+            shape = tuple(shape)
+        except TypeError:
             shape = (shape, shape)
-        assert all(x > 0 for x in shape), shape
+        if len(shape) != 2 or not all(isinstance(x, int) and x > 0 for x in shape):
+            raise ValueError(f"Expected shape to be a 2-tuple of positive integers, got: {shape}")
 
+        super().__init__(env)
```
RedTachyon, line 48: This should stay at the beginning of the constructor

gymnasium/wrappers/resize_observation.py
```diff
@@ -25,23 +25,29 @@ class ResizeObservation(gym.ObservationWrapper):
...
+    def __init__(self, env: gym.Env, shape: Union[tuple[int, int], int]) -> None:
         """Resizes image observations to shape given by :attr:`shape`.
 
         Args:
             env: The environment to apply the wrapper
             shape: The shape of the resized observations
         """
-        super().__init__(env)
-        if isinstance(shape, int):
+        if not isinstance(env.observation_space, Box):
```
RedTachyon, line 35: Please keep these checks as `assert` statements

gymnasium/wrappers/resize_observation.py
```diff
@@ -25,23 +25,29 @@ class ResizeObservation(gym.ObservationWrapper):
...
+            raise ValueError(f"Expected the observation space to be Box, actual type: {type(env.observation_space)}")
+        dims = len(env.observation_space.shape)
+        if not 2 <= dims <= 3:
+            raise ValueError(f"Expected the observation space to have 2 or 3 dimensions, got: {dims}")
+
+        try:
```
RedTachyon, line 41: Why this change?

RedTachyon (changes requested):
I did a basic code quality pass, but please fix pre-commit first

pseudo-rnd-thoughts (commented):
Could you add tests for this new functionality

ianyfan:
Review comments have been addressed and tests/docs added

gymnasium/wrappers/resize_observation.py
ianyfan: No real reason apart from personal stylistic preference, so I have changed it back to reduce noise

gymnasium/wrappers/resize_observation.py
ianyfan: I thought it would be better to do input checking before any heavy initialization (i.e. fail early), but I have changed it anyway

pseudo-rnd-thoughts (approved):

PR #232, opened 2023-01-02
New vector API updates - jax vector env, moving the vector classes

RedTachyon:
This contains the final (?) necessary feature for the vector API, and a few general code improvements.

The main thing is `functional_jax_env.FunctionalJaxVectorEnv`, which should properly reset the necessary vector envs.

I moved all vector env APIs into `experimental/vector/...` to put it in line with #196

This should probably be merged into the branch after light review, and the thorough review will happen before we merge the whole thing to main.


gymnasium/experimental/functional_jax_env.py
```diff
@@ -0,0 +1,258 @@
...
+        self.spec = spec
+
+        self._is_box_action_space = isinstance(self.action_space, gym.spaces.Box)
+
+        if self.render_mode == "rgb_array":
+            self.render_state = self.func_env.render_init()
+        else:
+            self.render_state = None
+
+        np_random, _ = seeding.np_random()
```
pseudo-rnd-thoughts, line 57: The second element of `seeding.np_random` returns the original seed. This is normally ignored for obvious reasons but for `None` this returns the seed used to reproduce `np_random`. 
Therefore, I would remove the next line in favor of the second element

gymnasium/experimental/functional_jax_env.py
```diff
@@ -0,0 +1,258 @@
...
+    """A vector env implementation for functional Jax envs."""
+
+    state: StateType
+    rng: jrng.PRNGKey
+
+    def __init__(
+        self,
+        func_env: FuncEnv,
+        num_envs: int,
+        observation_space: Space,
```
pseudo-rnd-thoughts, line 128: Can we infer this information from the func env and num envs?

gymnasium/experimental/functional_jax_env.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self._is_box_action_space = isinstance(self.action_space, gym.spaces.Box)
+
+        if self.render_mode == "rgb_array":
+            self.render_state = self.func_env.render_init()
+        else:
+            self.render_state = None
+
+        np_random, _ = seeding.np_random()
+        seed = np_random.integers(0, 2**32 - 1, dtype="uint32")
```
pseudo-rnd-thoughts, line 160: Can comment as above

gymnasium/experimental/functional_jax_env.py
```diff
@@ -0,0 +1,258 @@
...
+            self.steps = self.steps.at[to_reset].set(0)
+
+        observation = self.func_env.observation(next_state)
+        observation = _convert_jax_to_numpy(observation)
+
+        return observation, reward, terminated, truncated, info
+
+    def render(self):
+        """Returns the render state if `render_mode` is "rgb_array"."""
+        if self.render_mode == "rgb_array":
```
pseudo-rnd-thoughts, line 229: Add "human" support

gymnasium/experimental/functional_jax_env.py
```diff
@@ -0,0 +1,258 @@
...
+            raise NotImplementedError
+
+    def close(self):
+        """Closes the environments and render state if set."""
+        if self.render_state is not None:
+            self.func_env.render_close(self.render_state)
+            self.render_state = None
+
+
+def _convert_jax_to_numpy(element: Any):
```
pseudo-rnd-thoughts, line 244: Can we not use the `experimental.wrappers.jax_to_numpy` functions?

pseudo-rnd-thoughts (changes requested):
Personally, I would like to rewrite all of the vector implementations (its not that much code) rather than just wholesale copy and paste the files over

gymnasium/experimental/functional_jax_env.py
RedTachyon: I tried this first, but it causes an overflow error, at least on my system. The second return of `seeding.np_random` is a python int which can technically be as large as you want, and I don't know what are the practical bounds with the way we generate it. But for jax to be happy, the seed needs to fit in a C long, which leads to the `2**32-1` bound in the next line.

gymnasium/experimental/functional_jax_env.py
RedTachyon: Currently the FuncEnv spec doesn't require an observation/action space, but this is probably something we should change now that everything else is fleshed out, I'll update it soon

gymnasium/experimental/functional_jax_env.py
RedTachyon: Do we actually need this? As far as I can tell, this would duplicate the functionality of the `HumanRendering` wrapper, so we might as well use the wrapper when it's necessary

gymnasium/experimental/functional_jax_env.py
RedTachyon: I think using the wrapper will be tricky at this moment, since the environment would have to be registered with a default wrapper, which isn't a thing we support at the moment afaik. Without the wrapper being applied, CI freaks out because it expects everything to be in numpy, which we should probably change as we get more jax stuff. But I should be able to use the function defined in that file.

pseudo-rnd-thoughts: The functions for the wrapper are separate so we should be able to use the functions without the wrapper.

RedTachyon: Done

pseudo-rnd-thoughts (approved):
Im uncertain on changing the tests/vector to use the experimental vector environments as I think we should be rewriting those tests but this works for now

PR #234, opened 2023-01-02
Add a tutorial for vectorized envs using A2C. 

till2:
Content:
- how to create vectorized envs? 
- using vectorized envs for domain randomization to train more robust agents.
- building an A2C agent and training it on LunarLander-v2 using vector envs


docs/tutorials/vector_envs_tutorial.py
```diff
@@ -0,0 +1,542 @@
...
+        Returns:
+            state_values: A tensor with the state values, with shape [n_envs,].
+            action_logits_vec: A tensor with the action logits, with shape [n_envs, n_actions].
+        """
+        x = torch.Tensor(x).to(self.device)
+        state_values = self.critic(x) # shape of state_value_vec: [n_envs,]
+        action_logits_vec = self.actor(x) # shape of action_logits_vec: [n_envs, n_actions]
+        return (state_values, action_logits_vec)
+
+    def select_action(self, x: np.ndarray) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
pseudo-rnd-thoughts, line 125: You return four elements not three

pseudo-rnd-thoughts (changes requested):
The tutorial looks good. 
You note that the purpose of the tutorial is on speeding up training.

Not sure how painful this would be but could you
1. Create a new graph with the time (I believe that `RecordEpisodes` wrapper includes this info) to replace for X time steps using different numbers of environments, 1, 3, 10, 32. Log the performance as well.
2. Compare the performance for different vector implementations, `AsyncVectorEnv`, `SyncVectorEnv`, + `EnvPool` or `SampleFactory` if you have time.

till2 (commented):
Everything incorporated now.

- fix type hints
- add performance analysis section

till2 (commented):
done.

pseudo-rnd-thoughts (approved):
New performance section looks good and interesting. 
For pre-commit, if you run it twice in a row does it fix the issue?

PR #240, opened 2023-01-05
Deprecate use of `np.bool_` in tests.

jjshoots:


RedTachyon:
Why do we want to do this?

RedTachyon:
Wait what? Testing now on numpy 1.23.3 I'm getting e.g.
```python
isinstance(np.bool_(True), bool) == False
```

As far as I remember, the deprecation was that `np.bool` was aliased to `bool` and is now deprecated. `np.bool_` and `bool` are still different.

pseudo-rnd-thoughts:
The numpy docs agree with @RedTachyon 

https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_

jjshoots:
Interesting... Probably 2 am disillusionment then. Closing this.

PR #241, opened 2023-01-05
Allow sequence to accept stacked np arrays if the feature space is Box

jjshoots:
# Description

This allows the Sequence space to accept numpy arrays of size `[n, *base_size]` if the feature space of the Sequence is `Box(shape=base_size)`.

Fixes # (issue)

### DIFF
--- a/gymnasium/spaces/sequence.py
+++ b/gymnasium/spaces/sequence.py
@@ -8,6 +8,7 @@ from typing import Any
 import numpy as np
 import numpy.typing as npt
 
+from gymnasium.spaces import Box
 from gymnasium.spaces.space import Space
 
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
Personally, I'm very uncertain of this implementation. 

I don't understand the difference between the new and old `contains` functions. Could you provide an example where the old would fail and it won't with new?

There are a couple of places in the code that need modifying if we want the whole space to support stacked sequences

jjshoots:
@pseudo-rnd-thoughts The simplest case where the old would fail is the `test_stacked_box` test.

pseudo-rnd-thoughts:
Could the `contains` function not just have `Sequence` changed to `Iterable`. 
Personally, I prefer the `contains` functions to check if the sample is positively contained within the space rather than negatively if that makes any sense.

I guess we can add this but @kad99kev will be looking into adding this feature to the whole space.

pseudo-rnd-thoughts (approved):

PR #246, opened 2023-01-07
ENH: implement RecordVideoV0

younik:
This PR implements RecordVideoV0 wrappers in experimental.

Notable changes w.r.t. `wrappers.RecordVideo`:

- Removed the saved json
- Removed `metadata["semantics.async"]` legacy support
- Removed vectorized env support (which anyway I don't think was working properly before); I think the plan is doing a custom version for Vec envs
- Changed semantics of episode id: now the counter is incremented when `reset` is called; previously the counter increased when `terminated or truncated` was true. Practically, if the user didn't call reset after it, a video could have been started at the wrong moment. On the other hand, now a video can continue after `terminated or truncated`, and it is stopped at `reset` instead.
- Now it is allowed to define both step and episode triggers.


tests/experimental/wrappers/test_record_video.py
```diff
@@ -1 +1,163 @@
...
+    env.reset()
+    for _ in range(10):
+        action = env.action_space.sample()
+        env.step(action)
+
+    assert env.recording
+    action = env.action_space.sample()
+    env.step(action)
+    assert not env.recording
+    env.close()
```
pseudo-rnd-thoughts, line 112: Add test that the video is saved and remove the file

tests/experimental/wrappers/test_record_video.py
```diff
@@ -1 +1,163 @@
...
+        [make_env("CartPole-v1", i, render_mode="rgb_array") for i in range(2)]
+    )
+    envs = gym.wrappers.RecordEpisodeStatistics(envs)
+    envs.reset()
+    for i in range(199):
+        _, _, _, _, infos = envs.step(envs.action_space.sample())
+
+        # break when every env is done
+        if "episode" in infos and all(infos["_episode"]):
+            print(f"episode_reward={infos['episode']['r']}")
```
pseudo-rnd-thoughts, line 158: Remove print

pseudo-rnd-thoughts (changes requested):
This looks really good and the testing seems fairly complete.
The comments are fairly minor then we should be able to merge

gymnasium/experimental/wrappers/rendering.py
```diff
@@ -79,9 +81,192 @@ def render(self) -> RenderFrame | list[RenderFrame] | None:
...
+    If neither :attr:`episode_trigger` nor ``step_trigger`` is passed, a default ``episode_trigger`` will be employed.
+    By default, the recording will be stopped once reset is called. However, you can also create recordings of fixed
+    length (possibly spanning several episodes) by passing a strictly positive value for ``video_length``.
+    """
+
+    def __init__(
+        self,
+        env: gym.Env,
+        video_folder: str,
+        episode_trigger: Callable[[int], bool] = None,
```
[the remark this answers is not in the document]
younik: `step_idx` is increased during `step`, thus `step_trigger` is checked only during `step`.
On the other hand, `episode_idx` is increased during `reset`, and `episode_trigger` is called there.
This means they cannot trigger at the same time, it can happen they trigger one after the other, in that case, the video will be just one frame.

younik (commented):
Thanks, @pseudo-rnd-thoughts for reviewing this; I fixed all your comments

tests/experimental/wrappers/test_record_video.py
```diff
@@ -1 +1,166 @@
...
+        env.step(action)
+
+    render_out = env.render()
+    assert isinstance(render_out, List)
+    assert len(render_out) == n_steps + 1
+    render_out = env.render()
+    assert isinstance(render_out, List)
+    assert len(render_out) == 0
+    env.close()
+    shutil.rmtree("videos")
```
younik, line 137: @pseudo-rnd-thoughts this test should now check what you asked

pseudo-rnd-thoughts (approved):

Issue #158, opened 2022-11-26
[Proposal] Env.metadata["render_modes"] type should be set


younik:
### Proposal

Allow `metadata["render_modes"]` to be a set.
Currently, there are two checks that forbid this. One during make, where `metadata["render_modes"]` is checked to be a `Sequence` (and a warn is shown otherwise) and the other in the `PassiveEnvChecker` https://github.com/Farama-Foundation/Gymnasium/blob/075267a55fe01671bcbb65c3034ce5c38081f015/gymnasium/utils/passive_env_checker.py#L344 where it is checked to be one between `list` or `tuple` and an error is thrown otherwise.
These checks should instead force `metadata["render_modes"]` to be an `Iterable`.

After allowing `set` in checks, we should make `metadata["render_modes"]` a `set` in our environments and in the examples in docs (people use them as blueprints for making new environments). 

### Motivation

Set is more appropriate for `metadata["render_modes"]` since order doesn't matter and repetitions don't make any sense. 

### Alternatives

Leave everything as it is. 

### Additional context

Needed changes can be seen here: https://github.com/Farama-Foundation/Gymnasium/compare/main...younik:Gymnasium:render-modes-set

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

PR #248, opened 2023-01-08
ENH: allow metadata["render_modes"] to be a set

younik:
Addresses https://github.com/Farama-Foundation/Gymnasium/issues/158


pseudo-rnd-thoughts (approved):
This breaks backward compatibility only if users try to get an index in the render modes. But this is a strange and minor issue. 
Therefore, looks good to me

RedTachyon:
I think this was merged a bit too quickly. I don't necessarily have a problem with allowing the metadata to be a set in tests, but otherwise it seems like a breaking change with no good reason. The performance impact is negligible, the only advantage seems to be that it makes a bit more sense in terms of the semantics.

As Mark said, this breaks compatibility if someone tries to do `env.metadata['render_modes'][0]`, which is a reasonable way of doing "give me any render mode". Or if you know that you only have one render mode and want to get its name without actually hard-coding it.
Another breaking aspect, which might very well turn out to be relevant, is that `json.dumps(env.metadata)` doesn't work anymore. Is it actually used now? I don't really know, but it seems like a reasonable thing to do in some cases.
I wouldn't be surprised if there are more subtle interactions like this

Also worth mentioning that this introduces a bug (and tests for it) which makes it so that:
```python
isinstance(
    gym.Env().metadata["render_modes"], 
    dict
)
```
because `{}` in python is an empty dict, not a set.

I'd strongly consider reverting this, we should really avoid breaking things for a very small gain.

younik:
Thanks for spotting this @RedTachyon 
Yes, I agree that changing the metadata of envs is a breaking change for some user code.
If we don't want this, maybe we can leave the env `render_modes` as list, but still allowing it to be a set (and using set in documentation). What do you think?

Issue #236, opened 2023-01-03
[Bug Report] High values of up-bounded ints are never sampled in spaces.Box


khoda81:
### Describe the bug

When int spaces are only up-bounded, `space.sample()` never returns the high value. I investigated this issue and it turns out to be related to this line:
https://github.com/Farama-Foundation/Gymnasium/blob/78b50ff6bc5b09be57a9a8c9f10c687fadb746b1/gymnasium/spaces/box.py#L221
The high value is missed when taking the floor of the continuously sampled values. This could be a fix:
```python
sample[unbounded] = np.round(sample[unbounded])
sample[low_bounded] = np.floor(sample[low_bounded])
sample[upp_bounded] = np.ceil(sample[upp_bounded])
sample[low_bounded] = np.round(sample[low_bounded])
```

### Code example

```shell
>>> space = gymnasium.spaces.Box(low=-float("inf"), high=0, shape=(), dtype=int)
>>> 0 in space
True
>>> 0 in [space.sample() for _ in range(10000)]
False
```

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
Thanks for the issue but Im still a bit confused.

With your example, it seems unlikely that you will ever sample `0` even with `dtype=int` with a lower bound of `-inf`.
If you don't use `-inf` as the lower bound is it now possible or what are the values that are generated from this space.

@RedTachyon Do you understand?

khoda81:
@pseudo-rnd-thoughts Thanks for the reply
> With your example, it seems unlikely that you will ever sample `0` even with `dtype=int` with a lower bound of `-inf`.
```python
>>> from collections import Counter
>>> from gymnasium.spaces import *
>>> space = gymnasium.spaces.Box(low=-float("inf"), high=0, shape=(), dtype=int)
>>> Counter(space.sample() for _ in range(10000))
Counter({-1: 6286,
         -2: 2289,
         -4: 324,
         -3: 891,
         -5: 115,
         -6: 59,
         -8: 5,
         -7: 24,
         -10: 2,
         -9: 4,
         -11: 1})
```
It is not unlikely its impossible that's the issue. Look at how many -1s there are. The reason I expect thisto be a bug is that by using exponential distribution 0's should be more likely than -1s. It works the other way though:
```python
[...]

PR #249, opened 2023-01-09
Fixed `Box.sample` bug for up-bounded discrete or boolean dtypes

khoda81:
# Description

- Implements ks-testing for continuous boxes.
- Adds `scipy` to testing dependency.

Related to #236 
Note: this does not implement test for int and bool boxed.

## Type of change

# Checklist:


pseudo-rnd-thoughts (commented):
Scipy has dropped support for python 3.7 in v1.7.3 therefore, I have updated pyproject.toml to pin scipy to this version 

FYI, for the discrete case then you need to check against the Geometric distribution I believe.

This is probably for another PR but given that scipy is installed in testing, it might be worth changing the sampling testing to use scipy kstest as well.

pseudo-rnd-thoughts:
@khoda81 Hey, any progress on this PR?

khoda81:
Sorry for the delay, I will be adding the ks-test for discrete spaces and a fix for discrete samples soon.

gymnasium/spaces/box.py
```diff
@@ -211,13 +211,14 @@ def sample(self, mask: None = None) -> NDArray[Any]:
...
-            + self.high[upp_bounded]
+            + high[upp_bounded]
         )
 
         sample[bounded] = self.np_random.uniform(
             low=self.low[bounded], high=high[bounded], size=bounded[bounded].shape
         )
-        if self.dtype.kind == "i":
+
+        if self.dtype.kind != "f":
```
pseudo-rnd-thoughts, line 221: Does this effect integer types as well? Why we only applying this to floats now not integers

tests/spaces/test_spaces.py
```diff
@@ -189,15 +199,105 @@ def _chi_squared_test(dim, exp_freq, obs_freq):
...
-            assert variance < 80.23209784876272
-        else:
-            assert variance < CHI_SQUARED[degrees_of_freedom]
+
+        assert variance < scipy.stats.chi2.isf(ALPHA, df=degrees_of_freedom)
     else:
         raise NotImplementedError(f"Unknown sample testing for {type(space)}")
 
 
+def ks_test(sample, low, high, bounded_below, bounded_above):
```
pseudo-rnd-thoughts, line 208: Add docstring to explain function

tests/spaces/test_spaces.py
```diff
@@ -189,15 +199,105 @@ def _chi_squared_test(dim, exp_freq, obs_freq):
...
+        sample = high - sample
+    else:
+        # X ~ N(0.0, 1.0)
+        dist = scipy.stats.norm
+
+    _, p_value = scipy.stats.kstest(sample, dist.cdf)
+    assert p_value >= ALPHA
+
+
+def chi2_test(sample, low, high, bounded_below, bounded_above):
```
pseudo-rnd-thoughts, line 230: Add docstring to explain function

tests/spaces/test_spaces.py
```diff
@@ -189,15 +199,105 @@ def _chi_squared_test(dim, exp_freq, obs_freq):
...
+    variance = np.sum(
+        np.square(expected_frequency - observed_frequency) / expected_frequency
+    )
+    degrees_of_freedom = len(observed_frequency) - 1
+    critical_value = scipy.stats.chi2.isf(ALPHA, df=degrees_of_freedom)
+
+    assert variance < critical_value
+
+
+def binary_chi2_test(sample, low, high, bounded_below, bounded_above):
```
pseudo-rnd-thoughts, line 278: Add docstrings

pseudo-rnd-thoughts (changes requested):
Thanks, this looks good to me. 
Could you add the docstrings and answer my question on `"f"`, then good to merge

gymnasium/spaces/box.py
khoda81: Yes, previously `floor` was applied only to integers. Now it is applied to everything that is not float (bools, ints and units). I may change this to be more explicit. This fixes bools not being `floor`ed and having only `True` values in the output (because anything other than `0` would be casted to `True` and when we have a continuous variable it is very unlikely for it to be `0`).

pseudo-rnd-thoughts (approved):

PR #254, opened 2023-01-11
Minor fixes for the resize observation wrapper

RedTachyon:
See #230

It got merged before I could look at it again, so just fixing two small annoyances:
1. `self.shape` was assigned twice
2. `2 <= dims <= 3` was replaced with `dims == 2 or dims == 3` to make it a bit obvious what the intention is


pseudo-rnd-thoughts (approved):
Sorry for not getting your approval on the previous PR

PR #259, opened 2023-01-11
Fix type annotations of `callable` to `Callable`

ianyfan:
# Description

This fixes all instances of annotations of `var: callable` to `var: Callable`. It also adds the relevant import from `collections.abc` or `typing` depending on the presence of the `__future__` import. I've also tried adding the parameters/return types, but they are fairly complex so I didn't manage for some of them.

## Type of change

Type annotations

# Checklist:

### DIFF
--- a/gymnasium/envs/registration.py
+++ b/gymnasium/envs/registration.py
@@ -45,7 +45,7 @@ ENV_ID_RE = re.compile(
 )
 
 
-def load(name: str) -> callable:
+def load(name: str) -> Callable:
     """Loads an environment with name and returns an environment creation function.
 
     Args:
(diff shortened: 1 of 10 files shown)


pseudo-rnd-thoughts (approved):
Could you explain the difference between `Callable` and `callable`
Also if you are in the discord, then say hi and I can make you a dev

ianyfan:
`Callable` is the type, whereas `callable` is a function that tests whether its argument is callable. [An object is callable if it either has a `__call__` method or it is an object constructor.](https://docs.python.org/3/library/functions.html#callable)

I think `callable(obj)` is equivalent to `isinstance(obj, Callable)` but I'm not 100% sure. (there may be a corner case I'm not aware of)

PR #263, opened 2023-01-13
BlackJack Jax implementation

balisujohn:
# Description

This Pull request adds a Jax-Based functional BlackJack Environment to Gymnasium. 

### Some Design Notes

So this adds a new environments folder called `jax_toy_text` which is intended to house all the new jax toytext environments. The existing jax cartpole inherits from `FunctionalJaxEnv` which is in the `experimental` folder, but `BlackJackJaxEnv` inherits from `JaxEnv` instead which is custom defined inside `conversion.py` inside the `jax_toy_text` folder. Before merging probably it is ideal to generalize `FunctionalJaxEnv` so it can support Jax BlackJack Env.

Right now the assets used for rendering in `BlackJackJaxEnv` are in the toy_text asset directories, to avoid duplication of assets. 

A human rendering test is included in the main function of `blackjack.py` in `jax_toy_text`. This is intended for testing out the environment/rendering, and will be removed before this is merged. 

The tests are in their owm new file called `test_jax_blackjack.py` due to being somewhat different than the cartpole and pendulum tests. These will likely need to be merged, and I think more tests for the actually blackjack functions would be good. 

image

Human rendering works, but it does seem like a window appears and instantly is closed before the first persistent window opens on calling the first render, so maybe there is a subtle bug there.

This is a draft PR, and while the environment works for testing purposes and the tests pass, I think more thought is needed before merging this.

## Type of change

# Checklist:

additional checklist:
- [x] Fix pydocstyle 
- [ ] Add test to make sure blackjack transitions are correct
- [ ] Think about how to combine tests
- [ ] Figure out why a pygame window appears for a moment in the human rendering `__main__`  function in blackjack.py, and  remove it once rendering seems to be working seamlessly.


RedTachyon:
One point about the rendering, I didn't go through the full PR just yet.

At the design stage, my intention was that the functional envs only generally expose the `rgb_array` rendering mode, since that fits in the functional paradigm. With that, we have the `HumanRendering` wrapper (or whatever is the exact name) that provides human rendering in a more modular manner.

Is there a problem with this approach for this env?

.pre-commit-config.yaml
```diff
@@ -51,7 +51,7 @@ repos:
     rev: 6.1.1
     hooks:
       - id: pydocstyle
-        exclude: ^(gymnasium/envs/box2d)|(gymnasium/envs/classic_control)|(gymnasium/envs/mujoco)|(gymnasium/envs/toy_text)|(tests/envs)|(tests/spaces)|(tests/utils)|(tests/vector)|(tests/wrappers)|(docs/)
+        exclude: ^(gymnasium/envs/box2d)|(gymnasium/envs/classic_control)|(gymnasium/envs/mujoco)|(gymnasium/envs/toy_text)|(gymnasium/envs/jax_toy_text)|(tests/envs)|(tests/spaces)|(tests/utils)|(tests/vector)|(tests/wrappers)|(docs/)
```
pseudo-rnd-thoughts, line 54: Is it possible to docstring all the jax toy text environments. 
We also need to come up with a better name than `jax_toy_text`, the new classic_control + mujoco environments are `phys2d` and `phys3d`.

gymnasium/envs/jax_toy_text/blackjack.py
```diff
@@ -0,0 +1,463 @@
...
+    If the player achieves a natural blackjack and the dealer does not, the player
+    will win (i.e. get a reward of +1). The reverse rule does not apply.
+    If both the player and the dealer get a natural, it will be a draw (i.e. reward 0).
+
+    ### Version History
+    * v1: Initial versions release (1.0.0)
+    """
+
+    action_space = spaces.Discrete(2)
+    observation_space = spaces.Tuple(
```
pseudo-rnd-thoughts, line 194: If we make this a v2 then we could use a `Box(low=0, high=[32, 11, 2], dtype=np.int32)`, avoiding the issues with tuples of int and keeping the array

pseudo-rnd-thoughts (commented):
Initial comments

.pre-commit-config.yaml
balisujohn: I think it wouldn't be unreasonable to call them 'tabular' since all toy_text environments are designed with tabular RL in mind.  And sounds good wrt adding the doc coverage.

pseudo-rnd-thoughts: I actually really like the name `tabular`, lets use that for now

balisujohn:
> One point about the rendering, I didn't go through the full PR just yet.
> 
> At the design stage, my intention was that the functional envs only generally expose the `rgb_array` rendering mode, since that fits in the functional paradigm. With that, we have the `HumanRendering` wrapper (or whatever is the exact name) that provides human rendering in a more modular manner.
> 
> Is there a problem with this approach for this env?

Fixed, and I changed the tester to use the `HumanRendering` wrapper.

.pre-commit-config.yaml
balisujohn: ok pydocstyle is now enabled for `tabular` and it passes pre-commit

gymnasium/envs/jax_toy_text/blackjack.py
balisujohn: now has a `np.int32` typed `Box` observation space.

balisujohn: the environment is registered as Jax-Blackjack-v0 since it's the first version of the jax blackjack env.

gymnasium/envs/tabular/blackjack.py
```diff
@@ -0,0 +1,477 @@
...
+from gymnasium.wrappers import HumanRendering
+
+
+RenderStateType = Tuple["pygame.Surface", str, int]  # type: ignore  # noqa: F821
+
+
+deck = jnp.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10, 10, 10])
+
+
+def obs_from_device(obs):
```
pseudo-rnd-thoughts, line 28: If we are using a box obs spaec, is there any need for this change?

gymnasium/envs/tabular/blackjack.py
```diff
@@ -0,0 +1,477 @@
...
+
+    return (dealer_hand, new_player_hand, dealer_cards, player_cards + 1), key
+
+
+def dealer_stop(val):
+    """This function determines if the dealer should stop drawing."""
+    return sum_hand(val[1]) < 17
+
+
+def draw_card_wrapper(val):
```
pseudo-rnd-thoughts, line 84: Why is this needed?

gymnasium/envs/tabular/blackjack.py
```diff
@@ -0,0 +1,477 @@
...
+        dealer_hand = hand_state[0]
+        player_hand = hand_state[1]
+        dealer_cards = hand_state[2]
+        player_cards = hand_state[3]
+
+        # note that only a bust or player action ends the round, the player
+        # can still request another card with 21 cards
+        done = (is_bust(player_hand) * action) + ((jnp.logical_not(action)) * 1)
+
+        new_state = (dealer_hand, player_hand, dealer_cards, player_cards, done)
```
pseudo-rnd-thoughts, line 234: Could we use a `NamedTuple` as this is make code later easier to read with `state.dealer_hand` rather than `state[0]`

We might need to use a jax specify thing to do this

gymnasium/envs/tabular/blackjack.py
```diff
@@ -0,0 +1,477 @@
...
+    obs, info = env.reset()
+    print(obs, info)
+    env.render()
+
+    terminal = False
+    while not terminal:
+        action = int(input("Please input an action\n"))
+        obs, reward, terminal, truncated, info = env.step(action)
+        print(obs, reward, terminal, truncated, info)
+        env.render()
```
pseudo-rnd-thoughts, line 475: This is not actually necessary. Technically, on each `step` if render_mode="human" then `render` is called within the environment

tests/experimental/functional/test_jax_blackjack.py
```diff
@@ -0,0 +1,136 @@
+"""Tests for Jax Blackjack functional env."""
+
+
+import jax
+import jax.numpy as jnp
+import jax.random as jrng
+import pytest
+
+from gymnasium.envs.tabular.blackjack import BlackjackFunctional  # noqa: E402
```
pseudo-rnd-thoughts, line 9: Why do we need to add the noqa?

pseudo-rnd-thoughts (changes requested):
It looks generally good. 
Im unsure on the number of sub-functions defined at the start of the file. To me, they don't actually help explain the function or how it fits into the whole environment. Im uncertain we should just move the functions to the main environment functions or if documentation will solve this issue

tests/experimental/functional/test_jax_blackjack.py
balisujohn: removed

gymnasium/envs/tabular/blackjack.py
balisujohn: removed

gymnasium/envs/tabular/blackjack.py
balisujohn: changed to named tuple, Jax actually knows how to flatten namedtuples I think without requiring any custom stuff(we will find out when profiling the jitted code maybe).

gymnasium/envs/tabular/blackjack.py
balisujohn: removed

gymnasium/envs/tabular/blackjack.py
balisujohn: it's a hack to allow the draw_card function to accept a tuple of it's arg since the while loop eval structure only allows a single object to get passed into the while loop body function, and the draw_card function is used elsewhere with multiple args, so it seem unreasonable to make draw_card take only a single tuple as an argument. https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html

balisujohn:
I think the advantage of keeping the sub functions outside of the environment class is that it's much easier for a reader to find the standard `FuncEnv` functions if the class isn't cluttered with custom functions. I'm not sure how the doc will be affected though, so I think we could also put the custom functions inside the Class if it makes a big difference in the doc.

pseudo-rnd-thoughts (commented):
I'm accepting this and merging it but there are a couple of things we need to think about for the future

1. We need to add more testing of the jax-based environment to ensure that the functions are not recompiled each time the function is called. Chex has a function, `assert_max_traces` for this
2. If this version replaces the numpy based version we need to assert equivalence before hand
3. Investigate if we can use `jumpy` instead of `jax`

PR #264, opened 2023-01-13
Doc improvement: add comment describing what convolve does

metric-space:
# Description
This is a suggestion for doc improvement.
The use of convolution in average calculation (in the blackjack tutorial) doesn't jump out unless you've used the operation in the past. Felt a comment might aid someone else a bit lost as to why  `np.convolve` is used  

## Type of change

# Checklist:

### DIFF
--- a/docs/tutorials/training_agents/blackjack_tutorial.py
+++ b/docs/tutorials/training_agents/blackjack_tutorial.py
@@ -296,15 +296,10 @@ for episode in tqdm(range(n_episodes)):
 # ------------------------------
 #
 
-# Each element of a 1-D convolution output is the dot product
-# between a corresponding portion of the first signal and the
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Thanks for the PR and comment. You are correct that the current code is weird and isn't well explained but I think the comment can be significantly simpler and not explain how convolve works. 
Could the comment just be next to the first case that says, "`code ...` is used to compute a rolling average of the data to provide a more smooth graph"

metric-space:
@pseudo-rnd-thoughts  What you said makes more sense than what I had. Do I go ahead and put that line in or would you do that?

pseudo-rnd-thoughts:
@metric-space Im happy for you to use this PR

pseudo-rnd-thoughts (approved):

PR #282, opened 2023-01-21
Change `GymV22Compatibility` to `GymV21Compatibility` 

pseudo-rnd-thoughts:
In order to de-couple shimmy and gymnasium, such that they are not core dependencies of both.
For backward compatibility, we register dummy / fake gym compatibility environments.

However, it was founded in shimmy that the v22 compatibility is actually for v21. 
Therefore, shimmy is updating to use `GymV21Compatibility` rather than `GymV22Compatibility`.

As the shimmy is not updated yet, we need to keep use the shimmy V0.22 environments but as soon as this is released then we can update shimmy and then need to fix this issue


jjshoots (commented):
Lgtm

jjshoots (approved):

PR #284, opened 2023-01-24
Add `Stack` to `Sequence` space - #217

### LINKED ISSUE #217: [Proposal] Add `stack` parameter to `Sequence` space
### Proposal

The `Sequence` space requires that the samples always be a tuple of the sub-spaces. 
However, if for example a `Box` is the sub-space then it is a bit weird to have example sample `= (np.array([0, 1, 2]), np.array([3, 4, 5]))` as it could be more easily stacked to `np.array([[0, 1, 2], [3, 4, 5]])`

### Motivation

_No response_

### Pitch

I would propose that we add a parameter to `Sequence` called `stack: bool = False` to be backward compatible where if `True` then the resulting samples would be stacked. It might be possible to achieve this already with the vector `concatenate` function. 

This would require that class type return needs modifying along with the `sample` and `contains` functions. 

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

kad99kev:
# Description

Added a `stack` parameter to stack samples from `Sequence`.

Fixes #217

## Type of change

# Checklist:


pseudo-rnd-thoughts:
I have solved a couple of issues, in particular, with `flatten` and `unflatten` for `Sequence(stack=True)` 
https://github.com/pseudo-rnd-thoughts/Gymnasium/tree/kad99kev/main

PR #285, opened 2023-01-24
Update error.py

MiChaelinzo:
# Description
In this version of the code, I've removed the APIError class and its subclasses as they are deprecated. I've combined the Unregistered, NamespaceNotFound, NameNotFound, and VersionNotFound classes into a single class called InvalidEnvRequest which has a parameter indicating the specific error. I've kept the UnregisteredEnv, DeprecatedEnv, RegistrationError, UnseedableEnv, DependencyNotInstalled, UnsupportedMode, ResetNeeded, ResetNotAllowed, and InvalidAction classes, but I've renamed them to be more appropriate. I've also kept the UnregisteredBenchmark class.

Optimization suggestions:

The APIError class is deprecated and will be removed in version 1.0. It is recommended to remove it or replace it with a different class.

The APIConnectionError and AuthenticationError classes inherit from APIError and are also deprecated, so they should also be removed or replaced.

The unicode and str methods can be simplified by using the str method.

The InvalidRequestError class could be simplified by removing the param attribute and just including it in the message string.

The class hierarchy for the Unregistered, UnregisteredEnv, NamespaceNotFound, NameNotFound, and VersionNotFound classes can be simplified.

The APIError, APIConnectionError, InvalidRequestError, and AuthenticationError classes are not used in the code and can be removed.

The APIError class is also deprecated and should be removed or replaced.

The UnregisteredEnv, NamespaceNotFound, NameNotFound, and VersionNotFound classes are very similar, they can be combined into a single class with a parameter indicating the specific error.

it is a good practice to use specific error message for each error class rather than using the same message for all.

It's recommended to use PEP 8 conventions for naming the classes and methods, for example, naming the classes with CamelCase and methods with snake_case.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gymnasium/error.py
+++ b/gymnasium/error.py
@@ -43,42 +43,52 @@ class RegistrationError(Error):
 
 class UnseedableEnv(Error):
     """Raised when the user tries to seed an env that does not support seeding."""
+import warnings
 
+class Error(Exception):
+    """Error superclass."""
+
+class RegistryError(Error):
+    """Raised when the user requests an item from the registry that does not exist."""
+
+class UnregisteredEnv(RegistryError):
+    """Raised when the user requests an env from the registry that does not exist."""
+
+class DeprecatedEnv(Error):
+    """Raised when the user requests an env from the registry with an older version number than the latest env with the same name."""
+
+class RegistrationError(Error):
+    """Raised when the user attempts to register an invalid env. For example, an unversioned env when a versioned env exists."""
+
+class UnseedableEnv(Error):
+    """Raised when the user tries to seed an env that does not support seeding."""
 
 class DependencyNotInstalled(Error):
     """Raised when the user has not installed a dependency."""
 
-
 class UnsupportedMode(Error):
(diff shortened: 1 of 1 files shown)


gymnasium/error.py
```diff
@@ -43,42 +43,52 @@ class RegistrationError(Error):
...
-    """Raised when given an invalid value for a probability."""
+    def __init__(self, message):
+        self.message = message
 
+    def __str__(self):
+        return self.message
 
-class InvalidBound(Error):
-    """Raised when the clipping an array with invalid upper and/or lower bound."""
+class UnregisteredBenchmark(RegistryError):
```
pseudo-rnd-thoughts, line 90: I don't believe that the benchmarks exist anymore

MiChaelinzo: Yeah my fault

pseudo-rnd-thoughts:
Thanks for the changes but these changes are intended to be made at 1.0 not now

PR #287, opened 2023-01-24
Update __init__.py

MiChaelinzo:
# Description

Instead of importing everything from the gym_notices.notices module, you can import only the notices attribute, which will reduce the amount of code being imported.

The os.environ["PYGAME_HIDE_SUPPORT_PROMPT"] = "hide" line can be moved to the gym_notices.notices module.

The if sys.platform.startswith("linux"): block can be refactored to use the platform module and check if the current platform is Linux.

You can use from gymnasium.core import (Env, Wrapper, ObservationWrapper, ActionWrapper, RewardWrapper) instead of importing each class individually.

__all__ can be defined at the end of the module after all the imports and definitions, this way it will be easy to see what is exposed by the module.

__version__ can be defined as a constant at the top of the module

You can use gym_notices.notices.notices.get(__version__) instead of notices.notices.get(__version__).

You can remove the exception catch in the try-except block, as it can hide unexpected errors.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gymnasium/__init__.py
+++ b/gymnasium/__init__.py
@@ -1,17 +1,27 @@
 """Root `__init__` of the gymnasium module setting the `__all__` of gymnasium modules."""
 # isort: skip_file
 
-from gymnasium.core import (
-    Env,
-    Wrapper,
-    ObservationWrapper,
-    ActionWrapper,
-    RewardWrapper,
-)
+import os
+import sys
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
This PR does not contribute anything

PR #289, opened 2023-01-24
Reorder functions and refactor `registration.py`

pseudo-rnd-thoughts:
`registration.py` has been deserving of an update to clean up the code, most of the functions are actually ok.

This PR largely reorders functions, renames variables and adds comments to functions for additional explanations

Additionally, this PR updates the documentation for the registration functions


gymnasium/envs/registration.py
```diff
@@ -32,63 +30,118 @@
...
+    "make",
+    "spec",
+    "pprint_registry",
+    # Minor Functions
+    "namespace",
+    "get_env_id",
+    "parse_env_id",
+    "find_highest_version",
+    "load_env",
+    "load_plugin_envs",
```
RedTachyon, line 59: I don't think we should export these, they're pretty much for internal use

RedTachyon: (from get_env_id to load_plugin_envs)

gymnasium/envs/registration.py
```diff
@@ -32,63 +30,118 @@
...
-    mod_name, attr_name = name.split(":")
-    mod = importlib.import_module(mod_name)
-    fn = getattr(mod, attr_name)
-    return fn
 
+    id: str
+    entry_point: EnvCreator | str
+
+    # Environment arguments
+    kwargs: dict = field(default_factory=dict)
```
RedTachyon, line 89: The reordering of arguments has the potential to be a breaking change in (hopefully) rare cases where people construct EnvSpec from positional arguments. I'm also realizing there's the same potential issue in the vector API PR, we need to be careful here

gymnasium/envs/registration.py
```diff
@@ -32,63 +30,118 @@
...
+
+    * **id**: The string used to create the environment with :meth:`gymnasium.make`
+    * **entry_point**: A string for the environment location, ``(import path):(environment name)`` or a function that creates the environment.
+    * **reward_threshold**: The reward threshold for completing the environment.
+    * **nondeterministic**: If the observation of an environment cannot be repeated with the same initial state, random number generator state and actions.
+    * **max_episode_steps**: The max number of steps that the environment can take before truncation
+    * **order_enforce**: If to enforce the order of :meth:`gymnasium.Env.reset` before :meth:`gymnasium.Env.step` and :meth:`gymnasium.Env.render` functions
+    * **autoreset**: If to automatically reset the environment on episode end
+    * **disable_env_checker**: If to disable the environment checker wrapper in :meth:`gymnasium.make`, by default False (runs the environment checker)
+    * **kwargs**: Additional keyword arguments passed to the environment during initialisation
```
RedTachyon, line 82: If we do reorder arguments, the docstring should be adapted to that

gymnasium/envs/registration.py
```diff
@@ -566,48 +539,56 @@ def make(
...
-    _kwargs = spec_.kwargs.copy()
-    _kwargs.update(kwargs)
-
-    if spec_.entry_point is None:
-        raise error.Error(f"{spec_.id} registered but entry_point is not specified")
-    elif callable(spec_.entry_point):
-        env_creator = spec_.entry_point
+            raise error.Error(f"No registered env with id: {env_name}")
+
+    assert isinstance(env_spec, EnvSpec)
```
RedTachyon, line 572: Add an error message

gymnasium/envs/registration.py
```diff
@@ -694,71 +675,93 @@ def make(
...
+    """
+    env_spec = registry.get(env_id)
+    if env_spec is None:
         ns, name, version = parse_env_id(env_id)
         _check_version_exists(ns, name, version)
         raise error.Error(f"No registered env with id: {env_id}")
     else:
-        assert isinstance(spec_, EnvSpec)
-        return spec_
+        assert isinstance(env_spec, EnvSpec)
```
RedTachyon, line 695: Add error message

RedTachyon (commented):
A few small comments and one important issue (EnvSpec argument order). 
We probably want to revert this specific thing, there's no real benefit other than aesthetics.
What we can do instead is deprecate using `EnvSpec` with positional arguments, and then at some point later reorder them how we please.

I mostly looked at the top-level (signature) differences, and the internals of the key public-facing functions, I'm assuming the implementation of smaller functions (`_check_whatever`) wasn't changed beyond renaming variables.

gymnasium/envs/registration.py
pseudo-rnd-thoughts: Ok, will remove

gymnasium/envs/registration.py
pseudo-rnd-thoughts: OK, will revert

gymnasium/envs/registration.py
pseudo-rnd-thoughts: Agreed

gymnasium/envs/registration.py
pseudo-rnd-thoughts: Will do

pseudo-rnd-thoughts (commented):
@RedTachyon I made a couple more changes, primarily to the tests

* Moved the `registration.py` testing to `tests/envs/registration`
* Updated the tests for `make`, `register` and `spec`
* Added tests for `pprint_registry`. Im a bit uncertain of the `pprint_registry` implementation if you see the `test_pprint_namespace` however I think a case like this is very unlikely

tests/wrappers/utils.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import gymnasium as gym
 
 
-def has_wrapper(wrapped_env: gym.Env, wrapper_type: type) -> bool:
+def has_wrapper(wrapped_env: gym.Env, wrapper_type: type[gym.Wrapper]) -> bool:
```
RedTachyon, line 6: Can you double check if the `type[gym.Wrapper]` annotation is actually correct? It seems like it makes sense, but I can't find anything about it right now.

RedTachyon (approved):
Should be good now, one type annotation that I'm not 100% certain about, but if you know it's correct then it should be fine

tests/wrappers/utils.py
pseudo-rnd-thoughts: I will remove it as it is not necessary for the testing

PR #296, opened 2023-01-27
Update experimental vector API to the new autoreset order (separates `step` and `reset` from the same `step`)

RedTachyon:
List of changes:
- Pre-commit fix (not sure why it's appearing in the diff, but it's already on main and I thought it'd already be on the branch)
- Jax vector implementations are now properly registered
- Jax vector implementations properly handle time limits
- Fixed some doctest CI failures
- Changed the autoreset order. Now when an episode ends in one of the environments, there will be one "fake" step taken where the observation and info come from the reset, and reward = terminated = truncated = 0
- Adjusted some tests to accommodate for the above


pseudo-rnd-thoughts (approved):
I can't see any issues with this we just need to update documentation and make the change clear in the release notes

PR #304, opened 2023-01-31
Add Comet integration

nerdyespresso:
# Description

Add blurb about Comet's integration for Gymnasium.


jkterry1:
@nerdyespresso I assumed that when the email thread said [sponsor] that meant that you would fill [sponsor] in with the sponsor description we had agreed to on the call. I'm otherwise happy with the description, but I'd like to retain that level of explicit of a mention that you guys are a sponsor in the note.

nerdyespresso:
That works.. I've updated it! Thanks for the quick review @jkterry1

jkterry1:
@pseudo-rnd-thoughts you can merge this when you're happy

pseudo-rnd-thoughts (approved):

PR #319, opened 2023-02-10
Add mypy support

valentin-cnt:
# Description

This PR add mypy in the `.pre-commit-config.yaml` and the `pyproject.toml` files. For now the only file checked by  mypy is `gymnasium/core.py` and the goal to add files incrementally and to drop pyright. The configuration added in `pyproject.toml` is meant for mypy to run locally, but it should be removed once every errors are gone.

Besides, some changes are added to `gymnasium/core.py`  to pass mypy. There are some `# type: ignore` because the way the class `Env` and the wrappers class raise errors on mypy.
The two main reasons are:
- `ObsType`/`WrapperObsType` and `ActType`/`WrapperActType`  conflict
- Public variables in `Env` that become private in the wrapper classes

## Type of change

# Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,6 +57,17 @@ repos:
           - --explain
           - --convention=google
         additional_dependencies: ["toml"]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v1.0.0'
+    hooks:
+      - id: mypy
+        files: 'gymnasium/core.py'
+        args:
+          - --python-version=3.8
+          - --platform=All
+          - --strict
+          - --follow-imports=silent
+          - --ignore-missing-imports
   - repo: local
     hooks:
       - id: pyright
@@ -68,14 +79,3 @@ repos:
         additional_dependencies: ["pyright"]
         args:
           - --project=pyproject.toml
-  - repo: local
-    hooks:
-      - id: mypy
-        name: mypy
-        entry: mypy
-        language: node
-        pass_filenames: false
-        types: [python]
-        additional_dependencies: ["mypy"]
-        args:
-          - --config-file=pyproject.toml
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,3 +68,14 @@ repos:
         additional_dependencies: ["pyright"]
         args:
           - --project=pyproject.toml
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: mypy
+        language: node
+        pass_filenames: false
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts (changes requested):
In discussions with @RedTachyon, we agree that we should be using pyright **or** mypy and not both. Therefore, before merging this we need to work out which is best for gymnasium.

Second, I see what you were mentioning about `WrapperObsType` etc causing issues. Is there anyway that we can specify the wrapper might return alternative types to the base environment that is valid type hinting?

valentin-cnt:
> Second, I see what you were mentioning about `WrapperObsType` etc causing issues. Is there anyway that we can specify the wrapper might return alternative types to the base environment that is valid type hinting?

The problem is that the type hinting is back propagating to the `Env` class. 

```
class ActionWrapper(Wrapper[ObsType, WrapperActType]):
    """Superclass of wrappers that can modify the action before :meth:`env.step`.

    If you would like to apply a function to the action before passing it to the base environment,
    you can simply inherit from :class:`ActionWrapper` and overwrite the method :meth:`action` to implement
    that transformation. The transformation defined in that method must take values in the base environment’s
    action space. However, its domain might differ from the original action space.
    In that case, you need to specify the new action space of the wrapper by setting :attr:`self.action_space` in
    the :meth:`__init__` method of your wrapper.

    Among others, Gymnasium provides the action wrappers :class:`ClipAction` and :class:`RescaleAction` for clipping and rescaling actions.
    """

    def __init__(self, env: Env[ObsType, ActType]):
        """Constructor for the action wrapper."""
        super().__init__(env)

    def step(
        self, action: WrapperActType
    ) -> tuple[ObsType, SupportsFloat, bool, bool, dict[str, Any]]:
        """Runs the :attr:`env` :meth:`env.step` using the modified ``action`` from :meth:`self.action`."""
        return self.env.step(self.action(action))

    def action(self, action: WrapperActType) -> ActType | WrapperActType:
        """Returns a modified action before :meth:`env.step` is called.

        Args:
            action: The original :meth:`step` actions

        Returns:
            The modified actions
        """
        raise NotImplementedError
```

For example here, with `def action(self, action: WrapperActType) -> ActType | WrapperActType:`, it means that:
- `ActionWrapper.action(action)` returns `ActType | WrapperActType` 
- `ActionWrapper.env.step(self.action(action))` only take `ActType` as type of parameters since `env` is `Env` class

pseudo-rnd-thoughts:
I am closing this for now as we are uncertain of mypy over pyright. This PR will be reopen if we are interested in adding it over pyright

PR #321, opened 2023-02-12
Type hint fixes and added __all__ dunder

howardh:
- Made some changes so that pyright doesn't think `Env.step()` is supposed to return `None`.
- `infos` was mistyped as a `list[dist]` when it should be `dict`. Fixed the type hint and verified that the types are correct in the unit tests.
- Exposed the `colorize` and `EzPickle` utils

### DIFF
--- a/gymnasium/vector/vector_env.py
+++ b/gymnasium/vector/vector_env.py
@@ -147,7 +147,9 @@ class VectorEnv(gym.Env):
             actions: The actions to take asynchronously
         """
 
-    def step_wait(self) -> Tuple[Any, NDArray[Any], NDArray[Any], NDArray[Any], dict]:
+    def step_wait(
+        self, **kwargs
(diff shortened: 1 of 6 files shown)


gymnasium/vector/vector_env.py
```diff
@@ -146,7 +147,7 @@ def step_async(self, actions):
             actions: The actions to take asynchronously
         """
 
-    def step_wait(self, **kwargs):
+    def step_wait(self) -> Tuple[Any, NDArray[Any], NDArray[Any], NDArray[Any], dict]:
```
pseudo-rnd-thoughts, line 150: I think we need the `**kwargs` for `AsyncVectorEnv`

pseudo-rnd-thoughts (changes requested):
Thanks these look good, could you revert the `**kwargs` issue then will merge

pseudo-rnd-thoughts (approved):

Issue #314, opened 2023-02-08
[Bug Report] Jax, pytorch and dependency imports when importing gymnasium


vmoens:
### Describe the bug

Importing Jax and PyTorch when doing 
```python
import gymnasium
```
(which occurs via imports in `gymnasium.experimental.wrappers` is the cause of multiple issues:
- It does not catch errors that are unrelated to an `ImportError`. If pytorch or jax have errors such as `OSError` or other we can't import gymnasium for reasons that are unrelated to `gymnasium` itself. PyTorch and Jax should not condition `gymnasium` usage.
- Besides, when running scripts in parallel with multiprocessing in `spawn` mode, importing gymnasium is an operation one would expect to be super fast and hence not to cause major slowdowns in code execution. However, `import torch` is a very slow operation, and `import gymnasium` will suffer from it.

## Solution

Any import to optional dependencies should be hidden from the general library loading. 

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
Jax and PyTorch are optional installs but I agree that we import them at the beginning of the file and will cause slow downs if the modules are not actually used.
Do you know what the best solution to this is?

RedTachyon:
We should probably rearrange what is imported in which `__init__`, currently around `experimental`, but when those things become stable, then throughout the repo. 

For now the only "leaking" part do seem to be the wrappers (`gymnasium/__init__.py` -> `gymnasium/experimental/__init__.py` -> `gymnasium/experimental/wrappers/__init__.py` -> all individual wrappers), so if we remove that from the exports, it should be enough.

More generally, I'd consider removing `experimental` from the root `__init__` in general - this way, it only gets imported if someone explicitly does `import gymnasium.experimental<...>`. Even then, we should keep the different dependencies modular, so that e.g. accessing `gymnasium.experimental.VectorEnv` doesn't touch any jax or pytorch stuff


FWIW this is what I meant when I was talking about making sure that none of the new experimental features things leak out to the main repo, but I didn't actually check the current state, so that's my bad. I can try to unfuck this before the next release. (though it might be as simple as that one import I mentioned earlier)

RedTachyon:
Closed via #323, please reopen if this is (still/again) an issue

PR #323, opened 2023-02-12
Properly hide additional requirements between separate imports

RedTachyon:
See #314.

This is semi-WiP, it should be ready to review, but I also want to have another look at it before it's merged.

The general idea here is that a few more explicit import statements might be necessary to use `experimental` features, but fewer things will be automatically brought into scope when doing `import gymnasium as gym`

The new barriers are as follows:
- importing `gymnasium` doesn't import `experimental` (since we can potentially do wild things there)
- importing `experimental` doesn't import `wrappers` (since wrappers 
- importing `wrappers` doesn't import any of the conversion wrappers (which carry extra dependencies)

All individual modules can still be imported manually, so this shouldn't block any functionality.


floringogianu:
Currently running some experiments with this pull request. It already solved some issues I had in my codebase because of the automatic loading of `jax/jumpy`.

pseudo-rnd-thoughts (approved):
I agree this is a temporal solution but we should work out a better solution for the future, in particular when these become default.

As you wish to review the PR before merging it, I won't merge until @RedTachyon is happy

RedTachyon:
I did my sanity checks, I'm gonna go ahead and merge.

One note - this is not meant to be a temporary solution, but very permanent and something we need to be paying attention to in the future. As we introduce more optional dependencies (jax in the main repository or whatever else), correctly separating the submodules should be one of the main ways of keeping them under control. The only additional thing I can come up with that would be helpful is some clear message like "To use the JaxToTorch wrapper you must have torch installed" when someone tries to import a submodule without the right dependencies, but that's not so important.

PR #324, opened 2023-02-12
Properly handle different kwargs in the new vector API

RedTachyon:
`**kwargs` and `vector_kwargs` were mixed up between the different vectorization modes, now they are not.


pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts (commented):
Use 
```
Note: 
    Info
``` 
For an sphinx note in the documentation, i.e., https://gymnasium.farama.org/api/env/#gymnasium.Env.render

PR #327, opened 2023-02-12
Fix type hints errors in gymnasium/spaces

valentin-cnt:
# Description

Removed `reportInvalidTypeVarUse` and `reportGeneralTypeIssues` with pyright and fixed the new raised issues.

Files that are covered without errors: 
- [X] __init__.py
- [ ] box.py
- [X] dict.py
- [X] discrete.py
- [ ] graph.py
- [X] multi_binary.py
- [x] multi_discrete.py
- [X] sequence.py
- [X] space.py
- [x] text.py
- [X] tuple.py
- [ ] utils.py

## Type of change

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/gymnasium/spaces/dict.py
+++ b/gymnasium/spaces/dict.py
@@ -52,7 +52,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
     def __init__(
         self,
         spaces: None | dict[str, Space] | Sequence[tuple[str, Space]] = None,
-        seed: int | np.random.Generator | None = None,
+        seed: dict | int | np.random.Generator | None = None,
         **spaces_kwargs: Space,
     ):
         """Constructor of :class:`Dict` space.
@@ -110,7 +110,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
             ), f"Dict space element is not an instance of Space: key='{key}', space={space}"
 
         # None for shape and dtype, since it'll require special handling
-        super().__init__(None, None, seed)
+        super().__init__(None, None, seed)  # type: ignore
 
     @property
     def is_np_flattenable(self):
--- a/gymnasium/spaces/dict.py
+++ b/gymnasium/spaces/dict.py
@@ -52,7 +52,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
     def __init__(
         self,
         spaces: None | dict[str, Space] | Sequence[tuple[str, Space]] = None,
-        seed: dict | int | np.random.Generator | None = None,
+        seed: int | np.random.Generator | None = None,
         **spaces_kwargs: Space,
     ):
         """Constructor of :class:`Dict` space.
@@ -117,7 +117,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return all(space.is_np_flattenable for space in self.spaces.values())
 
-    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int]:
+    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int | None]:
         """Seed the PRNG of this space and all subspaces.
 
         Depending on the type of seed, the subspaces will be seeded differently
@@ -129,7 +129,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
         Args:
             seed: An optional list of ints or int to seed the (sub-)spaces.
         """
-        seeds: list[int] = []
+        seeds: list[int | None] = []
 
         if isinstance(seed, dict):
             assert (
@@ -226,7 +226,9 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
             for key, space in self.spaces.items()
         }
 
-    def from_jsonable(self, sample_n: dict[str, list[Any]]) -> list[dict[str, Any]]:
+    def from_jsonable(
+        self, sample_n: dict[str, list[Any]]
+    ) -> list[OrderedDict[str, Any]]:
         """Convert a JSONable data type to a batch of samples from this space."""
         dict_of_list: dict[str, list[Any]] = {
             key: space.from_jsonable(sample_n[key])
--- a/gymnasium/spaces/dict.py
+++ b/gymnasium/spaces/dict.py
@@ -117,7 +117,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return all(space.is_np_flattenable for space in self.spaces.values())
 
-    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int | None]:
+    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int]:
         """Seed the PRNG of this space and all subspaces.
 
         Depending on the type of seed, the subspaces will be seeded differently
@@ -129,7 +129,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
         Args:
             seed: An optional list of ints or int to seed the (sub-)spaces.
         """
-        seeds: list[int | None] = []
+        seeds: list[int] = []
 
         if isinstance(seed, dict):
             assert (
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 from typing import Any, Iterable, Mapping, Sequence
 
 import numpy as np
-from numpy.typing import NDArray
 
 from gymnasium.spaces.space import MaskNDArray, Space
 
@@ -56,7 +55,7 @@ class Discrete(Space[np.int64]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> np.int64 | NDArray[np.int64]:
+    def sample(self, mask: MaskNDArray | None = None) -> np.int64:
         """Generates a single random sample from this space.
 
         A sample will be chosen uniformly at random with the mask if provided
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -56,9 +56,7 @@ class Discrete(Space[np.int64]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(
-        self, mask: MaskNDArray | None = None
-    ) -> np.int64 | NDArray[np.signedinteger[Any]]:
+    def sample(self, mask: MaskNDArray | None = None) -> np.int64 | NDArray[np.int64]:
         """Generates a single random sample from this space.
 
         A sample will be chosen uniformly at random with the mask if provided
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from typing import Any, Iterable, Mapping, Sequence
 
 import numpy as np
+from numpy.typing import NDArray
 
 from gymnasium.spaces.space import MaskNDArray, Space
 
@@ -55,7 +56,9 @@ class Discrete(Space[np.int64]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> int:
+    def sample(
+        self, mask: MaskNDArray | None = None
+    ) -> np.int64 | NDArray[np.signedinteger[Any]]:
         """Generates a single random sample from this space.
 
         A sample will be chosen uniformly at random with the mask if provided
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -58,7 +58,7 @@ class MultiBinary(Space[NDArray[np.int8]]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> NDArray[Any]:
+    def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.bool_ | np.int8]:
         """Generates a single random sample from this space.
 
         A sample is drawn by independent, fair coin tosses (one toss per binary variable of the space).
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -4,12 +4,12 @@ from __future__ import annotations
 from typing import Any, Sequence
 
 import numpy as np
-import numpy.typing as npt
+from numpy.typing import NDArray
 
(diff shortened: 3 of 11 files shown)


gymnasium/spaces/text.py
```diff
@@ -4,12 +4,12 @@
...
 
 import numpy as np
-import numpy.typing as npt
+from numpy.typing import NDArray
 
 from gymnasium.spaces.space import Space
 
 
-alphanumeric: frozenset[str] = frozenset(
+alphanumeric: set[str] = set(
```
pseudo-rnd-thoughts, line 12: Why change from frozenset to set?

gymnasium/spaces/tuple.py
```diff
@@ -47,7 +47,7 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return all(space.is_np_flattenable for space in self.spaces)
 
-    def seed(self, seed: int | typing.Sequence[int] | None = None) -> list[int]:
+    def seed(self, seed: int | typing.Sequence[int] | None = None) -> list[int | None]:
```
pseudo-rnd-thoughts, line 50: When is the seed `None`?

gymnasium/spaces/utils.py
```diff
@@ -184,15 +184,17 @@ def _flatten_multidiscrete(
 def _flatten_tuple(space: Tuple, x: tuple[Any, ...]) -> tuple[Any, ...] | NDArray[Any]:
     if space.is_np_flattenable:
         return np.concatenate(
-            [flatten(s, x_part) for x_part, s in zip(x, space.spaces)]
+            np.array([flatten(s, x_part) for x_part, s in zip(x, space.spaces)])
```
pseudo-rnd-thoughts, line 187: Why do we need to add `np.array`?

gymnasium/spaces/utils.py
```diff
@@ -184,15 +184,17 @@ def _flatten_multidiscrete(
...
         )
     return tuple(flatten(s, x_part) for x_part, s in zip(x, space.spaces))
 
 
 @flatten.register(Dict)
 def _flatten_dict(space: Dict, x: dict[str, Any]) -> dict[str, Any] | NDArray[Any]:
     if space.is_np_flattenable:
-        return np.concatenate([flatten(s, x[key]) for key, s in space.spaces.items()])
+        return np.concatenate(
+            np.array([flatten(s, x[key]) for key, s in space.spaces.items()])
```
pseudo-rnd-thoughts, line 196: Why add the `np.array`?

gymnasium/spaces/utils.py
```diff
@@ -354,7 +356,17 @@ def _unflatten_graph(space: Graph, x: GraphInstance) -> GraphInstance:
...
 
-    def _graph_unflatten(unflatten_space, unflatten_x):
+    def _graph_unflatten_nodes(
+        unflatten_space: Box | Discrete, unflatten_x: NDArray[Any]
+    ) -> NDArray[Any]:
+        if isinstance(unflatten_space, Box):
+            return unflatten_x.reshape(-1, *unflatten_space.shape)
+        elif isinstance(unflatten_space, Discrete):
+            return np.asarray(np.nonzero(unflatten_x))[-1, :]
+
```
pseudo-rnd-thoughts, line 366: Add an error

pseudo-rnd-thoughts (changes requested):

gymnasium/spaces/dict.py
```diff
@@ -117,7 +117,7 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return all(space.is_np_flattenable for space in self.spaces.values())
 
-    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int]:
+    def seed(self, seed: dict[str, Any] | int | None = None) -> list[int | None]:
```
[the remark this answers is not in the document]
valentin-cnt: Because the return of the function `seed` is encapsulated in a list

```
    def seed(self, seed: int | None = None) -> list[int | None]:
        """Seed the PRNG of this space and possibly the PRNGs of subspaces."""
        self._np_random, seed = seeding.np_random(seed)
        return [seed]
```

gymnasium/spaces/text.py
valentin-cnt: Because `charset: set[str] | str = alphanumeric,` in args, so it was either accepting `frozenstr[str]` as type or doing the uppder thing. 

Since frozen set is already used after 
```
        self._char_set: frozenset[str] = frozenset(charset)
        self._char_list: tuple[str, ...] = tuple(charset)
        self._char_index: dict[str, np.int32] = {
            val: np.int32(i) for i, val in enumerate(tuple(charset))
        }
        self._char_str: str = "".join(sorted(tuple(charset)))
```
I didn't know which solutions to pick

gymnasium/spaces/utils.py
valentin-cnt: Because with the list, it raises this error :

```
Argument of type "list[FlatType]" cannot be assigned to parameter "arrays" of type "ArrayLike" in function "concatenate"
```

I didn't know which one is better between `np.array` or `np.asarray`

gymnasium/spaces/utils.py
valentin-cnt: https://github.com/Farama-Foundation/Gymnasium/pull/327#discussion_r1104369738

gymnasium/spaces/tuple.py
valentin-cnt: If `seed = None` in args, then the `super().seed(seed)` / `space.seed(seed)` will return `[None]`, no ?

gymnasium/spaces/dict.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Ok, then I think we should change to 
```python
self._np_random, np_random_seed = seeding.np_random(seed)
return [np_random_seed]
```

gymnasium/spaces/multi_binary.py
```diff
@@ -58,7 +58,7 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> npt.NDArray[np.int8]:
+    def sample(self, mask: MaskNDArray | None = None) -> NDArray[Any]:
```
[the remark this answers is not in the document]
valentin-cnt: So `NDArray[np.bool_ | np.int8]` resolves it

gymnasium/spaces/discrete.py
```diff
@@ -55,7 +56,9 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> int:
+    def sample(
+        self, mask: MaskNDArray | None = None
+    ) -> np.int64 | NDArray[np.signedinteger[Any]]:
```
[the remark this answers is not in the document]
valentin-cnt: Because the line 93 returns a `NDArray` if I understand correctly

But it could be changed to `np.int64 | NDArray[np.int64]` to be more precise

gymnasium/spaces/text.py
pseudo-rnd-thoughts: I think that frozensets are better than the sets due to their immutability in particular for class attributes that shouldn't change after being assigned

gymnasium/spaces/dict.py
```diff
@@ -52,7 +52,7 @@ class Dict(Space[typing.Dict[str, Any]], typing.Mapping[str, Space[Any]]):
     def __init__(
         self,
         spaces: None | dict[str, Space] | Sequence[tuple[str, Space]] = None,
-        seed: dict | int | np.random.Generator | None = None,
+        seed: int | np.random.Generator | None = None,
```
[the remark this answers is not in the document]
valentin-cnt: Because `Space.__init__` doesnt allow `dict` for the `seed` args
What would be the use case for a `dict` as `seed` ?

gymnasium/spaces/tuple.py
pseudo-rnd-thoughts:
```python
>>> from gymnasium import Space
>>> space = Space()
>>> print(f'{space.seed(None)=}')
space.seed(None)=[136770165915633426937684726807567110199]
```

gymnasium/spaces/dict.py
```diff
@@ -226,7 +226,9 @@ def to_jsonable(self, sample_n: Sequence[dict[str, Any]]) -> dict[str, list[Any]
             for key, space in self.spaces.items()
         }
 
-    def from_jsonable(self, sample_n: dict[str, list[Any]]) -> list[dict[str, Any]]:
+    def from_jsonable(
+        self, sample_n: dict[str, list[Any]]
+    ) -> list[OrderedDict[str, Any]]:
```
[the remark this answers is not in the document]
valentin-cnt: Well that is the raised error: 
```
Expression of type "list[OrderedDict[str, Any]]" cannot be assigned to return type "list[dict[str, Any]]"
  "list[OrderedDict[str, Any]]" is incompatible with "list[dict[str, Any]]"
    TypeVar "_T@list" is invariant
      "OrderedDict[str, Any]" is incompatible with "dict[str, Any]"
```

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: Looking at the documentation - https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html
I think it is the "inside" part of the list that need to be arrays, i.e., we should just wrap the `flatten` part like `np.array(flatten(...))`

gymnasium/spaces/utils.py
valentin-cnt: Can you be more specific ?

gymnasium/spaces/tuple.py
valentin-cnt: Fixed

gymnasium/spaces/dict.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: As `Space` just calls `seed` with the passed seed value then the `seed` type should be equal to the type hint of `Dict.seed`.

gymnasium/spaces/utils.py
valentin-cnt: Having `np.concatenate(np.array(flatten(s, x[key]) for key, s in space.spaces.items())` fails in tests

gymnasium/spaces/dict.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Ok, fair

gymnasium/spaces/discrete.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Ok, I think the type hint is correct, we should just fix the return types then

gymnasium/spaces/utils.py
pseudo-rnd-thoughts:
```python
else:
   raise TypeError("Expected the ... ")
```

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: It should be `np.concatenate([np.array(flatten(s, x[key])) for key, s in space.spaces.items()])` in my head

gymnasium/spaces/dict.py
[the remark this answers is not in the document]
valentin-cnt: So, if I understand well, `Space.__init__` and `Space.seed` should allow the dict type too ?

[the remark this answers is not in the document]
valentin-cnt: So the `Space.__init__` should allow `dict` as type of `seed` ?

gymnasium/spaces/utils.py
valentin-cnt: Done, thanks

gymnasium/spaces/discrete.py
```diff
@@ -55,7 +56,7 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> int:
+    def sample(self, mask: MaskNDArray | None = None) -> np.int64 | NDArray[np.int64]:
```
pseudo-rnd-thoughts, line 59: I still think this should just be `np.int64`

gymnasium/spaces/multi_binary.py
```diff
@@ -58,7 +58,7 @@ def is_np_flattenable(self):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> npt.NDArray[np.int8]:
+    def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.bool_ | np.int8]:
```
pseudo-rnd-thoughts, line 61: What case causes `np.bool_` as sample return type as it should be the same type as `class MultiBinary(Space[NDArray[np.int8]]):`

valentin-cnt: Well, this comes from `np_random.integers` that has a wild possiblity of type return, like int, bool and ndarray. 
That is actually I don't know how to deal with. 
Since `np_random.integers` accepts several type in args, I guess there would be a way to force the return type of the function based on the type of the args.

I have actually the exact same proble in the `graph.py` and `utils.py` files: there are some functions that can accept multiples spaces types but when used with specific spaces, the type raised errors (like `flatten_space` in  `utils.py`.

That is already the same kind of problem I encountered with this part of code https://github.com/Farama-Foundation/Gymnasium/pull/327#discussion_r1104355842 and I ended up splitting the function into 2 functions to handle each type cases

valentin-cnt: For example, even if `self.n` has a static type `np.int64`, this line is raises an error
```
tmp: np.int64 = self.np_random.integers(self.n)

Expression of type "ndarray[Any, dtype[int64]]" cannot be assigned to declared type "int64"
  "ndarray[Any, dtype[int64]]" is incompatible with "int64"
```

gymnasium/spaces/discrete.py
valentin-cnt: https://github.com/Farama-Foundation/Gymnasium/pull/327#discussion_r1104678853

gymnasium/spaces/multi_binary.py
pseudo-rnd-thoughts: Can we cast the data in some way to fix this, otherwise, raise an issue in numpy.

pseudo-rnd-thoughts (commented):
I think this looks good except for the question on `integers` where we expect a `np.int64` but numpy doesn't allow this currently. If there is no way of solving this issue then we can merge with the current implementation.

gymnasium/spaces/multi_binary.py
valentin-cnt: I will try to look for solutions and do another PR

valentin-cnt:
> I think this looks good except for the question on `integers` where we expect a `np.int64` but numpy doesn't allow this currently. If there is no way of solving this issue then we can merge with the current implementation.

I think it is better to merge everything for now, I reverted the things I didn't solve. 
A lot of the type errors left in `gymnasium/spaces` are due to functioncs that accept multiple distinct signatures. So I will try to do another PR about it

pseudo-rnd-thoughts (approved):

Issue #320, opened 2023-02-11
[Bug Report] Documentation is duplicated on the website


valentin-cnt:
### Describe the bug

A lot of the documentation on the website contain duplicated information.

Example:
- https://gymnasium.farama.org/api/wrappers/reward_wrappers/#gymnasium.wrappers.NormalizeReward
- https://gymnasium.farama.org/api/wrappers/observation_wrappers/#gymnasium.wrappers.FlattenObservation
- https://gymnasium.farama.org/api/wrappers/misc_wrappers/#gymnasium.wrappers.StepAPICompatibility

This is due because Sphinx collects the docstring for the `class` and the `__init__` which often say the same thing.
One solution could be to move every relevant documentation to the `class` docstring and configure Sphinx not to take the `__init__`. (I'm not familiar with Sphinx but I would guess there's a way)

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
This is a good point, for now, as the wrappers will be replaced by the experimental ones I would prefer to update those. 

The question is what information should be stored in the class docstring and the constructor (`__init__`) docstring. The problem is that the first line of the constructor docstring is not obviously different from the class docstring. My ideas are

1. We could modify the `pydocstyle` checker on the constructor docstring specifically on wrappers such that the first line is not necessary. 
2. Agree on a simple and effective first line of the constructor for pydocstyle and the website
3. Modify sphinx to ignore the first line of the constructor of wrappers only.

valentin-cnt:
I agree, I will to look for a way to parse the constructor (`__init__`) docstring to keep only the Args part

PR #329, opened 2023-02-14
Add docstring parser to remove duplicate in Gymnasium website

valentin-cnt:
# Description

As stated in #320, the documentation Gymnasium website contains duplicated content that can be confusing. Both the docstring from the `Class` and the `__init__` are displayed, meaning there is often a sentence repeated twice.

To keep the documentation the same, I just added a parser to the Sphinx `conf.py` to remove any lines that are written before a `Args` in the docstring. 

### Example 
```
class TransformReward(RewardWrapper):
    """Transform the reward via an arbitrary function.

    Warning:
        If the base environment specifies a reward range which is not invariant under :attr:`f`, the :attr:`reward_range` of the wrapped environment will be incorrect.

    Example:
        >>> import gymnasium as gym
        >>> from gymnasium.wrappers import TransformReward
        >>> env = gym.make("CartPole-v1")
        >>> env = TransformReward(env, lambda r: 0.01*r)
        >>> _ = env.reset()
        >>> observation, reward, terminated, truncated, info = env.step(env.action_space.sample())
        >>> reward
        0.01
    """

    def __init__(self, env: gym.Env, f: Callable[[float], float]):
        """Initialize the :class:`TransformReward` wrapper with an environment and reward transform function :attr:`f`.

        Args:
            f: A function that transforms the reward
        """
        super().__init__(env)
        assert callable(f)
        self.f = f
```
Here the parser will remove the lines before `Args`, which are the whitespace and `Initialize the :class:TransformReward wrapper with an environment and reward transform function :attr:f.`.

### Important

The parser targets every docstring that has the tag `class`, which are the `Class` and `__init__` ones. This means that if you place an `Args` section in the `Class` docstring, the lines before it will be removed it too.

I think only one class had this problem, but it should be noted as bad practice for the future.

## Type of change

# Checklist:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -66,6 +66,20 @@ napoleon_custom_sections = [("Returns", "params_style")]
 autoclass_content = "both"
 autodoc_preserve_defaults = True
 
+
+# This function removes the content before the parameters in the __init__ function.
+# This content is often not useful for the website documentation as it replicates
+# the class docstring.
+def remove_lines_before_parameters(app, what, name, obj, options, lines):
+    if what == "class":
+        result = [i for i in lines if i.startswith(":param")]
+        lines[:] = result if result else lines
+
+
+def setup(app):
+    app.connect("autodoc-process-docstring", remove_lines_before_parameters)
+
+
 # -- Options for HTML output -------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,8 +72,11 @@ autodoc_preserve_defaults = True
 # the class docstring.
 def remove_lines_before_parameters(app, what, name, obj, options, lines):
     if what == "class":
-        result = [i for i in lines if i.startswith(":param")]
-        lines[:] = result if result else lines
+        # ":" represents args values such as :param: or :raises:
+        idx_to_keep = next(
+            (i for i, line in enumerate(lines) if line.startswith(":")), 0
+        )
+        lines[:] = lines[idx_to_keep:]
 
 
 def setup(app):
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts (commented):
@mgoulao Could you confirm that this is a good idea? 
Otherwise, looks good to merge

mgoulao:
Why dont we remove the docstring under `__init__` and move `Args` to the class docstring? 

**Example**
```
class TransformReward(RewardWrapper):
    """Transform the reward via an arbitrary function.
   
    Args:
        f: A function that transforms the reward 
  
    Warning:
        If the base environment specifies a reward range which is not invariant under :attr:`f`, the :attr:`reward_range` of the wrapped environment will be incorrect.

    Example:
        >>> import gymnasium as gym
        >>> from gymnasium.wrappers import TransformReward
        >>> env = gym.make("CartPole-v1")
        >>> env = TransformReward(env, lambda r: 0.01*r)
        >>> _ = env.reset()
        >>> observation, reward, terminated, truncated, info = env.step(env.action_space.sample())
        >>> reward
        0.01
    """

    def __init__(self, env: gym.Env, f: Callable[[float], float]):
        super().__init__(env)
        assert callable(f)
        self.f = f
```

pseudo-rnd-thoughts:
@mgoulao As we have pydocstyle in pre-commit and this requires minimal changes to the code base + ide will still work as expected

pseudo-rnd-thoughts (approved):
Generating the website locally, I don't see any issues with the implementation and I agree that it makes it easier to read. 

If there is an issue we can revert this change for a future release.

valentin-cnt:
> I'm just afraid that we are creating too much complexity and that might exist some edge cases that we are not seeing.

I agree with the statement because it is kinda a hack, but I didn't find a better way to fix it without changing the current docstring format. This can be still discussed to change the format of the docstring (removing `__init__` content or moving `Args` to the Class docstring), but then it should affect every .py files

PR #335, opened 2023-02-17
Fix CI due to shimmy bump

pseudo-rnd-thoughts:
Shimmy released v0.2.1, which caused the CI to break for the compatibility test. This fixes it


SiddarGu:
Ignore the vercel check for now

PR #341, opened 2023-02-20
Add deprecated wrapper error in gymnasium.experimental.wrappers

valentin-cnt:
# Description

This PR adds an error check when an user tries to use a deprecated wrapper.

## Type of change

### Screenshots

# Checklist:


gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,38 @@
...
+
+class DeprecatedWrapper(ImportError):
+    """Exception raised when an old version of a wrapper is imported."""
+
+    pass
+
+
+def __getattr__(wrapper_name):
+    """Raises a DeprecatedWrapper exception when an old version of a wrapper is imported, or an ImportError if the wrapper being imported does not exist."""
+    wrapper_module = __name__
```
pseudo-rnd-thoughts, line 99: We probably want to check that the wrapper name fits the expected style.

Additionally, if it doesn't match, can use `difflib` to find close matches. See https://github.com/Farama-Foundation/Gymnasium/blob/31277a8f5b66f74592ccd91873e70e21f529eb28/gymnasium/envs/registration.py#L207 for examples

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,38 @@
...
+        is_valid_version = True
+    except ValueError:
+        is_valid_version = False
+
+    global_dict = globals()
+
+    if is_valid_version:
+        for act_version_num in range(1000):
+            act_wrapper_name = f"{base}{act_version_num}"
+            if act_wrapper_name in global_dict:
```
pseudo-rnd-thoughts, line 115: Can we not find the wrapper in the global with the same `base`? Then compare version number

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,38 @@
...
+                if version_num < act_version_num:
+                    raise DeprecatedWrapper(
+                        f"{base}{version_num} is now deprecated, use {act_wrapper_name} instead.\n"
+                        f"To see the changes made, go to "
+                        f"https://gymnasium.farama.org/api/experimental/wrappers/#{wrapper_module}.{act_wrapper_name}."
+                    )
+
```
pseudo-rnd-thoughts, line 122: Raise a different error if the version is greater than the wrapper's version.

pseudo-rnd-thoughts (changes requested):

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: Done

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: Done

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: I just removed it and use plain name instead

RedTachyon:
I am **extremely** hesitant about adding the module `__getattr__` thing, it can make things very confusing

valentin-cnt:
> I am **extremely** hesitant about adding the module `__getattr__` thing, it can make things very confusing

What would be the best alternative?

pseudo-rnd-thoughts:
@RedTachyon I understand that. To avoid issues, all that I wish would happen if to raise a more helpful error. 
In all cases, I want an `AttributeError` to raise, just if the attribute name matches one of the wrappers that already exist then we link to the documentation which will include the changes made to the wrapper.

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,68 @@
...
+    if wrapper_name == "__wrapped__" or "__test__":
+        raise AttributeError(
+            f"module 'gymnasium.experimental.wrappers' has no attribute {wrapper_name}"
+        )
+
+    base_name = wrapper_name[:-1]
+    version = wrapper_name[-1]
+
+    # Get all wrappers that start with the base wrapper name
+    wrappers = [name for name in globals().keys() if name.startswith(base_name)]
```
pseudo-rnd-thoughts, line 123: I think we can use `__all__` instead of `globals`

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,68 @@
...
+
+    # If the wrapper does not exist, raise an ImportError
+    if not wrappers:
+        raise ImportError(
```
pseudo-rnd-thoughts, line 127: I think this should be the AttributeError at the beginning. 
If there are no wrappers with the same name then we should just raise an attribute error. 
This should reduce the special case at the beginning and cover all possible weird imports

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +86,68 @@
...
+            f"cannot import name '{wrapper_name}' from 'gymnasium.experimental.wrappers'"
+        )
+
+    # Get the latest version of the wrapper
+    latest_version = max([int(name[-1]) for name in wrappers])
```
pseudo-rnd-thoughts, line 132: I don't hope for this to happen but we should probably support version numbers over 10

pseudo-rnd-thoughts (changes requested):
Could we add some tests. If necessary, we can modify the NormalizeReward wrapper to v1, there was a major bug found in v26 that was found and updated but never really talked about. 

@RedTachyon With my suggested changes, does this make you any happier with the idea?

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: Sure, I hesitated with it since `__all__` is manually typed, but both works

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: It seems to pass

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: I see, but it would be handier in the future to add a symbol between the name and the version of the wrapper, like with the environments. So the version checking would be more straighfoward

gymnasium/experimental/wrappers/__init__.py
pseudo-rnd-thoughts: True, but we are only adding this for classes with a shortened import name, `gymnasiu.wrappers.NormalizeRewardV0` rather than `gymnasium.wrappers.normalize.NormalizeRewardV0`

gymnasium/experimental/wrappers/__init__.py
pseudo-rnd-thoughts: I understand that but we decided that we should only use camel case and not mix camel and snake case, i.e., `NormalizeReward_v0`

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -86,3 +87,70 @@
     "VectorRecordEpisodeStatistics",
     "VectorListInfo",
 ]
+
+
+class DeprecatedWrapper(ImportError):
```
pseudo-rnd-thoughts, line 92: Move to `errors.py`

gymnasium/experimental/wrappers/lambda_reward.py
```diff
@@ -93,7 +93,7 @@ def __init__(
         super().__init__(env, lambda x: np.clip(x, a_min=min_reward, a_max=max_reward))
 
 
-class NormalizeRewardV0(gym.Wrapper):
+class NormalizeRewardV1(gym.Wrapper):
     r"""This wrapper will normalize immediate rewards s.t. their exponential moving average has a fixed variance.
```
pseudo-rnd-thoughts, line 97: Could you add to the documentation, the information from this PR (https://github.com/openai/gym/pull/3152) to explain the decision to increase the version number

pseudo-rnd-thoughts (changes requested):
Given that we have already implement a `__getattr__` in experimental.wrappers, I think we should explore updating this PR.
If we are not returning the updated wrapper, just raising a more helpful issue then I don't think there should be an issue with this PR

valentin-cnt:
> Given that we have already implement a `__getattr__` in experimental.wrappers, I think we should explore updating this PR. If we are not returning the updated wrapper, just raising a more helpful issue then I don't think there should be an issue with this PR

Would it be more interesting to do a new proper one ?

pseudo-rnd-thoughts:
> Would it be more interesting to do a new proper one ?

What do you mean?

valentin-cnt:
> > Would it be more interesting to do a new proper one ?
> 
> What do you mean? A new PR? I don't mind

Yes sorry, I meant a new PR, instead of having dozens of commit that doesnt add much

pseudo-rnd-thoughts:
Gymnasium uses flatten PRs such that all of the commits are merged into a single commit with the PR's name
Therefore, the commit messages don't actually matter or how many commits there are in a PR

gymnasium/experimental/wrappers/lambda_reward.py
valentin-cnt: I wasn't sure how to properly write it

pseudo-rnd-thoughts:
```
Note: 
    In v0.27, NormalizeReward was updated as the forward discounted reward estimate was incorrect computed in Gym v0.25+. For more detail, read [#3154](https://github.com/openai/gym/pull/3152)
```

valentin-cnt: Thank you! Fixed

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -99,21 +101,65 @@
...
+
+    base_name = wrapper_name[:-num_digits]
+
+    # Get all wrappers that start with the base wrapper name
+    wrappers = [name for name in __all__ if name.startswith(base_name)]
+
+    # If the wrapper does not exist, raise an AttributeError
+    if not wrappers:
+        raise AttributeError(
+            f"cannot import name '{wrapper_name}' from 'gymnasium.experimental.wrappers'"
```
pseudo-rnd-thoughts, line 137: Could you change to `raise AttributeError(f"module {__name__!r} has no attribute {name!r}")`, this is the official python error

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -99,21 +101,65 @@
...
     Returns:
-        Wrapper
+        The specified wrapper.
+
+    Raises:
+        AttributeError: If the wrapper does not exist.
+        InvalidVersionWrapper: If the version is not a digit or is greater than the latest version.
+        DeprecatedWrapper: If the version is not the latest.
     """
-    if name in _wrapper_to_class:
```
pseudo-rnd-thoughts, line 113: Why are we removing this? Personally I think that we should first check if the full wrapper name exists in the dictionary of wrappers then do the future checks to see if the wrapper version is wrong etc

gymnasium/experimental/wrappers/__init__.py
```diff
@@ -99,21 +101,65 @@
...
+    if version < latest_version:
+        raise DeprecatedWrapper(
+            f"{wrapper_name} is now deprecated, use {latest_wrapper_name} instead.\n"
+            f"To see the changes made, go to "
+            f"https://gymnasium.farama.org/api/experimental/wrappers/#gymnasium.experimental.wrappers.{latest_wrapper_name}."
+        )
+    # Raise an InvalidVersionWrapper exception if the version is greater than the latest
+    elif version > latest_version:
+        raise InvalidVersionWrapper(
+            f"{wrapper_name} is the wrong version number, use {latest_wrapper_name} instead."
```
pseudo-rnd-thoughts, line 160: I think we should be using the official python style just with additional information, `raise AttributeError(f"module {__name__!r} has no attribute {name!r}, did you mean XXX")`

pseudo-rnd-thoughts (changes requested):

gymnasium/experimental/wrappers/__init__.py
valentin-cnt: The proposed solution handles the case where an user can mistype the version number like `ClipRewardV!`
```python
    if version < 0:
        raise InvalidVersionWrapper(
            f"{wrapper_name} is not a valid version number, use {latest_wrapper_name} instead."
        )
```

valentin-cnt: But I'm fine removing it if it's too much checks

pseudo-rnd-thoughts: I think the InvalidVersionWrapper is ok, just I think the function should first check the simplest case where the wrapper name is correct then move to the next case rather than we check the simplest case currently last

pseudo-rnd-thoughts:
Looking at the code, Im worried how complex it is compared to how complex I think it needs to be 

Personally, I would want the code to look closer to this 
The commented code is pseudocode of what I think the actual code should do
```python
# base_wrapper_names = [wrapper_name.remove(version) for wrapper in __all__]


def __getattr__(name: str):
    if name in _wrapper_to_class:
        import_stmt = f"gymnasium.experimental.wrappers.{_wrapper_to_class[name]}"
        module = importlib.import_module(import_stmt)
        return getattr(module, name)
    else:
        # add helpful error message if version number has changed
        # base_wrapper = name.remove(version)
        if base_wrapper in base_wrapper_names:
            # latest_wrapper = __all__[base_wrapper]
            raise DeprecatedWrapper(
                f"{name!r} is now deprecated, use {latest_wrapper!r} instead.\n"
                f"To see the changes made, go to "
                f"https://gymnasium.farama.org/api/experimental/wrappers/#gymnasium.experimental.wrappers.{latest_wrapper}"
            )
        else:
            raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

pseudo-rnd-thoughts (approved):
Amazing, thanks @valentin-cnt

PR #355, opened 2023-03-02
Update `EnvSpec` to improve backward compatibility

pseudo-rnd-thoughts:
# Description

As noted by https://github.com/DLR-RM/stable-baselines3/pull/1327#issuecomment-1445219678, there are significant backward compatibility issues with https://github.com/Farama-Foundation/Gymnasium/pull/292

This PR updates `EnvSpec` such that it is a "full description of the whole env". Using the `EnvSpec` in `make` is now identical to `make(str)` as before. Critically, the difference with `make` in v0.27 is that the `EnvSpec` changes with each wrapper applied, including the default wrappers. As a result, `env.unwrapped.spec`'s `EnvSpec` has all parameters disabled such that `gym.make(env.unwrapped.spec)` is equivalent to just the base environment. 

The advantage of this implementation is that users can define an environment, `env` and perfectly replicate each stage
```python

while env is not env.unwrapped:
    recreated_env = gym.make(env.spec)
    assert recreated_env == env  # (this line will fail due to the environment equivalent not having a custom function)
    assert recreated_env.spec == env.spec

    env = env.env
```

Additionally, this PR updates the testing for `make` such that we check for both strings and `EnvSpec`.

### DIFF
--- a/gymnasium/core.py
+++ b/gymnasium/core.py
@@ -266,6 +266,8 @@ class Wrapper(
         self._reward_range: tuple[SupportsFloat, SupportsFloat] | None = None
         self._metadata: dict[str, Any] | None = None
 
+        self._cached_spec: EnvSpec | None = None
+
     def __getattr__(self, name: str) -> Any:
         """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
         if name == "_np_random":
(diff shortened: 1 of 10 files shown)


pseudo-rnd-thoughts:
Note, this breaks backward compatibility for `TimeLimit` as previously users were able to do `TimeLimit(env)` if the `env.spec.max_episode_steps is not None` as internally `TimeLimit` would use this value. 
Due to the changes in this PR, we need to remove this functionality and personally, I don't think that it makes much sense before.

rodrigodelazcano (approved):
The PR looks ready to be merged. I've been testing it with Minari to recreate environments from json environment specs and it works perfectly. Thanks @pseudo-rnd-thoughts !

tests/envs/registration/test_env_spec.py
```diff
@@ -86,43 +84,6 @@ def test_env_spec_to_from_json(env_spec: EnvSpec):
     assert env_spec == recreated_env_spec
 
 
-def test_wrapped_env_entry_point():
```
RedTachyon, line 89: Why is this test removed?

gymnasium/core.py
```diff
@@ -279,31 +281,46 @@ def __getattr__(self, name: str) -> Any:
...
-            wrapper_spec = WrapperSpec(
-                name=self.class_name(),
-                entry_point=f"{self.__module__}:{type(self).__name__}",
-                kwargs=kwargs,
-            )
-
-            # to avoid reference issues we deepcopy the prior environments spec and add the new information
-            env_spec = deepcopy(env_spec)
-            env_spec.applied_wrappers += (wrapper_spec,)
+        if self._cached_spec is None:
```
RedTachyon, line 284: Minor comment for slight unspagghettification - it might be better to do
```python
if self._cached_spec is not None:
    return self._cached_spec

env_spec = self.env.spec
...
```

Basically just flip the condition and do an early return to remove one level of indentation from the whole code block.

RedTachyon (commented):
Two minor comments, but I wasn't able to look at this in proper depth. If it works for "normal" usage (i.e. with the branch installed you can create an env, get the spec, recreate it and do regular stuff without any obvious bugs or redundant warnings), and it works for Minari purposes, then looks good to me as well

tests/envs/registration/test_env_spec.py
pseudo-rnd-thoughts: I moved the tests to `test_make.py` I believe

gymnasium/core.py
pseudo-rnd-thoughts: Updated

PR #358, opened 2023-03-05
Add `Wrapper.__setattr__` to allow users to set attribute on the base environment not the top level wrapper 

pseudo-rnd-thoughts:


RedTachyon:
I'll need to look at this in significantly more detail, but one point - is it actually a bug? I didn't spend much time thinking about this specific piece of code, but the behavior seems roughly aligned with what I expected it to be. Wrappers can have their own attributes, and for convenience we add automatic forwarding of attributes via `__getattr__`. Trying to add a similar functionality for setting attributes is a potential can of worms we might not want to get into.

Overall I'm not convinced right now that we want to do this at all. The one use I do see is with AsyncVecEnv, but that seems to just require one additional helper method.

pseudo-rnd-thoughts:
I generally agree with your view @RedTachyon 

I disagree with you that this is expected behaviour, you might be right that this might not be a bug but at least if you don't understand the attribute forward then it does look like a bug to me.

The simplest solution is to forward to modify the `VectorEnv.setattr` function to enable specifying at what "level" the attribute should be modified

WasabiFan:
I'd love to see this considered again. I have just gone through the same confusing discovery process as the linked issue: try to set a property, property doesn't work, turns out it's being eaten by the wrapper. And it doesn't seem that there's a straightforward alternative when using AsyncVectorEnv.

pseudo-rnd-thoughts:
@WasabiFan Hey, yes, we were planning on adding that to the `VectorEnv`, would you be able to add this feature on the `AsyncVectorEnv` or `SyncVectorEnv`?

PR #362, opened 2023-03-06
Update third_party_environments.md

markub3327:
# Description

Hello,

   the original `flappy-bird-gym` is outdated. It's not compatible with the latest Gymnasium. It doesn't implement:

* Truncation as the part of tuple returned by `env.step()`
* Info as part of tuple returned by `env.reset()`
* `render_mode` option
* `seed` and `options` in `env.reset()`
* native using Gymnasium instead of OpenAI Gym

Thanks.

### DIFF
--- a/docs/environments/third_party_environments.md
+++ b/docs/environments/third_party_environments.md
@@ -12,7 +12,7 @@ There are a number of Reinforcement Learning environments built by authors not i
 
 Supported fork of gym-retro with additional games, states, scenarios, etc. Open to PRs of additional games, features, and platforms since gym-retro is no longer maintained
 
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (approved):
Thanks for support gymnasium. 
Quick note on your readme example, you are using `done` incorrect, see https://gymnasium.farama.org/ for a correct implementation. We are hoping to have a blog post explaining the new step api soon

markub3327:
@pseudo-rnd-thoughts 

Sorry I changed the `done` to `terminated`. It signals that the bird is still alive and defines the endpoint of MDP. The `truncated` signal is not used because that has not a time limit.

Thanks.

pseudo-rnd-thoughts:
@markub3327 No worries, thanks for updating

PR #368, opened 2023-03-08
Use SupportsFloat in step_api_compatibility

software-samurai:
The problem is that gym.Env defines step return as Tuple[..., SupportsFloat, ...], and this then can't be passed into functions of `step_api_compatibility`. Alternative to my fix would be to define centrally what is a step return, and reuse the type alias in both places.

## Type of change

### DIFF
--- a/gymnasium/utils/step_api_compatibility.py
+++ b/gymnasium/utils/step_api_compatibility.py
@@ -1,5 +1,5 @@
 """Contains methods for step compatibility, from old-to-new and new-to-old API."""
-from typing import Tuple, Union
+from typing import Tuple, Union, SupportsFloat
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for fixing that, could you fix the pre-commit issue. See contribution.md for more information

RedTachyon:
Shouldn't we be starting to deprecate or remove the step api compatibility thing? It makes sense in gym, but technically speaking, Gymnasium never had a different step API

pseudo-rnd-thoughts:
> Shouldn't we be starting to deprecate or remove the step api compatibility thing? It makes sense in gym, but technically speaking, Gymnasium never had a different step API

Yes, but I think we should remove it when we remove the compatibility wrappers in v0.29

pseudo-rnd-thoughts (approved):

PR #403, opened 2023-03-24
Import `vector` in `gymnasium.__init__`

pseudo-rnd-thoughts:
For some reason, `import gymnasium` followed by `gymnasium.vector` fails as the module is not imported in `gymnasium/__init__.py`. More alarming this is not picked up by the CI


RedTachyon (approved):
Simple enough, but I don't know if it goes far enough/solves everything

RedTachyon:
Is there a reason why that shouldn't be `from gymnasium import envs, spaces, utils, vector, wrappers, error, logger` as from before #392?

PR #458, opened 2023-04-20
Update frozen_lake.py

Yanipo5:
# Description

Set max_episode_steps as 200 for 8x8 maps (currently 100 steps)

Fixes # (issue)

### Screenshots

image

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .cache
 .pytest_cache/
 __pycache__/
+.npmrc
 
 # Setuptools distribution and build folders.
 /dist/
@@ -37,6 +38,7 @@ tutorial/*.html
 # IDE files
 .eggs
 .tox
+.vscode
 
 # PyCharm project files
 .idea
--- a/gymnasium/envs/toy_text/frozen_lake.py
+++ b/gymnasium/envs/toy_text/frozen_lake.py
@@ -230,8 +230,8 @@ class FrozenLakeEnv(Env):
         self.nrow, self.ncol = nrow, ncol = desc.shape
         self.reward_range = (0, 1)
 
-        if self.nrow > 4:
-            self.max_episode_steps = 200
+        if self.spec and self.nrow > 4:
+            self.spec.max_episode_steps = 200
 
         nA = 4
         nS = nrow * ncol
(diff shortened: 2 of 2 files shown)


pseudo-rnd-thoughts:
Hey, why do you make this change? 
If users want to they can change the number of time steps for an environment with `gym.make("FrozenLake", max_episode_steps=200)`?

Yanipo5:
@pseudo-rnd-thoughts 
> Hey, why do you make this change? If users want to they can change the number of time steps for an environment with `gym.make("FrozenLake", max_episode_steps=200)`?

Cause the 8x8 => 200 is part of the specs,
Therefore it should be the default, no so?

> Episode End
> ...
> The length of the episode is 100 for 4x4 environment, 200 for 8x8 environment.

RedTachyon:
This is more of a documentation issue, as the two versions of the environment registered by default use 100 and 200 steps, for 4x4 and 8x8 respectively. For custom scenarios, people are free to set their own time limit.

https://github.com/Farama-Foundation/Gymnasium/blob/01308340aec7dd487c714e627d9f248eb646541b/gymnasium/envs/__init__.py#L136-L150

It might be useful to change the wording in the documentation to be a bit clearer, but I don't think this warrants a code change.

pseudo-rnd-thoughts:
Ahh, apologies. 

`FrozenLake8x8-v1` exists as the implementation of 8x8 and has `max_episode_steps=200` already so this PR is not necessary.  
 https://github.com/Farama-Foundation/Gymnasium/blob/01308340aec7dd487c714e627d9f248eb646541b/gymnasium/envs/__init__.py#L144

However, reading the documentation for [frozen lake](https://gymnasium.farama.org/environments/toy_text/frozen_lake/) so we need to fix this. Would you mind updating the documentation on [line](https://github.com/Farama-Foundation/Gymnasium/blob/01308340aec7dd487c714e627d9f248eb646541b/gymnasium/envs/toy_text/frozen_lake.py#L84) to include details about the 8x8 version

Yanipo5:
OK, I'll update the docs.
Though I have to say, my experience was that I didn't know about the 8x8 env variant, and I used "FrozenLike-v1" with desc="8x8".
Therefore I assumed it was playing 200 steps, which I was "surprised to learn was actually 100. 

Well maybe it's fine if we make it explicit in the docs.

PR #459, opened 2023-04-20
forzen lake Truncation doc

Yanipo5:
# Description

Fixes # (issue)

## Type of change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .cache
 .pytest_cache/
 __pycache__/
+.npmrc
 
 # Setuptools distribution and build folders.
 /dist/
@@ -37,6 +38,7 @@ tutorial/*.html
 # IDE files
 .eggs
 .tox
+.vscode
 
 # PyCharm project files
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 .cache
 .pytest_cache/
 __pycache__/
-.npmrc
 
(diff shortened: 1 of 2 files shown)


Kallinteris-Andreas:
1) do not change the `.gitignore`
2) what is the benefit of the change? You are likely to create confusion on how to use `make`.

pseudo-rnd-thoughts:
@Kallinteris-Andreas There is a large issue that the FrozenLake documentation doesn't note the existence of the `FrozenLake8x8` so we need to do a larger rewrite of the documentation. 
Additionally, I was worried that similar issues existed for other environments but in checking, I can't find another issue

Yanipo5:
>what is the benefit of the change? You are likely to create confusion on how to use make.
I think it is currently confusing. @Kallinteris-Andreas 

My experience was that I thought that by selecting the "8X8" map, I was granted 200 steps.
I had to debug the running environment to discover I actually had 100 steps on a 8X8 map.

My suggestion is either change the default to be if n_row > 3 ==> steps = 200 (#458)
Or specify that this requires the `FrozenLake8x8-v1`
Else, it is already confusing...

So what's the verdict here?

Kallinteris-Andreas:
as sorry got confused,
LTGT, besides the edit to `.gitignore`

Yanipo5:
Call I'll remove the .gitignore

pseudo-rnd-thoughts (approved):

PR #466, opened 2023-04-25
Skip Jax and PyTorch tests if module is missing

ChristofKaufmann:
# Description

Jax and PyTorch are optional dependencies. As such, tests importing them, should also be optional. This PR skips these tests, if the module is missing. This is an issue, because Jax is not available for Windows from conda-forge or PyPI and PyTorch is not available for Windows from conda-forge. A similar patch has been applied in https://github.com/conda-forge/gymnasium-feedstock/pull/28.

The same actually holds for OpenCV, which is required in [this test](https://github.com/Farama-Foundation/Gymnasium/blob/main/tests/wrappers/test_atari_preprocessing.py#L51). However, OpenCV is well available and skipping the test is not so easy, because it throws in the pytest parametrization.

## Type of change

  → Enhances windows compatibility of test suite.

# Checklist:

  → Not required.
  → This fixes (skips) tests if they are not applicable.

### DIFF
--- a/tests/envs/functional/test_jax.py
+++ b/tests/envs/functional/test_jax.py
@@ -1,11 +1,13 @@
-import jax
-import jax.numpy as jnp
-import jax.random as jrng
-import numpy as np
 import pytest
 
-from gymnasium.envs.phys2d.cartpole import CartPoleFunctional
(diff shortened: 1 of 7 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the changes

PR #480, opened 2023-05-02
Skip Mujoco tests if OpenGL is not available

ChristofKaufmann:
# Description

This PR skips the Mujoco tests in case OpenGL is not available, like in a docker container, see e.g. the section "MuJoCo OpenGL Errors" in [this comment](https://github.com/conda-forge/gymnasium-feedstock/pull/28#issuecomment-1520428770).

When selecting OSMesa by setting the environment variable `MUJOCO_GL=osmesa`, this does not lead to an `ImportError` from Mujoco, but to an `AttributeError` from OpenGL, if OSMesa is not available. This PR adds `AttributeError` to the expected errors, when making the test environments. It triggers a warning, but does not lead to a skipped test. This behavior is analog to the case when EGL is selected, but not available.

Additionally this PR tests if an OffScreenViewer object can be instantiated. It fails when setting the environment variable `MUJOCO_GL=glfw`, if GLFW is not available. In this case, the commit skips the rendering tests with Mujoco environments.

## Type of change

  → Enhances portability of test suite.

# Checklist:

  → Not required.
  → This fixes (skips) tests if they are not applicable.

### DIFF
--- a/bin/all-py.Dockerfile
+++ b/bin/all-py.Dockerfile
@@ -26,6 +26,7 @@ ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/root/.mujoco/mujoco210/bin"
 COPY . /usr/local/gymnasium/
 WORKDIR /usr/local/gymnasium/
 
-RUN pip install .[all,testing] --no-cache-dir
+# Test with PyTorch CPU build, since CUDA is not available in CI anyway
+RUN pip install .[all,testing] --no-cache-dir --extra-index-url https://download.pytorch.org/whl/cpu
 
 ENTRYPOINT ["/usr/local/gymnasium/bin/docker_entrypoint"]
--- a/tests/envs/test_rendering.py
+++ b/tests/envs/test_rendering.py
@@ -62,13 +62,19 @@ def test_render_modes(spec):
         if mode != "human":
             new_env = spec.make(render_mode=mode)
 
-            new_env.reset()
-            rendered = new_env.render()
-            check_rendered(rendered, mode)
+            try:
+                new_env.reset()
+                rendered = new_env.render()
+                check_rendered(rendered, mode)
 
-            new_env.step(new_env.action_space.sample())
-            rendered = new_env.render()
-            check_rendered(rendered, mode)
-
-            new_env.close()
+                new_env.step(new_env.action_space.sample())
+                rendered = new_env.render()
+                check_rendered(rendered, mode)
+            except Exception as e:
+                if "gladLoadGL error" in str(e):
+                    pytest.skip("OpenGL not available")
+                else:
+                    raise
+            finally:
+                new_env.close()
     env.close()
--- a/tests/envs/utils.py
+++ b/tests/envs/utils.py
@@ -22,6 +22,7 @@ def try_make_env(env_spec: EnvSpec) -> Optional[gym.Env]:
             return env_spec.make(disable_env_checker=True).unwrapped
         except (
             ImportError,
+            AttributeError,
             gym.error.DependencyNotInstalled,
             gym.error.MissingArgument,
         ) as e:


pseudo-rnd-thoughts:
I'm rerunning the CI but the error was `free(): invalid pointer` which isn't good. do we need to free / del the model, data and offscreen viewer?

ChristofKaufmann:
Yes, it is really bad and I don't understand why its happening. I can't reproduce it locally, yet. I am also not sure, whether I am doing something very wrong or just triggered an existing bug.

As a totally different alternative to catch OpenGL errors, I could also catch `mujoco.FatalError`s around `rendered = new_env.render()` in line [66](https://github.com/Farama-Foundation/Gymnasium/pull/480/files#diff-f1b56457af3165f520dff35591f892ca21871bd94583b261f1c46fb4ebec1503L66) and [70](https://github.com/Farama-Foundation/Gymnasium/pull/480/files#diff-f1b56457af3165f520dff35591f892ca21871bd94583b261f1c46fb4ebec1503L70) and reraise if the message is not `"gladLoadGL error"`.

What do you think is the best way to go forward here?

ChristofKaufmann:
I looked around a bit:

- I suspected [`glfw.terminate()`](https://github.com/Farama-Foundation/Gymnasium/blob/2feeb31378942ff3373915a4fe31faffb0d2b26c/gymnasium/envs/mujoco/mujoco_rendering.py#L278) in the `OffScreenViewer` because it does not initialize GLFW at all. I think it is a copy and paste error, but without effect. The documentation of GLFW [states](https://www.glfw.org/docs/3.3/group__init.html#gaaae48c0a18607ea4a4ba951d939f0901): "This function has no effect if GLFW is not initialized."
  → Should not hurt.
- When using OSMesa calling [`opengl_context.free()`](https://github.com/Farama-Foundation/Gymnasium/blob/2feeb31378942ff3373915a4fe31faffb0d2b26c/gymnasium/envs/mujoco/mujoco_rendering.py#L188) results in a call to Mujocos `free` function and there the [`osmesa.OSMesaDestroyContext(self._context)`](https://github.com/deepmind/mujoco/blob/42a16f65c2814249f4272aefc87e8821ccd4e9ae/python/mujoco/osmesa/__init__.py#L76) call is somehow not inside the if-branch. So it might be called with a `None` argument. The "documentation" (source code comments) don't say, what is happening in this case. To me the [source code](https://github.com/redox-os/mesa/blob/d90622910ba2d726b6f43ade0a364d0eb04dd3bb/src/mesa/drivers/osmesa/osmesa.c#L945) looks ok, if `None` is translated to `nullptr`, which I am not certain of.
  → **Candidate.**
- My try-except block might be unsafe (should not be), depending on where an exception is raised during the `render` call. However, in the failing test no exception is thrown, so it should be properly initialized and closed.
  → Should be fine.
- I might have overlooked something.
  → **Possibly more candidates!**

pseudo-rnd-thoughts:
I tested in https://github.com/Farama-Foundation/Gymnasium/pull/485 if `del` all of the objects works which it didn't however I believe that `del` isn't actually guaranteed to cause the object to be deleted.

Like you said, it would be helpful to know what object is calling `free` and causing the issue. 
Testing locally on a mac hasn't raised the issue either. 
Im going to see if we can identify the object that causes the issue, either `model`, `data` or `mjr`. Hopefully it is not a combination of them

pseudo-rnd-thoughts:
Ok, bad news. I simplified the CI to only run `tests/envs/test_rendering.py` and can't replicate the issue. This means that somehow the new code is interacting with the old code in a bad way.

ChristofKaufmann:
I saw your tests. Thank you for investigating. Very nice. However, I am not sure about your conclusion. From segmentation errors, I know that moving any unrelated part of code can hide or trigger them. These "free() invalid pointer" errors may behave similarly: I found an [issue](https://github.com/deepmind/mujoco/issues/644) with pytorch, where the order of imports triggered it.

ChristofKaufmann:
I could reproduce the issue in the docker container that is used in the build job. I think it is indeed a PyTorch issue, maybe [this](https://github.com/pytorch/pytorch/issues/15294) one. This should not happen with a conda package, since they have a proper tool chain for binaries.

In the build container, the order of imports triggers the bug indeed. As an example, the following script does not yield an error:

```python
from mujoco.glfw import GLContext
import torch
ctx = GLContext(100, 100)
```

but this one fails with the "free(): invalid pointer" error:

```python
from mujoco.glfw import GLContext
ctx = GLContext(100, 100)
#ctx.free() does not make a difference
import torch
```

The GLFW Context is made in the `OffScreenViewer`. I guess the error would be triggered also by other libraries that use GL. However, this seems like a different issue.

As solution to the error in this merge request, I like to suggest some workarounds → with my opinion:

1. I could try to import PyTorch before importing MuJoCo. → Seems quite dirty and hacky. Could break again.
2. We could switch for testing to a CPU build of PyTorch using `pip install torch --extra-index-url https://download.pytorch.org/whl/cpu`. This would speed up the tests and the CUDA Toolkit is not used in the docker container anyway. → If it really works, it is a clean solution, but should be done in another merge request with above script as new unit test, maybe.
3. Change the skipping code to `try...except` the `render()` call in the test and reraise if the error message is not `"gladLoadGL error"`. → Not sure how stable the message is. This could yield some issues if the message changes in the future.

Let me know, what you think.

pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts:
Amazing, thanks Chris for solving this

PR #495, opened 2023-05-08
Disallow Degenerate box space

### LINKED ISSUE #493: [Bug Report] Box contains treats float(inf) unexpectedly
### Describe the bug

When checking if a number is contained in a ``Box`` I get ``True`` but I expect it to be ``False``.

### Code example

```shell
Python 3.10.11 (main, May  4 2023, 15:26:51) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from gymnasium.spaces import Box
>>> Box(low=float("inf"), high=float("inf")).contains([-10])
True
>>> Box(low=-float("-inf"), high=float("inf")).contains([-10])
True
```

### System info

>>> gymnasium.__version__
'0.26.3'

### Additional context

_No response_

### Checklist

jjshoots:
# Description

This disallows the user from creating box spaces where the range is 0.

Fixes #493

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/spaces/box.py
+++ b/gymnasium/spaces/box.py
@@ -102,6 +102,11 @@ class Box(Space[NDArray[Any]]):
                 f"Box shape is inferred from low and high, expect their types to be np.ndarray, an integer or a float, actual type low: {type(low)}, high: {type(high)}"
             )
 
+        # check that we don't have a degenerate space
+        assert np.array(
+            low != high
+        ).all(), f"Some elements in low: {low} are equal to some elements in high: {high}, this will lead to a degenerate space and is not allowed"
+
         # Capture the boundedness information before replacing np.inf with get_inf
         _low = np.full(shape, low, dtype=float) if is_float_integer(low) else low
         self.bounded_below: NDArray[np.bool_] = -np.inf < _low
--- a/tests/spaces/test_box.py
+++ b/tests/spaces/test_box.py
@@ -61,7 +61,7 @@ def test_low_high_values(value, valid: bool):
     """Test what `low` and `high` values are valid for `Box` space."""
     if valid:
         with warnings.catch_warnings(record=True) as caught_warnings:
-            Box(low=value, high=value)
+            Box(low=-np.inf, high=value)
         assert len(caught_warnings) == 0, tuple(
             warning.message for warning in caught_warnings
         )
@@ -72,7 +72,7 @@ def test_low_high_values(value, valid: bool):
                 "expect their types to be np.ndarray, an integer or a float"
             ),
         ):
-            Box(low=value, high=value)
+            Box(low=-np.inf, high=value)
 
 
 @pytest.mark.parametrize(
@@ -314,3 +314,23 @@ def test_sample_mask():
         match=re.escape("Box.sample cannot be provided a mask, actual value: "),
     ):
         space.sample(mask=np.array([0, 1, 0], dtype=np.int8))
+
+
+@pytest.mark.parametrize(
+    "low, high",
+    [
+        (np.inf, np.inf),
+        (-np.inf, -np.inf),
+        (3.0, 3.0),
+        (np.array([-np.inf, 0]), np.array([-np.inf, np.inf])),
+    ],
+)
+def test_disallow_degenerate_spaces(low, high):
+    """Tests that we don't allow spaces with degenerate bounds, such as `Box(-np.inf, -np.inf)`."""
+    with pytest.raises(
+        AssertionError,
+        match=re.escape(
+            f"Some elements in low: {low} are equal to some elements in high: {high}, this will lead to a degenerate space and is not allowed"
+        ),
+    ):
+        Box(low, high, dtype=np.float32)


pseudo-rnd-thoughts:
As `np.random.uniform(low=0, high=0)` is supported, I think we should support this case despite how weird it is to only be able to sample for a single value. 

Therefore, I propose you just prevent three cases, `np.any(low > high)`, `np.any(low == np.inf)` and `np.any(high == -np.inf)`.
I think these are reasonable changes and should address the issue in https://github.com/Farama-Foundation/Gymnasium/issues/493

gymnasium/spaces/box.py
```diff
@@ -335,14 +328,26 @@ def _broadcast(
 ) -> NDArray[Any]:
     """Handle infinite bounds and broadcast at the same time if needed."""
     if is_float_integer(value):
-        value = get_inf(dtype, np.sign(value)) if np.isinf(value) else value
+        if np.isinf(value) and np.dtype(dtype).kind == "i":
+            value = (
```
pseudo-rnd-thoughts, line 332: `value = get_integer_infinite(dtype, np.sign(value))`
You have already checked that `np.isinf` above

Edit: If we use similar code to below, do we need the `get_integer_infinite` function?

gymnasium/spaces/box.py
```diff
@@ -335,14 +328,26 @@ def _broadcast(
...
+
+    elif isinstance(value, np.ndarray):
+        if np.any(np.isinf(value)) and np.dtype(dtype).kind == "i":
             # create new array with dtype, but maintain old one to preserve np.inf
-            temp = value.astype(dtype)
-            temp[np.isinf(value) & (value > 0.0)] = get_inf(dtype, 1.0)
-            temp[np.isinf(value) & (value < 0.0)] = get_inf(dtype, -1.0)
-            value = temp
+            bounded_value = value.astype(dtype)
+            bounded_value[np.isposinf(value)] = get_integer_infinite(dtype, 1.0)
```
pseudo-rnd-thoughts, line 344: Can we not just set `np.iinfo(dtype).max - 2` and same with line below, this is more clear that is happening

pseudo-rnd-thoughts (changes requested):

tests/spaces/test_box.py
```diff
@@ -314,3 +316,50 @@ def test_sample_mask():
...
+    [
+        (np.inf, np.inf, "positive_inf_below"),
+        (np.array([0, np.inf]), np.array([np.inf, np.inf]), "positive_inf_below"),
+        (-np.inf, -np.inf, "negative_inf_above"),
+        (np.array([-np.inf, -np.inf]), np.array([0, -np.inf]), "negative_inf_above"),
+        (5.0, 3.0, "reverse_bounded"),
+        (np.array([5.0, 6.0]), np.array([1.0, 5.99]), "reverse_bounded"),
+    ],
+)
+def test_disallow_degenerate_spaces(low, high, reason):
```
pseudo-rnd-thoughts, line 332: Could this test be updated to the suggested style on discord

tests/utils/test_passive_env_checker.py
```diff
@@ -37,11 +37,6 @@ def _modify_space(space: spaces.Space, attribute: str, value):
             spaces.Box(np.zeros(5), np.zeros(5)),
             "A Box observation space maximum and minimum values are equal. Actual equal coordinates: [(0,), (1,), (2,), (3,), (4,)]",
         ],
-        [
```
pseudo-rnd-thoughts, line 40: Why is this removed?

gymnasium/spaces/box.py
jjshoots: We haven't?

gymnasium/spaces/box.py
jjshoots: Good point.

tests/utils/test_passive_env_checker.py
jjshoots: We already don't allow reverse spaces by asserting this in the space. This warning will never appear.

gymnasium/spaces/box.py
jjshoots: Gotcha

gymnasium/spaces/box.py
```diff
@@ -121,6 +121,18 @@ def __init__(
             high.shape == shape
         ), f"high.shape doesn't match provided shape, high.shape: {high.shape}, shape: {shape}"
 
+        # check that we don't have invalid low or high
+        if not np.any(np.isnan(low) | np.isnan(high)):
+            assert np.all(
+                low <= high
+            ), f"Some low values are greater than high, low={low}, high={high}"
+        assert not np.any(
```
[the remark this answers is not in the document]
jjshoots: This is checking for reversed bounds. Low cannot be +inf and high cannot be -inf.

pseudo-rnd-thoughts:
@jjshoots Are you happy with the changes that I made? If so, I'm happy to merge

pseudo-rnd-thoughts (approved):

jjshoots:
@pseudo-rnd-thoughts Yep, SGTM.

PR #513, opened 2023-05-22
Update continuous_mountain_car.py

TachikakaMin:
change action[0] to action

# Description

There is no need to use action[0], which may cause bugs when using other repo.

1.  Other repo like tianshou will make action shape (1,) as one single number.
2.  Other env like pendulum also has action shape (1,) but doesn't have this code. 

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gymnasium/envs/classic_control/continuous_mountain_car.py
+++ b/gymnasium/envs/classic_control/continuous_mountain_car.py
@@ -146,6 +146,8 @@ class Continuous_MountainCarEnv(gym.Env):
     def step(self, action):
         position = self.state[0]
         velocity = self.state[1]
+        if len(action.shape) == 1:
+            action = action[0]
         force = min(max(action, self.min_action), self.max_action)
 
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
My reaction is that we shouldn't support/allow actions outside the action space
Can Tianshou update this on their side? It seems like more of an issue with their implementation than Gymnasium?

I see that this works for Pendulum but this seems like an accident than intended code
@RedTachyon Do you have any thoughts?

RedTachyon:
What exactly is the problem that this solves? As far as I can tell the only difference is that the type hint is removed for no apparent reason, and you can now pass a 0-dimensional array as an action, which isn't a valid action in the env's action space. And if the action is >2D for some reason, then weird stuff can happen.

TachikakaMin:
@pseudo-rnd-thoughts @RedTachyon

I think you are correct. I should fix it in tianshou rather than gym.

PR #518, opened 2023-05-24
Update record_video.py

voidflight:
# Description

Changed self.video_recorder.frames to self.video_recorder.recorded_frames in the reset() method; video_recorder.VideoRecorder() actually doesn't have a frames attribute, only a recorded_frames attribute. This should fix a bug with video recording where the video shows a frame from the previous rollout.

## Type of change

## Screenshots
Before:
image

After:
image

VideoRecorder class from gymnasium.wrappers.monitoring.video_recorder.py:
image


pseudo-rnd-thoughts:
@younik Is this right? Do we need to update `RecordVideoV0` as well?

pseudo-rnd-thoughts (approved):

mhbooll6:
handle

PR #534, opened 2023-06-01
Allow closing a closed env

### LINKED ISSUE #531: [Proposal] Explicitly requires to allow closing a closed environment
### Proposal

- Add a warning in the env checker when the custom environment that doesn't allow to be closed multiple times.
- Mention that it is recommended to be able to call the close function on an already closed environment.

### Motivation

It's usually possible to reset an environment several times.

```python
env = gym.make("CartPole-v1")
env.close()
env.close()
```

Is it good practice to do so? Since there's no simple way of checking whether an environment is closed, I propose to consider that calling the `close` method on a closed environment should not raise an error.

### Pitch

Give your opinion, I'll handle the PR if you're ok.

### Alternatives

Add an `is_closed` method, and always raise an error when the user tries to close a closed environment.

(But it would break so many things...)

### Additional context

See https://github.com/DLR-RM/stable-baselines3/pull/1525, and especially https://github.com/DLR-RM/stable-baselines3/pull/1525#issuecomment-1570998382

### Checklist

qgallouedec:
# Description

Fixes #531

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/core.py
+++ b/gymnasium/core.py
@@ -185,6 +185,7 @@ class Env(Generic[ObsType, ActType]):
         """After the user has finished using the environment, close contains the code necessary to "clean up" the environment.
 
         This is critical for closing rendering windows, database or HTTP connections.
+        Calling ``close`` on an already closed environment has no effect and won't raise an error.
         """
         pass
 
--- a/gymnasium/envs/mujoco/mujoco_rendering.py
+++ b/gymnasium/envs/mujoco/mujoco_rendering.py
@@ -603,7 +603,7 @@ class MujocoRenderer:
 
     The class has two main public methods available:
     - :meth:`render` - Renders the environment in three possible modes: "human", "rgb_array", or "depth_array"
-    - :meth:`close` - Closes all contexts initialized with the rendere
+    - :meth:`close` - Closes all contexts initialized with the renderer
 
     """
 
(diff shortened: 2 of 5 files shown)


pseudo-rnd-thoughts:
@qgallouedec Could we add something to the passive environment checker on `close` if the environment has been closed already just to warn the user?

qgallouedec:
Something like c1a35b4?

pseudo-rnd-thoughts:
> Something like [c1a35b4](https://github.com/Farama-Foundation/Gymnasium/commit/c1a35b4040db3fbe6eb1a9091e7c21a1e3c17ee0)?

Yes, that is great

pseudo-rnd-thoughts:
I have fixed this locally but it is more complex than I thought

We need to update `GenericTestEnv` entry point to the new value
```python
spec: EnvSpec = EnvSpec(
            "TestingEnv-v0", "tests.testing_env:GenericTestEnv", max_episode_steps=100
        ),
```

I realised that the recreated `GenericTestEnv` would be using the default `observation_space` and `action_space` rather than the updated spaces. The solution to this is more complex as we can't update `spec.update({"observation_space": observation_space, "action_space": action_space})` as `spec` is a class attribute not an object /instance attribute. 
This means that the final testing environment works but the other environments don't. 
Don't worry if what I said doesn't make sense, I think we ignore it but should be aware of it as a problem

pseudo-rnd-thoughts:
CI fixed in https://github.com/Farama-Foundation/Gymnasium/pull/564

Issue #357, opened 2023-03-04
[Question] AsyncVectorEnv "set_attr" not working as expected


FabioRaff:
### Question

I am trying to change the "goal" attribute of "FetchPickAndPlace-v2" (from gymnasium-robotics) using AsyncVectorEnv but it is not working as expected.

Code example:

```python
envs = gym.vector.make('FetchPickAndPlace-v2', num_envs=2)
actions = [[0, 0, 0, 0], [0, 0, 0, 0]]
new_goals = [[1, 1, 1], [1, 1, 1]]

obs, _ = envs.reset()
print(obs['desired_goal']) 
>>> [[1.28639574 0.69880261 0.42469975], [1.37864519 0.71064809 0.42469975]]

envs.set_attr('goal', new_goals)
```

if I check:

```python
print(envs.get_attr('goal'))
>>> ([1, 1, 1], [1, 1, 1])
```

it correctly prints the newly assigned goals, but after a step, the "desired_goals" key of the obs (which is simply self.goal attribute from the env) does not show the updated goals and rewards are also not calculated for the new goals:

```python
obs, rew, _, _, _ = envs.step(actions)
print(obs['desired_goal'])
>>> [[1.28639574 0.69880261 0.42469975], [1.37864519 0.71064809 0.42469975]]
```

Am I doing something wrong? How can I change instance attributes of vector environments?

_______
OS: Ubuntu 20.04
Python 3.7
gymnasium 0.27.1
gymnasium_robotics 1.2.0

pseudo-rnd-thoughts:
Ok, this is a really interesting problem

I can replicate this issue in the single agent case which is easier to understand
```python
env = gym.make("FetchPickAndPlace-v2")
print(env.unwrapped)
action = np.array([0, 0, 0, 0])
new_goal = np.array([1, 1, 1])

obs, _ = env.reset()
print(f'{obs["desired_goal"]=}')

print(f'pre-set {env.goal=}')
env.goal = new_goal
print(f'post-set {env.goal=}')

obs, _, _, _, _ = env.step(action)
print(f'{obs["desired_goal"]=}')
```

The problem is that an environment created from `make` is actually wrapped by several wrappers
```python
print(env)
>>> <TimeLimit<OrderEnforcing<PassiveEnvChecker<MujocoFetchPickAndPlaceEnv<FetchPickAndPlace-v2>>>>>
```

[...]

PR #535, opened 2023-06-01
Add deprecation warnings for features removed in Gymnasium v1.0

pseudo-rnd-thoughts:
# Description

In preparation for Gymnasium v1.0, we are starting to deprecate several legacy features that the dev team views cause more issues than gains. These features will raise warnings if users use them with the features eventually being removed in v1.0. 
I've shared below the rationale for removing each feature along with the proposed replacement code.

1. Remove `Wrapper.__get_attr__`, previously users could access variables not directly in the object (wrapper or env). For example, 
```python
import gymnasium as gym

env = gym.make("CartPole-v1")
print(env)
# <TimeLimit<OrderEnforcing<PassiveEnvChecker<CartPoleEnv<CartPole-v1>>>>>
print(type(env))
# TimeLimit
print(env.gravity)
# 9.8
```
While, the top wrapper (`TimeLimit`) does not include `gravity` as an attribute, `__get_attr__` would check if the variable exists in the wrapper's `env` attribute. However, this can cause unexpected issues
```python 
env.gravity = 10.0
print(env.gravity)
# 10
print(env.env.gravity)
# 9.8
print(env.unwrapped.gravity)
# 9.8
```
In this example, `env` claims that the variable `gravity` is equivalent to `10.0` however when you check what the base environment's `gravity` (`env.unwrapped`) is, it is still `9.8`, unchanged.
 
The reason for this is that the equivalent `__get_attr__` function for setting values was not implemented `__set_attr__`. As a result, the code above means that when we try setting a variable, the variable is only assigned on the top wrapper rather than the base environment that actually includes the variable. For a practical example see https://github.com/Farama-Foundation/Gymnasium/issues/357

We propose `Wrapper.get_wrapper_attr` as a replacement function that has the same functionality but is just explicit that the users can receive the variable from **anywhere** in the wrapper stack, rather than in a hidden way as is currently. For example
```python
env = gym.make("CartPole-v1")
print(env.gravity)
# raises AttributeError
print(env.get_wrapper_attr("gravity"))
# 9.8
```

An alternative solution is to implement the `__set_attr__` (https://github.com/Farama-Foundation/Gymnasium/pull/358) however we found this has a significant performance impact therefore, have not chosen this path.

2. Registration entry point - This was an undocumented feature provided initially for compatibility with Atari that allowed an external project to hook into Gymnasium such that on `import gymnasium` the module can register all of its environments. This is a relatively rare feature but it is used in a number of popular libraries, [atari](https://github.com/mgbellemare/Arcade-Learning-Environment), [minigrid](https://github.com/farama-Foundation/minigrid), and more. 
You can see this if you install minigrid, `pip install minigrid` then run `import gymnasium`, `gymnasium.pprint_registry()`, you should see there are a number of minigrid environment registered despite minigrid never being **explicitly** loaded.

We propose that the new code would require users to change to 
```python
import gymnasium as gym 

# either
import minigrid
gym.register_envs(minigrid)  # this is an optional step with `register_envs` being a no-operation function but allows IDEs to believe that the module was imported for something. 
env = gym.make("Minigrid-Empty-v0")
```
or 
```python
import gymnasium as gym
env = gym.make("minigrid:Minigrid-Empty-v0")  # here "minigrid:" tells `gym.make` to load the module `minigrid` first allowing it's environments to be registered before the environment is created
```

3. Remove `Env.reward_range`, this is a legacy attribute, that is very rarely utilised, representing the minimum and maximum reward of an environment. This attribute is not implemented within any Farama projects or any popular tutorials to our knowledge. As this is an attribute, there is little that we can do to notify users of the deprecation but due to its rarely we doubt this will cause significant issues

4. Removing `autoreset` parameter from `gymnasium.make` and `gymnasium.register` function. The `autoreset` argument was used to specify if the `AutoResetWrapper` should be added to the environment, however, due to its uncommon use, the addition of the `additional_wrappers` parameter and the wish to simplify the `make` and `register` functions, we are deprecating the parameter. 

5. Removing `gymnasium.vector.make` in favour of `gymnasium.make_vec`, as in v1.0 we will be introducing a new vector environment implementation. The implementation is very similar to the previous implementation (more details will be provided soon) but as we have modified several of the parameters in `vector.make` for `make_vec`, we have change the function name to reflect the change. Furthermore, we hope that `make_vec` will make more users aware and use the function.

If you believe that we have missed anything significant with these changes, please comment below or ask a question on the farama discord and I will try to respond.


gymnasium/envs/registration.py
```diff
@@ -600,10 +600,16 @@ def load_plugin_envs(entry_point: str = "gymnasium.envs"):
             fn = plugin.load()
             try:
                 fn()
+                logger.warn(f"The environment registration system on `import gymnasium` is deprecated and will be removed in v1.0 ({plugin.value}), please update code to `import {plugin.value}` separately or use to load the module `gym.make('{plugin.value}:env_name-v0')`")
```
RedTachyon, line 603: I'm pretty sure this will print out a bunch of warnings whenever you import gymnasium, if you happen to have shimmy/minigrid/etc installed, which is not a great thing. See out previous discussion about how to deprecate it, it might not be possible to do in a sane way.

gymnasium/core.py
```diff
@@ -272,15 +275,35 @@ def __init__(self, env: Env[ObsType, ActType]):
...
+        Warnings:
+            This feature is deprecated and removed in v1.0 and replaced with `env.get_attr(name})`
+        """
         if name == "_np_random":
             raise AttributeError(
                 "Can't access `_np_random` of a wrapper, use `self.unwrapped._np_random` or `self.np_random`."
             )
         elif name.startswith("_"):
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
+        logger.warn(f"This feature is deprecated and will be removed in v1.0 and replaced with `env.get_attr('{name}')`.")
```
RedTachyon, line 295: We should probably promote just doing `env.name` or `env.unwrapped.name`, since that covers like 99.9% of use-cases

gymnasium/envs/registration.py
```diff
@@ -600,10 +600,16 @@ def load_plugin_envs(entry_point: str = "gymnasium.envs"):
...
             except Exception:
                 logger.warn(f"plugin: {plugin.value} raised {traceback.format_exc()}")
 
 
+def register_envs(env_module):
```
RedTachyon, line 608: `env_module: types.ModuleType`?

pseudo-rnd-thoughts: I didn't know a `ModuleType` existed, I will update

gymnasium/envs/registration.py
pseudo-rnd-thoughts: I can't remember what we discussed as the alternative solution. I can't think of another way of deprecating this with just putting a warning here for now

RedTachyon:
FWIW I still think we absolutely can't have the plugin system warning show up automatically whenever you import gymnasium, it will actively make people just not update their code. 

There's pretty much no way to remove the warning beyond explicitly muting it before importing gymnasium -- even if you try to do e.g. `import shimmy` before that, internally, shimmy will import `gymnasium` before those environments are initialized, and when you import gymnasium, the plugin system kicks in anyways and prints the warning.

gymnasium/core.py
```diff
@@ -273,15 +277,48 @@ def __init__(self, env: Env[ObsType, ActType]):
...
             )
         elif name.startswith("_"):
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
+        logger.warn(
+            f"env.{name} to get variables from other wrappers is deprecated and will be removed in v1.0, "
+            f"to get this variable you can do `env.unwrapped.{name}` for environment variables or `env.get_attr('{name}')` that will search the reminding wrappers."
+        )
         return getattr(self.env, name)
 
+    def get_attr(self, name: str) -> Any:
```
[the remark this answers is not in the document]
elliottower: This seemed confusing to me as well when reading through the changes, but can't think of a better name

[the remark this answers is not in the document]
pseudo-rnd-thoughts: >  It's very similar to __getattr__ and getattr, and it should be obvious that it's just a custom method for convenience, not something fundamental

I agree but that is kinda the point, that the function is not equivalent to how the `__get_attr__` worked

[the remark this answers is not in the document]
RedTachyon: Kinda improvising here - what if we give it another optional argument where you specify the wrapper class or its name, and if it's left empty, then it gets the attribute from the first wrapper where it finds it? Then call it something like `get_wrapper_attr`

pseudo-rnd-thoughts:
> FWIW I still think we absolutely can't have the plugin system warning show up automatically whenever you import gymnasium, it will actively make people just not update their code.

> There's pretty much no way to remove the warning beyond explicitly muting it before importing gymnasium -- even if you try to do e.g. import shimmy before that, internally, shimmy will import gymnasium before those environments are initialized, and when you import gymnasium, the plugin system kicks in anyways and prints the warning.

As I believe that the plugin system is used only within farama projects, we can probably deprecate it without a warning, the problem is that users need to know that they need to load modules before they make the environment, unlike before. We could probably do some magic to note environments registered through the plugin system but I'm not sure it is worth it

PR #548, opened 2023-06-08
Remove jumpy due to its partial implementation

pseudo-rnd-thoughts:
Jumpy was added due to the belief that Jax-based environment were going to replace a number of current environment implementations.
However, due to several factors, this has not happened. 
Therefore, this PR removes Jumpy from the project in order that when jax-based environment are properly added, then Jumpy can be properly added and tested

### DIFF
--- a/gymnasium/experimental/vector/vector_env.py
+++ b/gymnasium/experimental/vector/vector_env.py
@@ -299,11 +299,11 @@ class VectorWrapper(VectorEnv):
         """Step all environments."""
         return self.env.step(actions)
 
-    def close(self, **kwargs):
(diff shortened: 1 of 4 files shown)


RedTachyon:
Can you elaborate on what exactly are the issues with using jumpy right now? I'd mainly prefer to avoid flip-flopping between having jumpy and no jumpy (in the scenario we get all the jax/jumpy stuff working later and want to get back to it)

pseudo-rnd-thoughts:
If Jax is installed then it will be imported which will slow gymnasium import. 
Additionally, we haven't got any testing that `jumpy` is implemented correctly

PR #557, opened 2023-06-15
Adding starting values to MultiDiscrete spaces

Rayerdyne:
Feature addition: `MultiDiscrete` spaces with a starting value, much like `Discrete` spaces. 

No additional dependencies required for this change.

## Best practice question:
To me it makes more sense to have the `start` argument of the `__init__()` function in second place, however if I did so it would break the code providing `dtype` by position. That's why I put it in the last position, should I change?

## New feature

# Checklist:

- [ ] I have made corresponding changes to the documentation (to the docstrings, is that it ?)
- [x] My changes generate no new warnings (AFAIK no, I'll see that with testing)
- [x] I have added tests that prove (my fix is effective or) that my feature works

### DIFF
--- a/gymnasium/experimental/vector/utils/space_utils.py
+++ b/gymnasium/experimental/vector/utils/space_utils.py
@@ -73,28 +73,21 @@ def _batch_space_box(space: Box, n: int = 1):
 
 @batch_space.register(Discrete)
 def _batch_space_discrete(space: Discrete, n: int = 1):
-    if space.start == 0:
-        return MultiDiscrete(
-            np.full((n,), space.n, dtype=space.dtype),
-            dtype=space.dtype,
-            seed=deepcopy(space.np_random),
-        )
-    else:
-        return Box(
-            low=space.start,
-            high=space.start + space.n - 1,
-            shape=(n,),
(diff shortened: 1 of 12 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for the implementation, could you make the following changes

* `start` should default to `np.zeros()` if None
* Like `self.nvec`, I think line 72 should be moved to 62 then have the testing afterwards 
* remove the `if self.start is None` as the maths should work with `start` as a matrix
* could you add a special test for `start`
* could you add a `MultiDiscrete` with `start` to `tests/spaces/utils.py`

pseudo-rnd-thoughts (changes requested):
Amazing, thanks for the changes, I realised that the last thing is that the `_apply_mask` function needs to modified to include the `start`

pseudo-rnd-thoughts:
It looks like the CI is failing because you need to update the `gymnasium/vector/utils` functions in `spaces.py`

pseudo-rnd-thoughts (approved):
Amazing, thanks for all your work. Looks good to me

PR #573, opened 2023-06-27
drop `python 3.7` support

Kallinteris-Andreas:
`python 3.7` becomes EOL today
so we should drop support for it

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -21,9 +21,6 @@ jobs:
       matrix:
         include:
         - os: ubuntu-latest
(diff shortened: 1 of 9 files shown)


pseudo-rnd-thoughts:
Thanks, I'm not going to merge this until v0.28.2 is out which is sadly being delayed a bit
Then for v0.29, I will merge this

pseudo-rnd-thoughts:
There is a change of plan, the next release will be v0.29 so could you update to main then I will merge

pseudo-rnd-thoughts (approved):
Thanks

Issue #575, opened 2023-06-28
[Proposal] add `utils.check_environments_match`


Kallinteris-Andreas:
### Proposal

Add:
```
def check_environments_match(
    env_a: gym.Env,
    env_b: gym.Env,
    num_steps: int,
    seed: int = None,
    skip_obs: bool = False,
    skip_rew: bool = False,
    skip_terminal: bool = False,
    skip_truncated: bool = False,
    skip_info: bool = False,
):
    """Checks if the environments `env_a` & `env_b` are identical.

    Args:
        num_steps: number of timesteps to test for, setting to 0 tests only resetting.
        skip_obs: If `True` it does not check for equivalence of the observation.
        skip_rew: If `True` it does not check for equivalence of the observation.
        skip_terminal: If `True` it does not check for equivalence of the observation.
        skip_truncated: If `True` it does not check for equivalence of the observation.
        skip_info: If `True` it does not check for equivalence of the observation.
    """
    assert env_a.action_space == env_b.action_space
    assert skip_obs or env_b.observation_space == env_b.observation_space

    obs_a, info_a = env_a.reset(seed)
    obs_b, info_b = env_b.reset(seed)
    assert skip_obs or data_equivalence(obs_a, obs_b)
    assert skip_info or data_equivalence(info_a, info_b)
    env_a.action_space.seed(seed)
    for _ in range(num_steps):
        action = env_a.action_space.sample()
        obs_a, rew_a, terminal_a, truncated_a, info_a = env_a.step(action)
        obs_b, rew_b, terminal_b, truncated_b, info_b = env_b.step(action)
        assert skip_obs or data_equivalence(obs_a, obs_b)
        assert skip_rew or data_equivalence(rew_a, rew_b)
        assert skip_terminal or terminal_a == terminal_b
        assert skip_truncated or truncated_a == truncated_b
        assert skip_info or data_equivalence(info_a, info_b)

        if terminal_a or truncated_a:
            obs_a, info_a = env_a.reset(seed)
            obs_b, info_b = env_b.reset(seed)
            assert skip_obs or data_equivalence(obs_a, obs_b)
            assert skip_info or data_equivalence(info_a, info_b)
```

Note: pettingzoo has 

https://github.com/Farama-Foundation/PettingZoo/blob/969c079164e6fd84b4878f26debfeff644aa2786/pettingzoo/test/seed_test.py#L16

### Motivation

It could enable:
```
[...]

PR #576, opened 2023-06-29
add `check_environments_match` function for checking if two environments are identical

Kallinteris-Andreas:
# Description

implements https://github.com/Farama-Foundation/Gymnasium/issues/575


gymnasium/utils/env_tests.py
pseudo-rnd-thoughts, line 1: Rename to `env_match`

pseudo-rnd-thoughts (changes requested):
Could you update the mujoco version test to use this function?

Kallinteris-Andreas:
`test_mujoco_v3` does not check for equality it checks for close "enought", which this function does not support

gymnasium/utils/env_tests.py
pseudo-rnd-thoughts: Sorry, I meant the filename rather the function name, I think `check_environments_match` is a good name

pseudo-rnd-thoughts (approved):
My only worry is that we don't have any testing but I don't imagine this being a popular function

PR #585, opened 2023-07-05
renable some disabled `pyright` checks

Kallinteris-Andreas:
Title says it all, `pyright` does not seem to create issues with those checks enabled


Kallinteris-Andreas:
Never mind it does generate errors

PR #600, opened 2023-07-10
Fix some MuJoCo env `TypeIssues`

Kallinteris-Andreas:


pseudo-rnd-thoughts (commented):
Does this not change the functionality with the `_initialize_simulation` not setting `self.model` anymore ?

Kallinteris-Andreas:
It is now set with `self.model, self.data = self._initialize_simulation()` (because with the way it was implemented before `pyright` could not figure out that `bastMujocoEnv` had `model` & `data` member variable)

pseudo-rnd-thoughts (approved):

PR #604, opened 2023-07-13
Removing Action Space comparison with CTRL for MuJoco Envs

reginald-mclean:
# Description

For Meta-World to use the Mujoco-Env in Gymnasium, the comparison between the Environment's action space and the ctrl of the arm had to be removed. Ignore all the extra commits, we are currently using my fork of Gymnasium in the Meta-World benchmarking task


pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts:
There seems to be a test that explicitly checks the actions are of the correct size. Is there anyway we can keep the test and the proposed change? @Kallinteris-Andreas any thoughts? We can just remove the tests if necessary

Kallinteris-Andreas:
Removing the tests from `MujocoEnv` is fine, `mujoco` has its own checks

The only downside it that `mujoco` has a worst error message (but that only affects the development of new environment, not the users)

Getting metaworld to use `mujoco` bindings would be great

Kallinteris-Andreas (approved):

reginald-mclean:
@Kallinteris-Andreas one of the tests specifically manipulates action sizes before applying them to an environment. The test expects an error to be raised but with the removal of the check, the error doesn't get raised. This function: https://github.com/Farama-Foundation/Gymnasium/blob/aa6d0c8787f63664fbdd027cdb8a9febe997c96f/tests/envs/test_action_dim_check.py#L15C15-L15C15

pseudo-rnd-thoughts:
To confirm, the issue is that within metaworld, we have mujoco environments where there are more or less actions than the number of arm controls. How is that possible? 
One solution would be to move the test to the actual environment rather than the base mujoco environment

reginald-mclean:
Mujoco let's you send forces to the actuators in a model between a specified range. In certain environments (ie FrankaKitchen) each joint is considered an actuator so when you apply an action what you are doing is specifying what position you want the joints to be set to. In Meta-World, the only actuator that actions control is the gripper. Thus there are only two ctrl forces that can be sent which is not equal to the Meta-World action space size of 4. Note: In Meta-World the last element of the action is used by both actuators just in opposite directions. Positive for one actuator closes that finger joint while negative for the other joint closes. Same thing for opening the gripper.

reginald-mclean:
Meta-World should (I will do this now) check that the action is the expected size of 4.

pseudo-rnd-thoughts:
Confusingly, the base mujoco environment isn't passed the action space but uses it in this test. We could add a parameter in the constructor for a custom number of controls

reginald-mclean:
Meta-World initializes the MujocoEnv, which calls ```self._set_action_space()``` in the base class that sets self.action_space via inference of the ctrl attributes of the actuators. Then Meta-World overrides that with it's own action space.

Kallinteris-Andreas:
```
if np.array(ctrl).shape != self.action_space.shape:
```
could be replaced with 
```
if np.array(ctrl).shape != (self.mode.nu, ):
```

Also, `MujocoEnv` (based on `mujoco` bindings) does seem to have this issue

pseudo-rnd-thoughts:
@Kallinteris-Andreas My propose is 
```python
    def _set_action_space(self):
        bounds = self.model.actuator_ctrlrange.copy().astype(np.float32)
        low, high = bounds.T
        self.action_space = spaces.Box(low=low, high=high, dtype=np.float32)
        self.num_actuator_ctrl = bounds.shape
        return self.action_space
```
such that we have a new parameter, `self.num_actuator_ctrl` then modify the error to test `ctrl.shape == self.num_actuator_ctrl` rather than action space

Kallinteris-Andreas:
We already have a "`self.num_actuator_ctrl`" in `self.model.nu`

Kallinteris-Andreas:
@reginald-mclean can confirm that your issue has been solved

reginald-mclean:
@Kallinteris-Andreas @pseudo-rnd-thoughts it works! thank you

PR #616, opened 2023-07-17
Limit cython version on mujoco-py option install

pseudo-rnd-thoughts:
# Description

Our CI randomly started to fail, python was released today so we suspect that cython 3 with the introduction on nogail caused the issue


Kallinteris-Andreas:
you also have to change "all"

PR #622, opened 2023-07-19
[WIP] Updates to functional jax vector envs

RedTachyon:
Putting this up so that I don't forget, and potentially someone looks at it in the meantime. I wrote this code a couple weeks ago, but forgot to finish it/make it into a proper PR.

We will have to double check this after we have a decision on the semantics for new vector env info dicts, but this should roughly reflect the general vibe of "record the full terminal observation/terminal info when there's a terminal step"


pseudo-rnd-thoughts:
Completed in https://github.com/Farama-Foundation/Gymnasium/pull/694

PR #626, opened 2023-07-24
Default `gym.make_env` to `custom` when `vector_entry_point` is specified

### LINKED ISSUE #625: [Bug Report] `vector_entry_point` not working as expected
### Describe the bug

I was expecting that when defining the `vector_entry_point` in the gym.register function that it would defer to my custom vectorized create function instead of looping over the environment create method and wrapping with sync/async. for my env, i wrote custom vectorization code, but its totally useless when wrapped with sync/async vec wrapper.

### Code example

```shell
Inside of "path/to/vec_func.py"

import gymnasium as gym
class CustomVecEnv(gym.vector.VectorEnv):
    pass

def vec_func(*args, **kwargs):
    return CustomVecEnv(*args, **kwargs)
```

```
gymnasium_register(
    id="MyEnv-v0",
    entry_point="path/to/func:func",
    vector_entry_point="path/to/vec_func:vec_func",
)
```

then

```
env = gym.vector.make("MyEnv-v0")
```

doesn't work, for me it just hangs, but if you get it not to hang it should return a sync/async instance of the env instead of returning the custom class env definition.
```


### System info

`gymnasium                     0.28.1`
`ubuntu 20.04`
`python = 3.8.12`

### Additional context

i'm using a custom vectorized environment. I can't share it, but it should be purely an interface issue. 

also, what is `vector_entry_point` even doing?

### Checklist

### PR DESCRIPTION
**Pull request by nollied**
# Description

When registering a custom environment with `vector_entry_point`, in order to use the custom vectorization you have to create it with `vectorization_mode="custom"` but this isn't ideal, so I set it to `vectorization_mode=None` by default and check if `vector_entry_point` is specified. if it is, default to `custom` otherwise use `async`.

I updated docstring to reflect.

Fixes #625 

## Type of change

### Screenshots

image

### CODE REVIEW

**Comment by pseudo-rnd-thoughts** [comment]
I am very uncertain of this, given that vectorisation mode has a default value this has a hidden change to most users. 
Furthermore, we cannot guarantee that the custom version is identical to a parallelised version of the single env version, I.e. sync and async

**Comment by RedTachyon** [comment]
While I'm definitely an enthusiast of smarter vectorization methods than sync/async, I agree with Mark here. Quoting the Zen of Python, "explicit is better than implicit". With this we would have a very implicit change of behavior between envs -- calling `gym.make_vec("SomeEnv")` with default arguments might be an AsyncVectorEnv, or it might be something completely different. 

It's better to just keep it consistent, so that calling it without arguments will always result in the same thing.

On that topic, I noticed it defaults to `async`, @pseudo-rnd-thoughts what do you think about changing it so that it defaults to `sync`? It's way way simpler, and counterintuitively can sometimes be more efficient than `async` due to the lack of `mp` overhead. Plus `sync` is more consistent, since `async` can have issues with start methods on different OS's

**Comment by pseudo-rnd-thoughts** [comment]
@RedTachyon I would agree with everything your saying.

For changing to sync, functionally there is no difference to users (I believe) and would allow users to op-in to a possible speed up and issues associated with async 

Todo - add / check that a test exist for async == sync vector env

**Comment by nollied** [comment]
if a user specifies a custom vectorization in the registration, then I think it's a fair assumption that user accepts the risks that their implementation might not be good. afterall, it's not going to influence any preexisting environments, nor ones with sync or async vectorization strategies. the only time this will impact anyone's work is if they:
1. have their own custom environment
2. AND they specify their own vectorization entry point in the registration function

I can't think of a world where they would do this, but then expect the vectorization function not to get used. that's why I designed it to be a situational default where it defaults to async (or sync or whatever) and when specified it prioritizes custom.

@pseudo-rnd-thoughts  @RedTachyon

**Comment by pseudo-rnd-thoughts** [comment]
The problem is that the first assumption is wrong, we as the dev team are interested in adding "custom" vector environments as for a number of environments.

This is as custom environments are faster than async in almost all cases. 

Therefore, I don't think can make this change

**Comment by nollied** [comment]
Well.. then can you add it to the registration spec what should be the default? That way the developers of the environment can decide what is recommended? I don't understand why you're pushing back so hard on a default parameter that would save users time.

**Comment by nollied** [comment]
Also the naming of "vector_entry_point" is confusing if you don't think we should default vector entry points to the custom entry point specified in the spec...

I, as a user, was expecting that by specifying a vector *entry point* that when calling make_vec it would use the class for my vectorized entry point.

**Comment by nollied** [comment]
@pseudo-rnd-thoughts 

Also, to register a function you don't even need to specify a "entry point"... you can optionally specify a "vector entry point" instead.

So... if you were to ONLY specify a vector entry point.... then wtf does your environment do if you make_vec it without "custom" as the vec type? 

It will do absolutely NOTHING!


https://github.com/Farama-Foundation/Gymnasium/blob/129697cf8eb1aa2374e1ac5c4c774631555fb363/gymnasium/envs/registration.py#L653

**Comment by RedTachyon** [comment]
A few points:
1. Chill lol
2. The main use-case of Gymnasium (and especially of the registry) isn't necessarily one developer creating an environment, registering it, and then using it themselves. It's someone creating an environment, publishing it, and then someone else (potentially many other people) using it.
3. There is actually a point in all of this -- since an env requires either an entry point, or a vector entry point (not necessarily both), then whatever the default value, `make_vec` might fail. Using an "adaptive" default value would indeed fix this, at the cost of clarity. So the choice is (a) no default value, users have to be explicit, (b) either `sync` or `custom` as a default value, clear error message if that happens to not work, or (c) `custom` if available, `sync` otherwise. I'm still slightly leaning towards keeping it more explicit in one way or another, but if we want to put a larger emphasis on vector envs with clever vectorization (which might be the right move), it might be worth considering, I'll need to think about it a bit more.

**Comment by nollied** [comment]
Sorry, I'm pretty passionate about the gymnasium stuff because it's a super fundamental piece of the RL literature and want to see it become more friendly to vectorized use-cases.

I appreciate you recognizing it's worth considering @RedTachyon, although for me it doesn't seem to sacrifice clarity at all.

Calling something an entry point usually means that, later when called, that code will enter that entry point. The default argument is something that would realistically only be changed from "custom" to something else when in development or experiencing some kind of bug with the custom vectorization.

Ideally, if someone submits a new environment for many others to use, and it has super optimized vectorization, no one would have a need to switch out of custom to async or sync. Again, unless they're developing or experiencing a bug.

**Comment by RedTachyon** [comment]
So where I'm at right now is that we either remove the default altogether, or go with this approach (or some variation of it). 

Basically I agree that we should promote vectorized envs more, since non-vectorized envs are realistically only for tutorials and toy problems. So at the very least, we should keep some symmetry between single and vectorized envs, and potentially make vectorized envs a bit more privileged. With status quo, single envs are privileged, because `sync`/`async` relies on the env having an `entry_point` as opposed to having a `vector_entry_point`. This PR shifts the privilege to vectorized envs, and adds a fall-back in case it really is a single-only env. Having this kind of "adaptive" default isn't entirely unheard of as far as I'm aware (can't mention any examples right now, but there's definitely some code like that, maybe in sklearn?), so while it might be a bit confusing, it's not the worst thing to have.

I'm now slightly leaning towards accepting this (in which case I'll do my best to convince Mark so we can get a consensus), but I need to let it stew for a bit more, and if so, I'll do an actual review pass of the code to make sure everything's correct.

**Comment by nollied** [comment]
@RedTachyon awesome! 

The reason for the conditional default is mostly due to the complexity of the vector entry point registration to begin with. When specifying the vector entry point, by default it is `None` -- This implies that async (or whatever default is defined for vectorization strategy) will be used when calling `make_vec`. Also you only need to specify the entry point or vector entry point (as mentioned before), so that adds even more complexity. 

Implicits in code is not necessarily a bad thing -- often it can be extremely useful. Take for example pytorch's pre-defined nn modules -- if you call the forward function with an unbatched input -- it will unsqueeze a batch dimension for you. That's an implicit that makes pytorch feel more fluid. It could instead raise an error and tell the user "hey, you should unqueeze this batch" and that would work -- but instead it's much more convenient just to unsqueeze it.

We have to pick the implicits that serve our users/developers the best. In this case, this change introduces something that's more predictable with literally just an extra if case (+2 lines of complexity) that is clarified in the docstring. It's not like this change adds +500 lines of code to implement this feature.

At most it will take 5 more seconds to understand what's going on, and it will save other users developing similar vectorizations as mine at least 20-30 minutes of time with trying to figure out why their code isn't being executed. Tradeoffs considered; this introduces negligible complexity and saves a lot of guess-work. I had to read the gymnasium source code to figure out why my vec entry point wasn't being executed. I had to get oriented in the code, and then figure out why. All in all it took me 20-30 minutes to figure out why.

Again, we're talking about an effectively 2 line code change with a clarifying docstring that's descriptive enough for users and also works as expected for vectorization developers.

Specific examples of conditional defaults are everywhere in many popular libraries. I tried to poke around and find a few but couldn't do so within 5 minutes. To be frank, it's not a wise spending of time to look into fore longer than that. Honestly, what should matter is how the code feels to developers and users; not whether or not it's written in someone's textbook of acceptable programming paradigms.

I'm obviously invested in this enough to spend 3+ hours contributing and defending an effectively +2 line change -- having been a user of gym for years (now gymasium) I feel obligated to speak up here.

Like I said, I'm developing legitimate industry vectorized environments (that will get used by many other people btw -- to reference previous messages in this thread) and got significant performance improvements using my own custom vectorization strategy. When plugging it into our RL training code, I noticed it wasn't being called for some peculiar reason. I spent 20-30 minutes figuring out why -- made a change to gymnasium -- and my code was working as expected (I found out I could just call make_env with "custom" vectorization strategy, but decided to contribute because i wanted to save other's in my shoes a bit of time).

So I mean -- yes i'm sure there are many other solutions, but it's an extremely tiny change and not worth investigating more than we already have.

**Comment by nollied** [comment]
Due to this -- i think it's is a must-have feature to maintain support for pre-existing RL libraries that use the `make_vec` function call internally. Otherwise all RL libraries will have to implement their own version of this conditional default argument anyways.

However we fix this -- either with +2 lines or a full blown refactor -- i think it's a necessary change.

**Comment by pseudo-rnd-thoughts** [comment]
Sorry, I've been ill the last couple of days therefore I haven't responded.

In thinking about this, I'm more in favour than I was previously but I still don't like the adaptive behaviour.
My idea would be `vectorization_mode` defaults to `None` which uses the `vector_entry_point`.   
Otherwise, an error would be raised saying that a vectorization mode needs to be selected, i.e., `async` and `sync`. 
Therefore, by default, `vector_entry_point` is used.  

Why, thinking about this in the future, we would like EnvPool, SampleFactory and any other vector-only libraries to be supported by Gymnasium and we don't want users to call `gym.make_vec("Envpool/Cartpole-v1", vectorization_mode="custom")` This is ugly at best. 
I understand that the adaptive code would work, I just don't like the silent change. I would prefer that this is explicit to the user on what is happening.  

If we take this approach, I think we need several things (some of these might be done)
1. update the docs for environments with vector versions, i.e., cartpole (there aren't many)
2. link in the error to a section on the website with a description of the vectorization modes with their advantages and disadvantages
3. I think we should make this change on v1.0.0 branch as this is a large change to v0.29.0
4. update a large amount of our code to explicitly use the `vectorization_mode="sync"`
5. thoroughly check that the existing custom vector environments within gymnasium act basically equivalent to the non-vector versions

**Comment by nollied** [comment]
@pseudo-rnd-thoughts okay, I think that's a great idea!

It does make a lot of sense, and it means we don't have to decide which fallback default is better for users "async" or "sync", but also it will use the vec entry point when available.

However, what if a user puts in "sync" or "async" out of habit and aren't aware an env has "custom"?

Can we further add a warning for this case to tell users that their environment has a custom mode and they are recommended to enable i?

**Comment by pseudo-rnd-thoughts** [comment]
> However, what if a user puts in "sync" or "async" out of habit and aren't aware an env has "custom"?

It is completely possible for environments to have both a vector and single env implementation, i.e., a entry point and vector entry point like CartPole does currently. In this case, the user has the choice and we can't really force the user to either. We could raise a warning to the user to alert them but it is their choice in the end

**Comment by nollied** [comment]
> It is completely possible for environments to have both a vector and single env implementation, i.e., a entry point and vector entry point like CartPole does currently. In this case, the user has the choice and we can't really force the user to either. We could raise a warning to the user to alert them but it is their choice in the end

Are you saying that if an environment has both an entry point and vector entry point available -- the make_vec call should error out? The way I interpreted [your previous message](https://github.com/Farama-Foundation/Gymnasium/pull/626#issuecomment-1666503159) was:
1. when environment spec has `entry_point` **and** `vec_entry_point` specified:
    - default None, resolve to `vec_entry_point`
2. when environment spec has `vec_entry_point` **only**
    - default None, resolve to `vec_entry_point`
3. when environment spec has `entry_point` **only**
    - default to None, raise an exception if None, tell the user they must select `sync` or `async`.

If this is truly your case, I would agree to compromise here (although, I still think it would be better to have a situational default from the perspective of user experience).

I understand the exception raise there in ambiguity, and to let the user learn about the interface while using it.

My suggestion for warning the user is not to force the user into some kind of decision around sync or async, but rather to give them a heads-up in case they are unaware of the custom vectorization stuff.

To clarify, here's what I mean exactly:
1. when environment spec has `vec_entry_point` specified **and**  the user has chosen `sync` or `async` as their `vectorization_mode`:
    - **warn** that user that using `sync` or `async` may be suboptimal, and the environment they're using actually has `vec_entry_point` defined which is likely to be much faster and more stable than their selection + a link to read more in the docs.

**Comment by RedTachyon** [comment]
100% no on raising a warning if someone explicitly chooses `sync` or `async`.

My idea for the order of choosing the right implementation is roughly:

```python
if mode is None:
    if getattr(env_spec, "vector_entry_point"):
        mode = "custom"
    else:
        mode = "sync"
```

nollied:
@RedTachyon I do think that's best, but I was proposing a potential compromise since @pseudo-rnd-thoughts is pretty adamant about wanting to raise an exception.

The warning would be only applicable if we went with a compromise if @pseudo-rnd-thoughts is unwilling to waiver in this.

By the way, in the code it says we're allowed to (in the registration function) specify a `entry_point` **OR** a `vector_entry_point`, but you actually can't. you can do just an entry_point OR both an entry point and vector_entry_point but never just a vector entry point.

Honestly, I feel like a redesign of the registration function is due alongside the `make_vec` call, especially if you want to postpone this change to some kind of v1.0.0 as [@pseudo-rnd-thoughts claims is best](https://github.com/Farama-Foundation/Gymnasium/pull/626#issuecomment-1666503159). There are quite a lot of concerns and weird bugs that are all around it and it simply fails in many cases to do what it's supposed to.

I feel like it's also kinda peculiar that we split `make` and `make_vec` into 2 separate functions, but `register` is only 1 lumping together both single and vector env.

If we really want to become super friendly to vector and non-vector envs under a mostly unified interface, it seems that having both be supported seamlessly together is super important. Sort of in the same paradigm how pytorch does their implicit batch dimension stuff with their pre-defined torch.nn.Modules and functional methods.

Tangent aside, @RedTachyon just to clarify I think you and I are in full agreeance, but @pseudo-rnd-thoughts is not, and my suggestion of the warning was building on his idea to resolve to a sort of compromise. I'd much rather prefer not to compromise, but I mean, we do need consensus... right?

nollied:
So, we agree on:
1. Let's change default of vector_mode to "custom" (or renamed version, but same function)
2. I don't have a strong preference for the vector mode name -- other than the suggestion that maybe we just use the class itself instead of a string. Like vector_mode=gym.sync or gym.async (aliases for the class) or vector_mode=MyVecEnvClass. If we did this, maybe just calling it "vector_entry_point_override" instead of vector_mode is the way to go.

Also, another source of complexity: sync and async vector strategies are wrappers, but vector entry point isn't a wrapper, it's a different class entirely which may or may not be a wrapper.

My point about PyTorch's implicit batch dimension was more about trying to have the vectorized environments be the default interface, since num_envs 1 is a special case of vectorization. This is what I and my coworkers were actually hoping for.

And, having an implicit batch dimension in RL frameworks isn't really a hard thing to account for and in my mind is totally acceptable.

Yeah backwards compatibility is never easy to manage -- but there are definitely creative ways to do so.

Introducing breaking changes can piss off a lot of users, including those developing RL frameworks. But I mean, the future of RL environments seems to have been trending in the direction of vectorization for years. It makes sense to make gymnasium more vec native. Plus, I saw that some frameworks don't even really use the gym vec env stuff at all (stablebaselines3) they kinda do but really they wrote their own gymnasium vector wrapper and totally disregard the truncated flag.

Oh and another thought -- why are infos a dictionary and not a list and why is it so hard to use? I understand it's supposed to be more tensorized -- a dictionary of tensors instead of list of dictionaries, but... I found this to be the hardest part of the vec interface to get working. At least we should add default base class functionality to handle this for users.

nollied:
Particularly with the final observation stuff. Man that got confusing

nollied:
@RedTachyon sync and async are wrappers, they take in a single env and vectorize it.

Custom vec envs do not take in the single version, they are standalone

pseudo-rnd-thoughts:
@nollied Are you happy to make the changes in this PR discussed in https://github.com/Farama-Foundation/Gymnasium/pull/626#issuecomment-1666503159

Updated list would be
1. Update default `vectorization_mode="vector_entry_point"`
2. Update all of the tests to use the intended `vectorization_mode`
3. Update the docs for environments with vector versions, i.e., cartpole (there aren't many) to note they exist and how to gym them
4. A section on the website with a description of the vectorization modes with their advantages and disadvantages 
5. Update PR branch to v1.0.0 branch as this is a large change to v0.29.0
6. Add tests that thoroughly check the existing custom vector environments within gymnasium act basically equivalent to the non-vector versions. 

The thing to do is confirm what happens if the environment spec doesn't have a `vector_entry_point`, do we default to `async` or `sync` or raise an error

PR #630, opened 2023-07-26
[Fix] reset buffer of stateful env MaxAndSkipObservationV0

xmaples:
# Description

Observation buffer should be reset in the stateful observation wrapper `MaxAndSkipObservationV0`.

## Type of change

### DIFF
--- a/gymnasium/wrappers/stateful_observation.py
+++ b/gymnasium/wrappers/stateful_observation.py
@@ -499,3 +499,11 @@ class MaxAndSkipObservationV0(
         max_frame = self._obs_buffer.max(axis=0)
 
         return max_frame, total_reward, terminated, truncated, info
+
+    def reset(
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
@xmaples thanks for this, @LucasAlegre is this correct

LucasAlegre:
Hi @pseudo-rnd-thoughts @xmaples 

I believe this change is not necessary. The step method always fills the buffer before computing the max over the buffer. I can not see why you would need to clear the buffer in the reset method.

pseudo-rnd-thoughts:
@LucasAlegre i think your right and this would only be an issue if n==1 however we do not allow this

xmaples:
You are right, cos of my misunderstanding of MaxAndSkip (skip) with `FrameStack` (stack). Sorry for your time.

PR #620, opened 2023-07-18
ENH: Add NormalizeReward & NormalizeObservation vector wrappers

younik:
# Description

Add `NormalizeObservationV0` and `NormalizeRewardV1` wrappers.

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #632, opened 2023-07-27
ENH: Add NormalizeReward & NormalizeObservation vector wrappers

younik:
Continues from https://github.com/Farama-Foundation/Gymnasium/pull/620

Comment on the wrapper:

> The NormalizeReward wrapper normalizes the standard deviation of the discounted returns of every step in an episode.
As we care only about the magnitude of the standard deviation, the wrapper estimates it with the _discounted past rewards_:
$\gamma^{t} * r(0) + ... \gamma * r(t-1) + r(t), \forall t \in [0, T]$
Instead of the discounted return. This has the advantage of updating the std during the rollout without waiting until the end of the episode. 
>
> **Is the magnitude of the std of the _discounted past rewards_ equal to one of discounted returns?**
In general no, but for a good choice of gamma and typical rewards values, yes.

### DIFF
--- a/gymnasium/wrappers/vector/stateful_observation.py
+++ b/gymnasium/wrappers/vector/stateful_observation.py
@@ -1,0 +1,84 @@
+"""A collection of stateful observation wrappers.
+
+* ``NormalizeObservationV0`` - Normalize the observations
+"""
+from __future__ import annotations
+
+import numpy as np
+
+import gymnasium as gym
+from gymnasium.core import ActType, ObsType
+from gymnasium.experimental.vector.vector_env import VectorObservationWrapper
+from gymnasium.experimental.wrappers.utils import RunningMeanStd
+
+
+__all__ = ["NormalizeObservationV0"]
+
+
+class NormalizeObservationV0(VectorObservationWrapper, gym.utils.RecordConstructorArgs):
+    """This wrapper will normalize observations s.t. each coordinate is centered with unit variance.
+
+    The property `_update_running_mean` allows to freeze/continue the running mean calculation of the observation
+    statistics. If `True` (default), the `RunningMeanStd` will get updated every step and reset call.
+    If `False`, the calculated statistics are used but not updated anymore; this may be used during evaluation.
+
+    Note:
+        The normalization depends on past trajectories and observations will not be normalized correctly if the wrapper was
+        newly instantiated or the policy was changed recently.
+    """
+
+    def __init__(self, env: gym.Env[ObsType, ActType], epsilon: float = 1e-8):
+        """This wrapper will normalize observations s.t. each coordinate is centered with unit variance.
+
+        Args:
+            env (Env): The environment to apply the wrapper
+            epsilon: A stability parameter that is used when scaling the observations.
+        """
+        gym.utils.RecordConstructorArgs.__init__(self, epsilon=epsilon)
+        gym.ObservationWrapper.__init__(self, env)
+
+        self.obs_rms = RunningMeanStd(
+            shape=self.single_observation_space.shape,
+            dtype=self.single_observation_space.dtype,
+        )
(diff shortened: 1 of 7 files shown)


pseudo-rnd-thoughts:
Could we add the comment from the previous pr to the describe of the wrapper. We need to do that on the unnormalized version

gymnasium/wrappers/vector/stateful_reward.py
```diff
@@ -0,0 +1,80 @@
...
+    def update_running_mean(self, setting: bool):
+        """Sets the property to freeze/continue the running mean calculation of the reward statistics."""
+        self._update_running_mean = setting
+
+    def step(
+        self, action: ActType
+    ) -> tuple[ObsType, SupportsFloat, bool, bool, dict[str, Any]]:
+        """Steps through the environment, normalizing the reward returned."""
+        obs, reward, terminated, truncated, info = super().step(action)
+        self.discounted_reward = (
```
pseudo-rnd-thoughts, line 71: I'm still very certain about this value particularly when terminated = 1. Could we change the variable name to more accurately reflect what it means?

tests/experimental/wrappers/vector/test_normalize_observation.py
```diff
@@ -0,0 +1,51 @@
...
+import numpy as np
+import pytest
+
+from gymnasium import spaces
+from gymnasium.vector import AsyncVectorEnv, SyncVectorEnv
+from gymnasium.wrappers.vector import NormalizeObservationV0
+from tests.testing_env import GenericTestEnv
+
+
+@pytest.mark.parametrize("class_", [AsyncVectorEnv, SyncVectorEnv])
```
pseudo-rnd-thoughts, line 11: I don't think we need to test for both. I would just use sync

tests/experimental/wrappers/vector/test_normalize_observation.py
```diff
@@ -0,0 +1,51 @@
...
+def test_normalize_obs(class_):
```
pseudo-rnd-thoughts, line 12: Is it possible to test that the mean and std are very close for the standard wrapper and vector version after ~1000 steps? If there is a known seed this could help

This will actually test that the vector wrapper works very similarly to the normal wrapper

tests/experimental/wrappers/vector/test_normalize_reward.py
```diff
@@ -0,0 +1,53 @@
...
+        return self.observation_space.sample(), float(done), done, False, {}
+
+    def thunk():
+        return GenericTestEnv(step_func=step_func, reset_func=reset_func)
+
+    return thunk
+
+
+@pytest.mark.parametrize("class_", [AsyncVectorEnv, SyncVectorEnv])
+def test_normalize_rew(class_):
```
pseudo-rnd-thoughts, line 30: Could we change the tests in a similar tests to the observation one

pseudo-rnd-thoughts (changes requested):
Thanks for changing to v1.0.0 branch. 
I suggest a way for us to test that wrappers and vector work similarly.
We have a version of this for all wrappers but doesn't work for n>1

gymnasium/wrappers/vector/stateful_observation.py
```diff
@@ -0,0 +1,84 @@
...
+    def vector_observation(self, observation: ObsType) -> ObsType:
+        """Defines the vector observation normalization function.
+
+        Args:
+            observation: A vector observation from the environment
+
+        Returns:
+            the normalized observation
+        """
+        return self._normalize_observations(observation)
```
pseudo-rnd-thoughts, line 66: Do we not need to iterate over the vector observations to normalize each observation?

pseudo-rnd-thoughts (approved):
Looks good

Issue #377, opened 2023-03-13
[Bug Report] Observation is outside the given space. 


n30111:
### Describe the bug

The changes in the `LunarLander` observation space is inaccurate and not capturing all possible values that the environment generate. 

In the original gym <= 0.23, observation space : https://github.com/openai/gym/blob/252c333337eaee4dd6a48b52ef32c0763dd14d65/gym/envs/box2d/lunar_lander.py#L162

Which was later changed to : https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/envs/box2d/lunar_lander.py#L320 

The changes doesn't cover all possible scenarios. Which results in observation outside the given space error while training agents on the environment. 

```
ray.exceptions.RayTaskError(ValueError): ray::RolloutWorker.apply() (pid=24754, ip=10.5.0.4, repr=<ray.rllib.evaluation.rollout_worker.RolloutWorker object at 0x7fe687b04580>)
  File "/lib/python3.8/site-packages/ray/rllib/utils/actor_manager.py", line 183, in apply
    raise e
  File "/lib/python3.8/site-packages/ray/rllib/utils/actor_manager.py", line 174, in apply
    return func(self, *args, **kwargs)
  File "/lib/python3.8/site-packages/ray/rllib/execution/rollout_ops.py", line 86, in <lambda>
    lambda w: w.sample(), local_worker=False, healthy_only=True
  File "/lib/python3.8/site-packages/ray/rllib/evaluation/rollout_worker.py", line 914, in sample
    batches = [self.input_reader.next()]
  File "/lib/python3.8/site-packages/ray/rllib/evaluation/sampler.py", line 92, in next
    batches = [self.get_data()]
  File "/lib/python3.8/site-packages/ray/rllib/evaluation/sampler.py", line 277, in get_data
    item = next(self._env_runner)
  File "/lib/python3.8/site-packages/ray/rllib/evaluation/sampler.py", line 616, in _env_runner
    active_envs, to_eval, outputs = _process_observations(
  File "//lib/python3.8/site-packages/ray/rllib/evaluation/sampler.py", line 850, in _process_observations
    prep_obs = preprocessor.transform(raw_obs)
  File "/lib/python3.8/site-packages/ray/rllib/models/preprocessors.py", line 206, in transform
    self.check_shape(observation)
  File "lib/python3.8/site-packages/ray/rllib/models/preprocessors.py", line 74, in check_shape
    raise ValueError(
[...]

PR #633, opened 2023-07-28
Fix Lunar Lander termination criteria

jjshoots:
# Description

This fixes lunar lander termination criteria.

Fixes [this](https://github.com/Farama-Foundation/Gymnasium/issues/377#issuecomment-1655610667)


pseudo-rnd-thoughts (changes requested):
Without a version change we can't do this therefore I would preemptively reject this PR

To my knowledge, all the issue requires changing is the bound

jjshoots:
@pseudo-rnd-thoughts I'm not sure if that's an ideal thing to do. As it stands the environment doesn't check for when the lander goes out of the map in the y direction, only x. So in theory, the lander could be y = 1000 and the environment would still be valid as long as x is in [-1, 1]. If we change the bounds, what it should be is kind of weird since no matter what number we give it, there will always be a valid y value outside of that.

The change I made simply makes it so that the environment additionally checks the y coordinate. I agree that this requires a version bump though.

pseudo-rnd-thoughts:
I agree that it doesn't make much sense in terms of visualisation currently but this doesn't break the environment specification as it is currently. To me, the cost of version bumping to benefit of adding this is too much

Therefore, I would set the y bound to something quite high, 1000 which is difficult to achieve

jjshoots:
That's fair, I'll modify the bounds then. Although I should mention that in some instances, the observation space bounds are used for normalisation of the observation itself, so this may result in previously trained models performing differently. Although, agreed that just changing the bounds is a less disruptive change.

pseudo-rnd-thoughts:
We have modified the observation two or three times over the last few versions (which we should note somewhere)

PR #635, opened 2023-07-28
Allow changing the size of the viewing window for human rendering

logan-dunbar:
# Description

In debugging I often want to view what's happening with human rendering, but having the window take up a quarter of my screen every time it opened and me needing to resize it with the mouse was frustrating. This change allows (width=, height=) to be passed to gym.make() which will set the human window size properly.

With the way I have currently implemented it the "quarter of screen area" that is currently the default will never be hit because of the defaults on width/height that presently exist. Perhaps we could add a human_width/human_height optionals which will still allow the current default behaviour and then allow customizing with this pull requests changes.

## Type of change

# Checklist:


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good to me. I can see why this is so helpful

pseudo-rnd-thoughts:
Thanks for the PR, looks good to me.

pseudo-rnd-thoughts (approved):

PR #638, opened 2023-08-01
optimize reinforce_invpend_gym_v26.py

mzamini92:
Can we calculate the mean of log probabilities outside the loop and use the `torch.stack` function, to avoid redundant computations and make the update step more efficient? This should improve the performance of the update function and speed up the training process.

# Description

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/docs/tutorials/training_agents/reinforce_invpend_gym_v26.py
+++ b/docs/tutorials/training_agents/reinforce_invpend_gym_v26.py
@@ -207,6 +207,10 @@ class REINFORCE:
         log_prob_mean = log_probs.mean()
     
         # Update the loss with the mean log probability and deltas
+        # Note: The multiplication here is not a standard element-wise multiplication.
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks, I'm just thinking about this for unfamiliar users (like myself)

Is `loss = -log_prob_mean * deltas.mean()` an element-wise operation as `log_prob_mean` and `deltas` are tensor? 
If so, do we need not need to compute the mean over the resulting tensor? 

Could you add a comment explain that the multiple is an element-wise operation as I think would be unclear to most users initially

mzamini92 (commented):
Added comments

mzamini92 (commented):
@pseudo-rnd-thoughts you were right I forgot to add sum. modified it.

PR #639, opened 2023-08-01
Update type hints for all internal environments

robertoschiavone:
# Description

This commit adds strict type checking to all environments, comforming them to the documentation.

It shouldn't break anything and no new tests should be required.

## Type of change
# Checklist:

### DIFF
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -1,11 +1,14 @@
+from __future__ import annotations
+
 from os import path
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import Any, SupportsFloat
 
 import numpy as np
 from numpy.typing import NDArray
 
 import gymnasium as gym
 from gymnasium import error, logger, spaces
+from gymnasium.core import ActType, ObsType, RenderFrame
 from gymnasium.spaces import Space
 
 
@@ -35,11 +38,11 @@ class BaseMujocoEnv(gym.Env[NDArray[np.float64], NDArray[np.float32]]):
         model_path,
         frame_skip,
         observation_space: Space,
-        render_mode: Optional[str] = None,
+        render_mode: str | None = None,
         width: int = DEFAULT_SIZE,
         height: int = DEFAULT_SIZE,
-        camera_id: Optional[int] = None,
-        camera_name: Optional[str] = None,
+        camera_id: int | None = None,
+        camera_name: str | None = None,
     ):
         """Base abstract class for mujoco based environments.
 
@@ -102,8 +105,8 @@ class BaseMujocoEnv(gym.Env[NDArray[np.float64], NDArray[np.float32]]):
     # methods to override:
     # ----------------------------
     def step(
-        self, action: NDArray[np.float32]
-    ) -> Tuple[NDArray[np.float64], np.float64, bool, bool, Dict[str, np.float64]]:
+        self, action: ActType
+    ) -> tuple[ObsType, SupportsFloat, bool, bool, dict[str, Any]]:
         raise NotImplementedError
 
     def reset_model(self) -> NDArray[np.float64]:
@@ -113,7 +116,7 @@ class BaseMujocoEnv(gym.Env[NDArray[np.float64], NDArray[np.float32]]):
         """
         raise NotImplementedError
 
-    def _initialize_simulation(self) -> Tuple[Any, Any]:
+    def _initialize_simulation(self) -> tuple[Any, Any]:
         """
         Initialize MuJoCo simulation data structures mjModel and mjData.
         """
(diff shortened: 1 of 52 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for doing this, could you add `from __future__ import annotations` as the first line of code then run pre-commit. This will update the code to nicer type annotations in cases

pseudo-rnd-thoughts (approved):

gymnasium/envs/mujoco/mujoco_env.py
```diff
@@ -309,7 +312,7 @@ def render(self):
 
     def _get_viewer(
         self, mode
-    ) -> Union["mujoco_py.MjViewer", "mujoco_py.MjRenderContextOffscreen"]:
+    ) -> mujoco_py.MjViewer | mujoco_py.MjRenderContextOffscreen:
```
Kallinteris-Andreas, line 315: This requires `mujoco_py` to be imported,
if `mujoco_py` is not imported does this cause issues?

gymnasium/envs/mujoco/mujoco_env.py
```diff
@@ -380,7 +383,7 @@ def __init__(
 
     def _initialize_simulation(
         self,
-    ) -> Tuple["mujoco._structs.MjModel", "mujoco._structs.MjData"]:
+    ) -> tuple[mujoco._structs.MjModel, mujoco._structs.MjData]:
```
Kallinteris-Andreas, line 386: same here

gymnasium/envs/mujoco/mujoco_env.py
Kallinteris-Andreas, line 315: This requires `mujoco_py` to be imported,
if `mujoco_py` is not imported does this cause issues?

Kallinteris-Andreas (commented):
besides those i am good with the `mujoco` envs

pseudo-rnd-thoughts:
@robertoschiavone Could you test if @Kallinteris-Andreas comments do cause an issue with mujoco or mujoco-py is not installed

gymnasium/envs/mujoco/mujoco_env.py
robertoschiavone: No more than usual. There's an "import guard" (not sure it's the right way to call it) at the beginning of the file

```python
try:
    import mujoco_py
except ImportError as e:
    MUJOCO_PY_IMPORT_ERROR = e
else:
    MUJOCO_PY_IMPORT_ERROR = None
  ```

and later on an exception gets thrown

```python
        if MUJOCO_PY_IMPORT_ERROR is not None:
            raise error.DependencyNotInstalled(
                f"{MUJOCO_PY_IMPORT_ERROR}. "
                "(HINT: you need to install mujoco-py, and also perform the setup instructions "
                "here: https://github.com/openai/mujoco-py.)"
            )
```

Therefore `_get_viewer` is effectively unreachable if `mujoco` is not installed.

RedTachyon:
I left a few comments inside of the code, but I'm gonna pull that out to the top level here.

I'm not 100% sure, but maybe like 90% sure that adding the type hints of `ObsType` etc. is actually harmful and breaks the type system.

Attaching two screenshots. One is PyCharm's type analysis with this PR merged:
image

The second one is after removing the `ObsType` type hint from `CartPoleEnv.reset`:
image


The conclusion is that with this merged, in `obs, info = env.reset()`, `obs` is only detected as an `ObsType` (whereas we know it should be `ndarray`), which is a generic without any bounds and not attached to anything, so it's pretty much a synonym for `Any`. We see this because the type analysis detects it as just `ObsType`, and doesn't complain about calling a non-existent `ndarray.test()` method.

Removing (part) of the changes of this PR results in correct behavior - `obs` is detected as `ndarray`, and trying to do `ndarray.test()` correctly raises a type warning.


@pseudo-rnd-thoughts I'd strongly consider reverting this and maybe redoing it in a smaller, more careful scope. We don't need to have explicit type hints for every subclass, that's what the generic is for, but if we want to add them, we have to add them correctly.

gymnasium/envs/mujoco/mujoco_env.py
RedTachyon: I think this is only true due to `from __future__ import annotations`, so it's not necessarily a problem in this case, but we need to be careful about it.

pseudo-rnd-thoughts:
@RedTachyon I think your are right partially but we just need replace `ObsType` and `ActType` with `np.ndarray` in most cases plus `Env[np.ndarray, np.ndarray]` in the `class` definition

RedTachyon:
Yea, my point was that adding explicit `np.ndarray` type hints (or the equivalent for some envs) is probably harmless, but fundamentally inconsequential for the type system if `Env[np.ndarray, np.ndarray]` is also specified. But the way it's done now, it doesn't work (also `SupportsFloat` shouldn't be used here, since these specific envs just return a float)

pseudo-rnd-thoughts:
@RedTachyon I'm happy to add a REWARD type with float (maybe int and np.ndarray) for v1.0.0

RedTachyon:
@pseudo-rnd-thoughts I think these are two separate issues? Even if the base annotation on `Env` is some `Reward`, however it would be defined, the specific environments can/should still be annotated as simply `float`, since that's literally what they return, we don't need any indirection

PR #640, opened 2023-08-01
refactor: EzPickle -> @ezpickle :recycle:

robertoschiavone:
# Description

This commit changes EzPickle from being a class to being a decorator.

The edits are only internal, they don't change or break the existing APIs.

Pros:

- Removed every instance of multiple inherintance in the codebase.
- Automated serialization of constructor arguments, no need to manually copy/paste them and keep track of them in case the constructor changes.
- I discovered at least one instance of useless arguments, where such argument is only serialized and then never used anymore (`default_camera_config` in `ant_v5.py`).

Cons:

- none that I can think of
  
## Type of change

- [x] Refactor

# Checklist:

### DIFF
--- a/docs/tutorials/training_agents/reinforce_invpend_gym_v26.py
+++ b/docs/tutorials/training_agents/reinforce_invpend_gym_v26.py
@@ -193,32 +193,31 @@ class REINFORCE:
         """Updates the policy network's weights."""
         running_g = 0
         gs = []
-    
+
         # Discounted return (backwards) - [::-1] will return an array in reverse
         for R in self.rewards[::-1]:
             running_g = R + self.gamma * running_g
             gs.insert(0, running_g)
-    
+
         deltas = torch.tensor(gs)
-    
+
         log_probs = torch.stack(self.probs)
-    
+
         # Calculate the mean of log probabilities for all actions in the episode
         log_prob_mean = log_probs.mean()
-    
+
         # Update the loss with the mean log probability and deltas
         # Now, we compute the correct total loss by taking the sum of the element-wise products.
         loss = -torch.sum(log_prob_mean * deltas)
-    
+
         # Update the policy network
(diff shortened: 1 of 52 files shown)


RedTachyon:
Worth noting that this absolutely is a breaking change, since 3rd-party envs can use `EzPickle` to keep consistency with internal gymnasium envs, since `EzPickle` is explicitly an exported class from its module.

In principle I'm not opposed to removing multiple inheritance, but we need to be very careful about this. At the very least we'd probably need to keep `EzPickle` (the class) around for a long time. The only point where I can see it being removable is for 1.0, but even then I'm not sure if it's worth the hassle. 

Even if we keep `EzPickle`, that doesn't mean we can't convert internal code to something better, but I'm not sure if a decorator applied to the constructor is the right way to go, I haven't seen a pattern like this... ever, probably for a good reason. Modifying the class methods inside of it is even more opaque.

With all my distaste for multiple inheritance, in this case it might not be the worst thing, since all it should do is just record some things and then pass the arguments on to the next parent class. But as I said, if there's a solution that's significantly less cursed (as a ratio to how breaking of a change it is)

RedTachyon (changes requested):
Just marking the review to avoid an accidental merge

Kallinteris-Andreas:
Even if `@ezpickle` is added, `EzPickle` should be kept for backwards compatibility with a deprecation warning

pseudo-rnd-thoughts (changes requested):
While I agree that this is a relatively elegant solution and prevents the needs for listing all of the arguments. 
I think there is another issue not mentioned is the type checking, I might be wrong but if you initialise an Env, the environment type is not `Env` but an `ezpickle` function

pseudo-rnd-thoughts:
Due to the issue I list above, unless the issue doesn't exist or we can find a solution, I'm going to close the issue

PR #642, opened 2023-08-01
fix: gitignore :wrench:

robertoschiavone:
# Description

A fresh gitignore. Despite having way more lines, it's easier to maintain since it's autogenerated, and it covers more tools, editors, operating systems.

I, for example, use `pyenv` and I have to constantly remind myself not to accidentally commit my local `.python-version`. Now I don't have to worry anymore 🙃 

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,43 +1,416 @@
-*.swp
-*.pyc
-*.py~
-.DS_Store
-.cache
-.pytest_cache/
-__pycache__/
+# Created by https://www.toptal.com/developers/gitignore/api/git,python,sublimetext,jetbrains+all,jupyternotebooks,visualstudiocode,macos,linux,windows
+# Edit at https://www.toptal.com/developers/gitignore?templates=git,python,sublimetext,jetbrains+all,jupyternotebooks,visualstudiocode,macos,linux,windows
 
-# Setuptools distribution and build folders.
-/dist/
-/build
-/wheels
-/wheelhouse
+### Git ###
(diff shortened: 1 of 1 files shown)


RedTachyon:
We probably don't need a fully automatically-generatable .gitignore, so please make sure that all the entries that were present previously, are also present now. Some of them seem rather specific to gymnasium, like vizdoom.ini, rllab_mujoco etc. so they won't get automatically picked up by the toptal generator. While I don't know the exact story behind the inclusion of each of those items, we should assume they're here for a reason and not remove them without a good reason otherwise.

.gitignore
```diff
@@ -1,43 +1,416 @@
...
+Thumbs.db
+Thumbs.db:encryptable
+ehthumbs.db
+ehthumbs_vista.db
+
+# Dump file
+*.stackdump
 
-junk
-MUJOCO_LOG.txt
```
Kallinteris-Andreas, line 31: should not be removed

.gitignore
```diff
@@ -1,43 +1,416 @@
...
+# Folder config file
+[Dd]esktop.ini
 
-rllab_mujoco
```
Kallinteris-Andreas, line 33: should not be removed

Kallinteris-Andreas (commented):
I am generally against 'generic' `.gitignore` files, only project specific `temp/cache` files should be excluded, it is the responsibility of the developer to edit it locally to ignore your IDE files

PR #643, opened 2023-08-02
Revert "Update type hints for all internal environments"

RedTachyon:
Reverts Farama-Foundation/Gymnasium#639

See my comment on the orginal PR

### DIFF
--- a/gymnasium/envs/box2d/bipedal_walker.py
+++ b/gymnasium/envs/box2d/bipedal_walker.py
@@ -1,16 +1,12 @@
-from __future__ import annotations
-
-
 __credits__ = ["Andrea PIERRÉ"]
 
 import math
-from typing import TYPE_CHECKING, Any, SupportsFloat
+from typing import TYPE_CHECKING, List, Optional
 
 import numpy as np
 
 import gymnasium as gym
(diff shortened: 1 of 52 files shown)


pseudo-rnd-thoughts:
While your comment is correct, I think it only applies to the `ObsType` and `ActType` rather than the rest of the type hints. 
Therefore, I am uncertain if we need to revert all of the changes or just the `ObsType` and `ActType` ones

RedTachyon:
From what I see, a good majority of changes in this was adding `ObsType` and `ActType` to `reset` and `step` signatures. Every single one of those is incorrect/harmful, so I think it's much safer to just revert the whole thing, and then readd the correct changes. The alternative is making a new PR selectively undoing these changes (so going through each environment, changing `ObsType`, `ActType`, `SupportsFloat` and checking if there's any more bugs), and then doing a three-way code review to make sure everything is correct and every env is actually updated (reverted?), which is a nightmare. If we revert, we simply make a new PR, and make sure there are no bugs in the new PR.

I know reverting PRs is a pain, but I do think it's the right move when a PR introduces new bugs, and this 100% did, plus it's pretty huge (in terms of just the number of changes)

PR #657, opened 2023-08-07
Add more examples to docs for wrappers

jjshoots:
# Description

This adds more example code to the docs for the wrappers.

Todo:

- [x] atari_preprocessing.py
- [x] jax_to_numpy.py
- [x] jax_to_torch.py
- [x] lambda_action.py
- [x] lambda_observation.py
- [x] lambda_reward.py
- [x] numpy_to_torch.py
- [x] rendering.py
- [x] stateful_action.py
- [x] stateful_observation.py
- [x] stateful_reward.py

### DIFF
--- a/gymnasium/wrappers/jax_to_numpy.py
+++ b/gymnasium/wrappers/jax_to_numpy.py
@@ -103,6 +103,29 @@ class JaxToNumpyV0(
     Notes:
         The Jax To Numpy and Numpy to Jax conversion does not guarantee a roundtrip (jax -> numpy -> jax) and vice versa.
         The reason for this is jax does not support non-array values, therefore numpy ``int_32(5) -> DeviceArray([5], dtype=jnp.int23)``
+
+    Example:
+        >>> import gymnasium as gym
+        >>> env = gym.make("JaxEnv-vx")
+        >>> env = JaxToNumpyV0(env)
+        >>> obs, _ = env.reset(seed=123)
+        >>> type(obs)
+        <class 'numpy.ndarray'>
+
+        >>> action = env.action_space.sample()
+        >>> obs, reward, terminated, truncated, info = env.step(action)
+        >>> type(obs)
+        <class 'numpy.ndarray'>
+
+        >>> type(reward)
+        <class 'float'>
+
+        >>> type(terminated)
+        <class 'bool'>
+
+        >>> type(truncated)
+        <class 'bool'>
+
     """
 
     def __init__(self, env: gym.Env[ObsType, ActType]):
--- a/gymnasium/wrappers/jax_to_numpy.py
+++ b/gymnasium/wrappers/jax_to_numpy.py
@@ -111,21 +111,16 @@ class JaxToNumpyV0(
         >>> obs, _ = env.reset(seed=123)
         >>> type(obs)
         <class 'numpy.ndarray'>
-
         >>> action = env.action_space.sample()
         >>> obs, reward, terminated, truncated, info = env.step(action)
         >>> type(obs)
         <class 'numpy.ndarray'>
-
         >>> type(reward)
         <class 'float'>
-
         >>> type(terminated)
         <class 'bool'>
-
         >>> type(truncated)
         <class 'bool'>
-
     """
 
     def __init__(self, env: gym.Env[ObsType, ActType]):
--- a/gymnasium/wrappers/jax_to_numpy.py
+++ b/gymnasium/wrappers/jax_to_numpy.py
@@ -105,21 +105,21 @@ class JaxToNumpyV0(
         The reason for this is jax does not support non-array values, therefore numpy ``int_32(5) -> DeviceArray([5], dtype=jnp.int23)``
 
     Example:
-        >>> import gymnasium as gym                                     # doctest: SKIP
-        >>> env = gym.make("JaxEnv-vx")                                 # doctest: SKIP
-        >>> env = JaxToNumpyV0(env)                                     # doctest: SKIP
-        >>> obs, _ = env.reset(seed=123)                                # doctest: SKIP
-        >>> type(obs)                                                   # doctest: SKIP
+        >>> import gymnasium as gym                                     # doctest: +SKIP
+        >>> env = gym.make("JaxEnv-vx")                                 # doctest: +SKIP
+        >>> env = JaxToNumpyV0(env)                                     # doctest: +SKIP
+        >>> obs, _ = env.reset(seed=123)                                # doctest: +SKIP
+        >>> type(obs)                                                   # doctest: +SKIP
         <class 'numpy.ndarray'>
-        >>> action = env.action_space.sample()                          # doctest: SKIP
-        >>> obs, reward, terminated, truncated, info = env.step(action) # doctest: SKIP
-        >>> type(obs)                                                   # doctest: SKIP
+        >>> action = env.action_space.sample()                          # doctest: +SKIP
+        >>> obs, reward, terminated, truncated, info = env.step(action) # doctest: +SKIP
+        >>> type(obs)                                                   # doctest: +SKIP
         <class 'numpy.ndarray'>
-        >>> type(reward)                                                # doctest: SKIP
+        >>> type(reward)                                                # doctest: +SKIP
         <class 'float'>
-        >>> type(terminated)                                            # doctest: SKIP
+        >>> type(terminated)                                            # doctest: +SKIP
         <class 'bool'>
-        >>> type(truncated)                                             # doctest: SKIP
+        >>> type(truncated)                                             # doctest: +SKIP
         <class 'bool'>
     """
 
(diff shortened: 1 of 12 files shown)


pseudo-rnd-thoughts:
In going to try to fix the normalize reward test in #659

gymnasium/wrappers/jax_to_numpy.py
```diff
@@ -103,6 +103,29 @@ class JaxToNumpyV0(
     Notes:
         The Jax To Numpy and Numpy to Jax conversion does not guarantee a roundtrip (jax -> numpy -> jax) and vice versa.
         The reason for this is jax does not support non-array values, therefore numpy ``int_32(5) -> DeviceArray([5], dtype=jnp.int23)``
+
+    Example:
+        >>> import gymnasium as gym
+        >>> env = gym.make("JaxEnv-vx")
```
pseudo-rnd-thoughts, line 109: This will fail because `JaxEnv-vx` doesn't exist and annoying we don't have any environments that return jax data currently

gymnasium/wrappers/jax_to_numpy.py
```diff
@@ -103,6 +103,29 @@ class JaxToNumpyV0(
...
+        >>> env = JaxToNumpyV0(env)
+        >>> obs, _ = env.reset(seed=123)
+        >>> type(obs)
+        <class 'numpy.ndarray'>
+
+        >>> action = env.action_space.sample()
+        >>> obs, reward, terminated, truncated, info = env.step(action)
+        >>> type(obs)
+        <class 'numpy.ndarray'>
+
```
pseudo-rnd-thoughts, line 119: Remove the new lines

gymnasium/wrappers/jax_to_torch.py
```diff
@@ -121,6 +121,30 @@ class JaxToTorchV0(gym.Wrapper, gym.utils.RecordConstructorArgs):
...
+        >>> env = JaxtoTorchV0(env)
+        >>> obs, _ = env.reset(seed=123)
+        >>> type(obs)
+        <class 'torch.Tensor'>
+
+        >>> action = torch.tensor(env.action_space.sample())
+        >>> obs, reward, terminated, truncated, info = env.step(action)
+        >>> type(obs)
+        <class 'torch.Tensor'>
+
```
pseudo-rnd-thoughts, line 138: Remove the new lines

gymnasium/wrappers/lambda_action.py
```diff
@@ -21,7 +21,19 @@
 class LambdaActionV0(
     gym.ActionWrapper[ObsType, WrapperActType, ActType], gym.utils.RecordConstructorArgs
 ):
-    """Applies a function to the ``action`` before passing the modified value to the environment ``step`` function."""
+    """Applies a function to the ``action`` before passing the modified value to the environment ``step`` function.
+
+    Example:
```
pseudo-rnd-thoughts, line 26: Is there a way of showing the impact of the LambdaAction wrapper? Currently this doesn't show that the action has been modified

gymnasium/wrappers/lambda_observation.py
```diff
@@ -564,6 +565,31 @@ class RenderObservationV0(
...
        This was previously called ``PixelObservationWrapper``.
+
+    Example:
+        Replace the observation with the rendered image:
+        >>> import gymnasium as gym
+        >>> env = gym.make("CartPole-v1", render_mode="rgb_array")
+        >>> env = RenderObservationV0(env, render_only=True)
+        >>> obs, _ = env.reset(seed=123)
+        >>> image = env.render()
+        >>> (obs == image).all()
```
pseudo-rnd-thoughts, line 576: Could you change to `np.all(obs == image)` to be more clear to the user

gymnasium/wrappers/lambda_observation.py
```diff
@@ -564,6 +565,31 @@ class RenderObservationV0(
 
     Notes:
        This was previously called ``PixelObservationWrapper``.
+
+    Example:
+        Replace the observation with the rendered image:
+        >>> import gymnasium as gym
+        >>> env = gym.make("CartPole-v1", render_mode="rgb_array")
+        >>> env = RenderObservationV0(env, render_only=True)
```
pseudo-rnd-thoughts, line 573: Add an `env.observation_space` line and `env.step`

gymnasium/wrappers/lambda_observation.py
```diff
@@ -564,6 +565,31 @@ class RenderObservationV0(
...
+        >>> obs, _ = env.reset(seed=123)
+        >>> image = env.render()
+        >>> (obs == image).all()
+        True
+
+        Add the rendered image to the original observation as a dictionary item:
+        >>> import gymnasium as gym
+        >>> env = gym.make("CartPole-v1", render_mode="rgb_array")
+        >>> env = RenderObservationV0(env, render_only=False)
```
pseudo-rnd-thoughts, line 582: Add an `env.observation_space` line

gymnasium/wrappers/lambda_observation.py
```diff
@@ -564,6 +565,31 @@ class RenderObservationV0(
...
+        >>> obs, _ = env.reset(seed=123)
+        >>> obs.keys()
+        dict_keys(['state', 'pixels'])
+
+        >>> obs["state"]
+        array([ 0.01823519, -0.0446179 , -0.02796401, -0.03156282], dtype=float32)
+
+        >>> (obs["pixels"] == env.render()).all()
```
pseudo-rnd-thoughts, line 590: Change to `np.all`

gymnasium/wrappers/lambda_action.py
pseudo-rnd-thoughts: We can probably have two environments, one with an unmodified environment and a modified environment and step through the different actions and should the same output as the LambdaAction modifies the action

pyproject.toml
```diff
@@ -154,3 +154,4 @@ reportUnboundVariable = "warning"
 
 [tool.pytest.ini_options]
 filterwarnings = ["ignore::DeprecationWarning:gymnasium.*:"]
+addopts = "--ignore=gymnasium/wrappers/jax_to_numpy.py --ignore=gymnasium/wrappers/jax_to_torch.py"
```
pseudo-rnd-thoughts, line 157: Does this ignore the tests as well?

pseudo-rnd-thoughts (changes requested):

gymnasium/wrappers/jax_to_numpy.py
jjshoots: Theoretically this file will be ignored by pydoctest due to the changes in the `pyproject.toml` file.

pyproject.toml
pseudo-rnd-thoughts: Ohh, I missed that. Maybe that is fine but we should note that somewhere

pseudo-rnd-thoughts:
Also, update to v1.0.0 which has fixed the CI issues

pyproject.toml
jjshoots: The alternative is [this](https://github.com/Farama-Foundation/Gymnasium/blob/v1.0.0/gymnasium/wrappers/vector/record_episode_statistics.py#L26) but I'm not so sure that's a good idea since it seems possible to lose track of how many skips are in the repo itself.

jjshoots:
@pseudo-rnd-thoughts The remaining bug is not caused from my end I think. I think I'd want to get this merged before I start dealing with vector wrappers.

pseudo-rnd-thoughts:
I reran the test and it seems there are still issues with doctests

jjshoots:
@pseudo-rnd-thoughts Do you have any idea how to make doctest ignore the output of a certain function? Tried a bunch of things with the remaining error and non of them stuck.

pseudo-rnd-thoughts:
"+doctest: skip" I think there are some examples in the project already

jjshoots:
`doctest: +skip` skips the the code though, this makes sense for things like `JaxEnv-vx`, but the video record function works, except the printout depends on the system in use. Ideally doctest should run it to check that the videos are indeed saved and there are a correct number of them, but ignore the printouts that show where the videos are saved.

pseudo-rnd-thoughts:
@jjshoots I just remembered that `RecordVideo` has a `disable_logger` parameter to prevent this from printing much
Could you use this to make the doctests work?

gymnasium/wrappers/rendering.py
```diff
@@ -195,7 +195,7 @@ class RecordVideoV0(
         ...
         >>> env.close()
         >>> len(os.listdir("./save_videos2"))
-        3
+        1
```
pseudo-rnd-thoughts, line 198: I agree that this is the correct value it doesn't seem to make sense with the test

jjshoots:
@pseudo-rnd-thoughts Ready for merge/review.

pseudo-rnd-thoughts (approved):
Thanks for making the changes, looks good now

PR #673, opened 2023-08-18
add examples for vector wrappers

jjshoots:
### Description

This adds examples for the vector wrappers.

### To-Do:

- [x] dict_info_to_list.py
- [x] jax_to_numpy.py
- [x] jax_to_torch.py
- [x] numpy_to_torch.py
- [x] common.py
    - The pprint at the end is skipped because the value of `info["episode"]["t"]` is the time taken per episode, which has run-to-run variation.
- [x] stateful_observation.py
    - `DtypeObservation` will fail because of #705.
- [x] stateful_reward.py
- [x] vectorize_action.py
- [x] vectorize_observation.py
- [x] vectorize_reward.py


pseudo-rnd-thoughts:
FYI, I have found that the CI will hang if you don't close a vector env, in particular with async that render stuff

jjshoots:
@pseudo-rnd-thoughts Gotcha, I was trying to figure that out.

gymnasium/wrappers/vector/vectorize_observation.py
```diff
@@ -53,10 +86,49 @@ def single_observation(self, observation: ObsType) -> ObsType:
...
+                [ 0.02852531,  0.02858594,  0.0469136 ,  0.02480598]],
+        <BLANKLINE>
+               [[ 0.03517495, -0.000635  , -0.01098382, -0.03203924],
+                [ 0.03517495, -0.000635  , -0.01098382, -0.03203924]]],
+              dtype=float32)
+    """
 
-    class VectorizedEnv(Env):
-        """Fake single-agent environment uses for the single-agent wrapper."""
+    class _SingleEnv(Env):
```
pseudo-rnd-thoughts, line 130: Is there a reason why you changed the name? I think the name is better but is there any other reason?

pseudo-rnd-thoughts (changes requested):
Otherwise, looks good

gymnasium/wrappers/vector/dict_info_to_list.py
```diff
@@ -24,26 +24,16 @@ class DictInfoToList(VectorWrapper):
 
     Example:
         >>> import numpy as np
-        >>> dict_info = {
```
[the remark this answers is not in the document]
jjshoots: Initially when reading it it seemed redundant, but now that you mention it, I probably shouldn't have. Going to put it back in.

gymnasium/wrappers/vector/vectorize_observation.py
pseudo-rnd-thoughts: Fair, that makes sense

pseudo-rnd-thoughts (approved):

PR #687, opened 2023-08-26
Avoid alpha versions from being the latest version in the docs

mgoulao:
Added an if statement to the docs versioning workflow that only builds the latest version of the docs if the version name doesn't contain an "a".


pseudo-rnd-thoughts (approved):
Thanks for the PR

PR #698, opened 2023-08-30
Update sync_vector_env.py

mzamini92:
Isn't it better to use map() function which is a built-in function that is optimized for performance, so it will run faster than a for loop. Additionally, the map() function does not need to store the results of the function calls in memory, which can save memory.

# Description

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


pseudo-rnd-thoughts:
You might be right that `map` is faster than a for loop but as `reset` is run so infrequently then I don't think that this change is worth it. Until you can show that this significantly increases performance, I'm going to close this

Issue #705, opened 2023-09-04
[Bug Report] `DtypeObservation` not working for `float32` -> `float64` in vector envs


jjshoots:
### Describe the bug

Not sure if this is intended, but the `gymnasium.wrappers.vector.DtypeObservation` doesn't work for converting a `float32` env into a `float64` one.

### Code example

```python
>>> import numpy as np
>>> import gymnasium as gym
>>> envs = gym.make_vec("CartPole-v1", num_envs=3)
>>> obs, info = envs.reset(seed=123)
>>> envs.close()
>>> obs.dtype
dtype('float32')
>>> envs = gym.make_vec("CartPole-v1", num_envs=3)
>>> envs = DtypeObservation(envs, dtype=np.float64)
>>> obs, info = envs.reset(seed=123)
>>> envs.close()
>>> obs.dtype
dtype('float32')
```

### System info

Gymnasium v1.0.0

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
https://github.com/Farama-Foundation/Gymnasium/blob/74e6e13b78188068fdf5a595fb409926dec48fa8/gymnasium/spaces/box.py#L277 is the cause.
I can't remember why I have this commented out, I suspect that there is an issue with something else, I've made #707 to experiment

pseudo-rnd-thoughts:
Fixed by #707

PR #707, opened 2023-09-05
Add dtype check to `Box.__eq__` and `MultiDiscrete.__eq__`

pseudo-rnd-thoughts:
# Description

Reported an issue with `wrappers.vector.DtypeObservation` which is caused by checking if two observation spaces however the `Box` space doesn't check the `dtype` in its equivalence functions (https://github.com/Farama-Foundation/Gymnasium/issues/705)
This PR updates the `Box` and `MultiDiscrete` spaces to check `dtype`

### DIFF
--- a/gymnasium/spaces/box.py
+++ b/gymnasium/spaces/box.py
@@ -274,7 +274,7 @@ class Box(Space[NDArray[Any]]):
         return (
             isinstance(other, Box)
             and (self.shape == other.shape)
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
I'm sure why this wasn't included in the previous PR, I can't see any commit that reverted the change so I'm going to merge it

PR #708, opened 2023-09-07
Update env_checker.py

chrisyeh96:
Fix typo about Box maximum limit. Add return type annotations.

### DIFF
--- a/gymnasium/utils/env_checker.py
+++ b/gymnasium/utils/env_checker.py
@@ -63,7 +63,7 @@ def data_equivalence(data_1, data_2) -> bool:
         return False
 
 
-def check_reset_seed(env: gym.Env):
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts:
Thanks for spotting that, could you fix the tests. You should just need to update the `IGNORE_WARNINGS` list to include the new case

chrisyeh96:
@pseudo-rnd-thoughts: I have updated the `CHECK_ENV_IGNORE_WARNINGS` list

chrisyeh96:
Updated the `CHECK_ENV_IGNORE_WARNINGS` in two more test files. This should fix the failing tests.

pseudo-rnd-thoughts (approved):
Amazing, thanks for the changes

Issue #735, opened 2023-10-09
[Bug Report] "Basic usage" documentation incorrect/incomplete


smarianimore:
### Describe the bug

**Reproduce**:
1. Copy code [here](https://gymnasium.farama.org/content/basic_usage/#interacting-with-the-environment)
2. pip install gymnasium
3. execute code 

**Expectation**: see "lunar lander" screen appear and game playing randomly

**Actual**: error in missing dependency from box2d

_Weird fact 1_: error messages refer to "bipedal walker" env (unrelated to lunar lander from user perspective)

**Solution**:
1. pip install swig
2. pip install gymnasium[box2d]

_Weird fact 2_: I came up with solution 1 because trying to do 2 directly resulting in error "failed building wheel" and had to google search solution cause no error message pointed to missing dependency from swig package

**Suggestion**: either fix gymnasium check to avoid checking envs the user is not using or improve documentation [here](https://gymnasium.farama.org/content/basic_usage/#interacting-with-the-environment)

### Code example

```shell
import gymnasium as gym
env = gym.make("LunarLander-v2", render_mode="human")
observation, info = env.reset()

for _ in range(1000):
    action = env.action_space.sample()  # agent policy that uses the observation and info
    observation, reward, terminated, truncated, info = env.step(action)

    if terminated or truncated:
        observation, info = env.reset()

env.close()
```

### System info

- gymnasium version 0.29.1 installed via pip for Python 3.11
- OS: Ubuntu 22.04.2
- Python 3.11

### Additional context

Code example taken from official documentation, [here](https://gymnasium.farama.org/content/basic_usage/#interacting-with-the-environment)

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
Thanks for pointing that out, I will see if we can update the basic usage page to make this more clear

smarianimore:
feel free to ask for a pull request if that's the intended behaviour and thus updating the doc pages is all that's needed

PR #740, opened 2023-10-13
Add more introductory pages (and update "create custom environment")

pseudo-rnd-thoughts:
# Description

In reviewing the website, I realised that we currently have a pretty bad set of introductory pages for new users. 
Therefore, this PR adds several key pages that I believe will be very helpful for new users:
* basic usage - Same as before just updated and addresses https://github.com/Farama-Foundation/Gymnasium/issues/735
* training an agent - Most users will be interested in using a built-in gymnasium environment to train a basic RL agents to begin with. The easiest solution is to move the "Solving blackjack using Q-learning" tutorial with minor changes. We should look to add tutorials for training agents to play Atari and Mujoco environments.
* create a custom environment - The second most common thing for users should be creating their own custom environment. The easiest solution again is to move the "Creating a custom environment" tutorial but this tutorial needs large scale changes (@kir0ul is working on this). We should look to add a section at the end linking to third-party environments. 
* Recording agents - While uncommon, I think having a page about recording data on agents during training and evaluation is helpful. In particular how to use the `RecordVideo` and `RecordEpisodeStatistics`. 
* Speeding up environments - A bit more niche but critical is using vector environments. I think "speeding up environment" is more helpful to new users than "how to use vector environments" and we can note other techniques that can be used. 
* Compatibility with OpenAI Gym - Same as before (show be updated for v1.0.0)
* Migration guide - Same as before but should include updating to v1.0.0 

Todo
- [ ] Add compiling from python to rst (we support this for tutorials but we don't want the gallery aspect)
- [ ] Finish "Recording agents"
- [ ] Modify "Speeding up the environment" to add an explanation of vector environments
- [ ] Update training an agent to be more introductory


pseudo-rnd-thoughts:
closed for #791

Issue #786, opened 2023-11-22
[Bug Report] Extra kwargs in register don't work


younik:
### Describe the bug

I am trying to register an env defining env-specific kwargs. However, this doesn't work, and I get the following error:

```bash
venv/lib/python3.11/site-packages/gymnasium/envs/registration.py:677: in register
    new_spec = EnvSpec(
E   TypeError: EnvSpec.__init__() got an unexpected keyword argument 'num_envs'
```

I believe the problem is the following line:
https://github.com/Farama-Foundation/Gymnasium/blob/b312a0334e28ed69ca1fa2bcffbb8ab239a9b448/gymnasium/envs/registration.py#L624

That we should change like this:
```diff
- **kwargs,
+ kwargs=kwargs,
```

### Code example

```shell
In file:
class MyEnv(gym.Env):
    def __init__(self, num_envs: int):
        pass


In init:

register(
    id="VecBreedGym",
    entry_point="file:MyEnv",
    num_envs=1,
)
```

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
Yes, your right. I wonder how long this has been an issue for? 
@younik Are you happy to make a PR?

younik:
> Yes, your right. I wonder how long this has been an issue for?
> @younik Are you happy to make a PR?

Sure, I opened a PR.

I found from tests, that the way to go before was:

```python
gym.register(
   ...
   kwargs={"arg1": value1}
```

Instead of:

```python
gym.register(
   ...
   arg1=value1
```

Also the past behavior is fine, but then we don't want this signature
https://github.com/Farama-Foundation/Gymnasium/blob/b312a0334e28ed69ca1fa2bcffbb8ab239a9b448/gymnasium/envs/registration.py#L570

But instead simply kwargs.


I think the new implementation is better, however, it is a breaking change. So, maybe we can simply change the signature of the register, and keep the current behavior (specifying kwargs of env as a dict in register)

younik:
Okay, in the end, I did the non-breaking change, so you can register the environment using the named argument kwargs passing a dictionary of arguments for the env.

pseudo-rnd-thoughts:
[...]

PR #788, opened 2023-11-23
Fix type for `register(kwargs)`

younik:
Fixes https://github.com/Farama-Foundation/Gymnasium/issues/786

### DIFF
--- a/gymnasium/envs/registration.py
+++ b/gymnasium/envs/registration.py
@@ -567,7 +567,7 @@ def register(
     disable_env_checker: bool = False,
     additional_wrappers: tuple[WrapperSpec, ...] = (),
     vector_entry_point: VectorEnvCreator | str | None = None,
-    kwargs: dict = {},
+    kwargs: dict | None = None,
 ):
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Could you add a test to confirm that this is working and so we don't mess this up again

gymnasium/envs/registration.py
```diff
@@ -567,7 +567,7 @@ def register(
     disable_env_checker: bool = False,
     additional_wrappers: tuple[WrapperSpec, ...] = (),
     vector_entry_point: VectorEnvCreator | str | None = None,
-    **kwargs: Any,
+    kwargs: dict = {},
```
pseudo-rnd-thoughts, line 570: I know that pycharm shouts at me if I do this as `{}` is mutable object. 
Could you change to `dict | None = None`

younik: True, done!

pseudo-rnd-thoughts (approved):

PR #789, opened 2023-11-24
Add support for `JaxToNumpy` to handle NamedTuples (fixes #780)

### LINKED ISSUE #780: [Bug Report] TypeError: <NamedTuple>.__new__() missing 1 required positional argument: '<argument>'
### Describe the bug

wrapper conversion jax_to_numpy fails for NamedTuples in 

As there are no specific handling of NamedTuples installed, the Iterable will be used

```
@jax_to_numpy.register(abc.Iterable)
def _iterable_jax_to_numpy(
    value: Iterable[np.ndarray | Any],
) -> Iterable[jax.Array | Any]:
    """Converts an Iterable from Numpy arrays to an iterable of Jax Array."""
    return type(value)(jax_to_numpy(v) for v in value)
```

This tries to use positional creation of NamedTuple, will fail as you should use NamedTupleClass._make(Iterable)
I think it should look something like this

```
@jax_to_numpy.register(NamedTuple)
def _namedtuple_jax_to_numpy(
    value: NamedTuple,
) -> NamedTuple:
    """Converts an NamedTuple from Numpy arrays to a NamedTuple of Jax Array."""
    return type(value)._make(jax_to_numpy.jax_to_numpy(v) for v in value)
``` 

guess there should be a numpy_to_jax too?  

### Code example

```shell
from typing import NamedTuple

import jax
from gymnasium.experimental.wrappers.jax_to_numpy import jax_to_numpy, numpy_to_jax


class Observe(NamedTuple):
    board: jax.Array
    player: jax.Array


a_jax = Observe(board=jax.numpy.zeros(shape=(8, 8)),
                player=jax.numpy.zeros(8))

b_np = jax_to_numpy(a_jax)
c_jax = numpy_to_jax(b_np)
```

### System info

pip install gymnasium
Gymnasium 0.29.1
Running on Windows 11 Pro (should not matter)
Python 3.11

### Additional context

_No response_

### Checklist

RogerJL:
They caused "missing N required positional argument: 'name', ..." in _iterable_numpy_to_jax or _iterable_jax_to_numpy as the for loop resulted in a generator that NamedTuples could not handle.

NamedTuple is not a useful base class, every used named type needs to be registered individually. jax_to_numpy.register_namedtuple(NewNamedTuple)

[Bug Report] TypeError: <NamedTuple>.__new__() missing 1 required positional argument: '<argument>'

# Description

Fixes # (780)

## Type of change

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up) - did changes that looks Wrong

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -79,6 +79,11 @@ authors:
     email: omar.younis98@gmail.com
     affiliation: University of Bologna
     orcid: 'https://orcid.org/0009-0004-2783-7932'
+  - given-names: Roger
+    family-names: Larsson
+    email: roger.larsson@slentelit.se
+    affiliation: Slentelit AB
+    orcid: 'https://orcid.org/0009-0004-7872-3142'
 identifiers:
   - type: doi
     value: 10.5281/zenodo.8127025
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -79,11 +79,6 @@ authors:
     email: omar.younis98@gmail.com
     affiliation: University of Bologna
     orcid: 'https://orcid.org/0009-0004-2783-7932'
-  - given-names: Roger
-    family-names: Larsson
-    email: roger.larsson@slentelit.se
-    affiliation: Slentelit AB
-    orcid: 'https://orcid.org/0009-0004-7872-3142'
 identifiers:
   - type: doi
     value: 10.5281/zenodo.8127025
--- a/gymnasium/wrappers/jax_to_numpy.py
+++ b/gymnasium/wrappers/jax_to_numpy.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import functools
 import numbers
 from collections import abc
-from typing import Any, Iterable, Mapping, NamedTuple, SupportsFloat, Type
+from typing import Any, Iterable, Mapping, NamedTuple, SupportsFloat
 
 import numpy as np
 
@@ -71,6 +71,7 @@ def _namedtuple_numpy_to_jax(
 ) -> NamedTuple:
     """Converts an NamedTuple from Numpy to Jax."""
     # "To prevent conflicts with field names, the method and attribute names start with an underscore."
+    # noinspection PyProtectedMember
     return type(value)._make(numpy_to_jax(v) for v in value)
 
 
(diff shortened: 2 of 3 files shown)


RogerJL:
Feels like register_namedtuple() should be added to documentation at a higher level, but I can not find out where...

pseudo-rnd-thoughts (changes requested):
Thanks Roger for looking into this.
I was surprised that you would need to register the NamedTuple so I had a look into it
I found a StackOverflow that discussed a solution to this with this proposed solution 

```python
@numpy_to_jax.register(abc.Iterable)
def _iterable_numpy_to_jax(
    value: Iterable[np.ndarray | Any],
) -> Iterable[jax.Array | Any]:
    """Converts an Iterable from Numpy Arrays to an iterable of Jax Array."""
    if hasattr(value, "_make"):
        return type(value)._make(numpy_to_jax(v) for v in value)
    else:
        return type(value)(numpy_to_jax(v) for v in value)
```
and 
```python
@jax_to_numpy.register(abc.Iterable)
def _iterable_jax_to_numpy(
    value: Iterable[np.ndarray | Any],
) -> Iterable[jax.Array | Any]:
    """Converts an Iterable from Numpy arrays to an iterable of Jax Array."""
    if hasattr(value, "_make"):
        return type(value)._make(jax_to_numpy(v) for v in value)
    else:
        return type(value)(jax_to_numpy(v) for v in value)
```

For testing, I had 
```python
class TestingNamedTuple(NamedTuple):
    a: np.ndarray
    b: np.ndarray
```

with adding this in the list of parameterized options
```python
        (TestingNamedTuple(np.array([1, 2], dtype=np.int32), np.array([1.0, 2.0], dtype=np.float32)),
         TestingNamedTuple(np.array([1, 2], dtype=np.int32), np.array([1.0, 2.0], dtype=np.float32)))
```

Could you make these changes and remove all of the rest

CITATION.cff
```diff
@@ -79,6 +79,11 @@ authors:
     email: omar.younis98@gmail.com
     affiliation: University of Bologna
     orcid: 'https://orcid.org/0009-0004-2783-7932'
+  - given-names: Roger
```
pseudo-rnd-thoughts, line 82: For the paper here, we have only included authors that have provide significant API changes / support over the years so could you remove this please

RogerJL: Done, and pushed

pseudo-rnd-thoughts (approved):
Thanks @RogerJL for the changes, all good

PR #805, opened 2023-11-30
[DRAFT] Add functionality of FireReset to AtariPreprocessing wrapper

### LINKED ISSUE #781: [Proposal] Add the functionality of FireResetEnv to the AtariPreprocessing wrapper
### Proposal

I am porting an example of DQN on Atari Breakout from Gym/Stable-Baselines3 to Gymnasium, and unfortunately there is no equivalent to the [Stable-Baselines3's FireResetEnv wrapper](https://stable-baselines3.readthedocs.io/en/master/common/atari_wrappers.html#stable_baselines3.common.atari_wrappers.FireResetEnv) in Gymnasium. The very good [AtariPreprocessing wrapper](https://gymnasium.farama.org/api/wrappers/misc_wrappers/#gymnasium.wrappers.AtariPreprocessing) seems to handle all the other functionality present in [Stable-Baselines3's Atari wrappers](https://stable-baselines3.readthedocs.io/en/master/common/atari_wrappers.html), but not this.

I propose building this into the existing AtariPreprocessing as an optional feature, and if there is interest, I am more than happy to do this myself and submit a pull request :)

### Motivation

Breakout requires an agent to use **FIRE** to launch the next "ball" after losing a life or losing the game.
Without this feature, the agent can get stuck if it hasn't learned to press **FIRE** after losing a life (or losing the game), and the **FIRE** action otherwise does nothing.
I am not sure if other Atari environments also have this problem, or only Breakout.

I found [this long discussion](https://github.com/openai/baselines/issues/240) about the functionality in the gym repo. Regardless, I don't understand the reason this wasn't ported, or (apparently) ported and then later removed? If it's default behavior in SB3, there are certainly people who would be interested in using it in Gymnasium.

### Pitch

I would propose adding this functionality directly to the AtariPreprocessing wrapper, with a corresponding argument to switch the behaviour on and off. The default behaviour would remain the same. The wrapper would monitor the return value of `lives()` from the ALE (as it already does for `terminal_on_life_loss`), and apply an action with `self.env.step(1)` if lives have decreased. For reference, here is the [source code of the FireResetEnv wrapper in Stable-Baselines3](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/atari_wrappers.html#FireResetEnv).

This implementation would be independent of the episodic life behavior (i.e. you have FireReset without episode life, or vice versa, or both features together).

### Alternatives

One alternative is to directly use Stable-Baselines3's FireResetEnv wrapper, but this only works if also using the EpisodicLifeEnv wrapper. The fire reset functionality only triggers on an environment reset, not on a loss of life.
Maybe it's also possible to add the **FIRE** action to the environment's `getStartingActions`, but this is probably messy and backwards-incompatible.

### Additional context

_No response_

### Checklist

balazsgyenes:
# Description

Breakout requires an agent to use **FIRE** to launch the next "ball" after losing a life or losing the game.
Without this feature, the agent can get stuck if it hasn't learned to press **FIRE** after losing a life (or losing the game), and the **FIRE** action otherwise does nothing.

This PR integrates the functionality of [Stable-Baselines3's FireResetEnv wrapper](https://stable-baselines3.readthedocs.io/en/master/common/atari_wrappers.html#stable_baselines3.common.atari_wrappers.FireResetEnv) into Gymnasium's [AtariPreprocessing wrapper](https://gymnasium.farama.org/api/wrappers/misc_wrappers/#gymnasium.wrappers.AtariPreprocessing).

Fixes #781 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/atari_preprocessing.py
+++ b/gymnasium/wrappers/atari_preprocessing.py
@@ -1,7 +1,7 @@
 """Implementation of Atari 2600 Preprocessing following the guidelines of Machado et al., 2018."""
 from __future__ import annotations
 
-from typing import Any, SupportsFloat
+from typing import Any, Literal, SupportsFloat
 
 import numpy as np
 
@@ -27,8 +27,10 @@ class AtariPreprocessing(gym.Wrapper, gym.utils.RecordConstructorArgs):
     - Noop Reset: Obtains the initial state by taking a random number of no-ops on reset, default max 30 no-ops.
     - Frame skipping: The number of frames skipped between steps, 4 by default.
     - Max-pooling: Pools over the most recent two observations from the frame skips.
-    - Termination signal when a life is lost: When the agent losses a life during the environment, then the environment is terminated.
+    - Termination signal when a life is lost: When the agent loses a life during the environment, then the environment is terminated.
         Turned off by default. Not recommended by Machado et al. (2018).
+    - Fire after life is lost: executes a FIRE action on reset or when a life is lost, for environments that are fixed until firing.
+        Turned off by default.
     - Resize to a square image: Resizes the atari environment original observation shape from 210x180 to 84x84 by default.
     - Grayscale observation: Makes the observation greyscale, enabled by default.
     - Grayscale new axis: Extends the last channel of the observation such that the image is 3-dimensional, not enabled by default.
@@ -50,6 +52,7 @@ class AtariPreprocessing(gym.Wrapper, gym.utils.RecordConstructorArgs):
         frame_skip: int = 4,
         screen_size: int = 84,
         terminal_on_life_loss: bool = False,
+        fire_after_life_loss: bool | Literal["auto"] = False,
         grayscale_obs: bool = True,
         grayscale_newaxis: bool = False,
         scale_obs: bool = False,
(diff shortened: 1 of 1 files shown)


balazsgyenes:
Note: I'm not quite done with this feature yet, and have not tested and fully documented the changes. Sorry for the delay, I will do it as soon as I can. In the meantime, feedback is very much appreciated.

In addition, I noticed another potential bug that I wanted to ask about before addressing. The wrapper currently takes the max over the last two observations of the N frame skip frames. However, if the wrapped environment is done (either due to truncated or terminated), no further frames are saved, and the `obs_buffer` can be in any state (2 old observations from the previous step, 1 new / 1 old, 2 new observations). This has probably never caused any issues because Atari environments are typically not trained with time limits (I think?), and the final observation is ignored anyway if `terminated=True`. However, it is incorrect if `truncated=True`.

A potential fix would use a `deque` object from the collections library to save every observation, with only the last 2 observations retained. This would likely come with a performance penalty, since e.g. for `frame_skip=4`, half the observations are not even generated. @pseudo-rnd-thoughts, do you think this bug should be addressed?

pseudo-rnd-thoughts:
Reading through https://github.com/openai/baselines/issues/240 then I believe that `FireReset` is not necessary as a feature as it is implemented in `ale-py` automatically. Therefore, SB3 and CleanRL etc who use it should be able to remove it from the environment stack

Issue #594, opened 2023-07-09
[Question] Possible ways to implement "exclusive dictionary" action spaces


SiyuanQi:
### Question

In Gymnasium we can define an action space with two possible actions like below:
```python
box = gymnasium.spaces.Box(0,1)
action_space = gymnasium.spaces.Dict({ 'a' : box, 'b' : box })
```
where both action `a` and `b` would be executed. 

I wonder if there's a way in Gymnasium to implement an "exclusive dictionary" or "discrete dictionary", so that the agent can only choose from either `a` or `b` to perform?
In a more general case, `a` and `b` can be different composite spaces (and possibly "exclusive dictionaries"), which makes the space hierarchical.

pseudo-rnd-thoughts:
This is an interesting idea, we considered adding this more for PettingZoo where you could have a variable number of agents acting in a timestep. 
This is easy to implement, I'm just uncertain if this is common enough for us to add it
@RedTachyon Do you have any thoughts?

SiyuanQi:
> This is an interesting idea, we considered adding this more for PettingZoo where you could have a variable number of agents acting in a timestep. This is easy to implement, I'm just uncertain if this is common enough for us to add it @RedTachyon Do you have any thoughts?

I think it might be easy to have this space, but how to make an appropriate flatten version of that would be something to think about.
I feel there are a lot of use cases for that, basically when an agent can perform only one action at a time but with different potential parameters for each possible action.

RedTachyon:
I'm cautiously positive about this, and I feel like this might have a solid mathematical foundation somewhere behind it, much like algebraic data types. We already have product types (tuples and dicts), this would be a sum type. A practical-ish example of a sum space would be something like:
```python
Union(
    Discrete(num_keys),
    Box(..., size=(2,))
)
```
which roughly represents what you can do with a keyboard and mouse - press a button or move the mouse (ignoring some domain-specific nuances like the ability to press multiple buttons at once)

To have it a bit more grounded, could you share your specific application? Like where would this actually be useful?

Flattening probably would be fine the same way as for Tuple spaces, it can't be bijective anyways

pseudo-rnd-thoughts:
My only issue is the name, `Union`, generally implies only one of some options is selected where I understood `exclusive` to mean a couple of the options can be used at the same time. I proposed `Container` previously but I don't think it is a much better name

RedTachyon:
[...]

PR #812, opened 2023-12-04
Add a OneOf space for exclusive unions

RedTachyon:
See #594 

This is more of an exclusive tuple, but effectively the same thing. 

I opted for the name OneOf as opposed to Union, sacrificing some math-notation-niceness, but making it more immediately obvious what it is. Plus I think GRPC uses `oneof` similarly.

Some design choices might be controversial:
- A sample in this space has to contain the index of the subspace, as well as the proper sample. This is to disambiguate spaces like `OneOf(Discrete(5), Discrete(5))`
  - This makes mathematical sense - elements of a [disjoint union](https://en.wikipedia.org/wiki/Disjoint_union) of sets have to contain the index, and this is the best mathematical model for what we're doing
  - Supporting repeated subspaces is imo very important, otherwise cases like `OneOf(Discrete(n), Discrete(m))` need special treatment when `n==m`, and nobody likes that
- It flattens to a box of size `max_size + 1`, where `max_size` is the maximum size of any subspace. The 1 is the index, and it comes at the beginning of the representation
- Flattened samples are filled with `nan`s if they don't use up all the space
  - Because of this, a `Box` can now be nullable, which is used when checking belonging to a space

Overall I think this is the best combination of questionable choices, but I'm not super attached to them if someone has better options.

I also think this is useful to actually include for two reasons:
- Practical - it enables things like "move a joystick or press a key"
- Mathematical - with this, we're kinda emulating type theory with product (Tuple) and sum (OneOf) types. Maybe we'll see a math nerd use it for some cool theoretical results.


pseudo-rnd-thoughts:
Spending some time thinking about this new space, I'm not sure about adding it. 
This is where I would like us to a `gymnasium-contrib` for random extra features that should not be core but could be helpful to use users. 
As this space seems to me very rare in use and for the environment is makes it difficult to use as it can be unclear what option was taken. 

Therefore, I would currently be in favour of closing this PR and not merging but if I have missed something, please say

RedTachyon:
> for the environment is makes it difficult to use as it can be unclear what option was taken

I don't think that's true? By design, it includes the index of the space. So in a space `OneOf(Discrete(5), Box(-1, 1, (2,)), Discrete(3))`, samples from each respective subspace would be `(0, 3)`, `(1, [0.3, -0.2])`, `(2, 1)`, with no ambiguity

> As this space seems to me very rare in use

There are two things here. Is it currently rarely in use? Yes, because it doesn't exist. People will naturally gravitate to things that exist and are supported.

If we add it, will it still be rarely in use? Probably to some extent, same as the other "weird" spaces like graph or text. But I think it's likely that it will see some interesting use - certain video games (stacraft), desktop environments, maybe even some industrial simulations. It's basically the generic version of "You can do A or B", with A or B having some internal structure. 

Finally, it really makes sense in terms of theory. Ask any programming language theory nerd, they'll tell you that algebraic datatypes are the best thing ever. Every time you write `TypeA | TypeB`, i.e. a sum type, that's a OneOf space. Pattern matching that was recently introduced in Python, is basically a utility for sum types (if it's a string, do this, and if it's a float, do that). 

You could get more creative and use it for observations - resizeable observations, custom in-environment error handling, active perception (do you want to read some in-game text, or look around?)

My point with research tooling is always that it should be as flexible as reasonably possible. I don't think adding this space has significant downsides or that much maintenance overhead, but it adds a ton of flexibility, and allows expressing things that are very natural to think about.

A reasonable contrib-y middle ground might be a new `experimental`? Then over time, we'd either let it chill there, fizzle out, or merge it in the main code if people use it. But tbh I'd still prefer to just add it as a "regular" supported space, it just makes too much sense.

daniel-redder:
Hello, I am a member of The University of Georgia's THINC Lab, which focuses on RL, IRL, and Robotics.  (see article, [https://onlinelibrary.wiley.com/doi/full/10.1002/aaai.12131](AI Mag OASYS)) discusses Open Task environments. 

These are environments where groups of states (and associated actions) change dynamically in the environment. An example we reference is a RideShare domain

image

Where each passenger constitutes a task with actions associated to each. 

A use case for the OneOf space could be to handle making non-task-specific actions when handling the dynamic nature of an open-task environment. A good example of this type of global action is NOOP. 

```
d = Dict(
    {
        "taskID": Discrete(len(tasks)),
        "taskActions": Discrete(len(B)),
    }
    
)

c = Discrete(2) #NOOP
```

gymnasium/spaces/utils.py
```diff
@@ -525,3 +558,21 @@ def _flatten_space_text(space: Text) -> Box:
...
+    highs = np.array([np.max(flatten_space(s).high) for s in space.spaces])
+
+    overall_low = np.min(lows)
+    overall_high = np.max(highs)
+
+    low = np.concatenate([[0], np.full(max_flatdim - 1, overall_low)])
+    high = np.concatenate([[num_subspaces - 1], np.full(max_flatdim - 1, overall_high)])
+
+    dtype = np.result_type(*[s.dtype for s in space.spaces if hasattr(s, "dtype")])
+    return Box(low=low, high=high, shape=(max_flatdim,), dtype=dtype, nullable=True)
```
pseudo-rnd-thoughts, line 578: Remove `nullable`

gymnasium/vector/utils/space_utils.py
```diff
@@ -121,6 +122,11 @@ def _batch_space_dict(space: Dict, n: int = 1):
     )
 
 
+@batch_space.register(OneOf)
```
pseudo-rnd-thoughts, line 125: Remove and add to `_batch_space_custom` as we have a known number of batch spaces, whereas `Sequence` is unknown

pseudo-rnd-thoughts (changes requested):
Reading through the code, I understand much more what is happening. 
With a couple of changes, I think we can get this merged quickly. 

The primary change I made was to `flatten` such that the fill value is the `flat_sample`'s first value removing the need for `nan` support. As a result, `nullible` 
can be removed from `Box`. 

I added comments to the rest of the PR of proposed changes (along with removing `nullible` from `Box`)

Finally, you need to add the following to `gymnasium/wrappers/utils` 
```python
@create_zero_array.register(OneOf)
def _create_one_of_zero_array(space: OneOf):
    return 0, create_zero_array(space.spaces[0])
```

gymnasium/spaces/utils.py
```diff
@@ -271,7 +271,7 @@ def _flatten_oneof(space: OneOf, x: tuple[int, Any]) -> NDArray[Any]:
     max_flatdim = flatdim(space) - 1  # Don't include the index
     if flat_sample.size < max_flatdim:
         padding = np.full(
-            max_flatdim - flat_sample.size, np.nan, dtype=flat_sample.dtype
+            max_flatdim - flat_sample.size, flat_sample[0], dtype=flat_sample.dtype
```
RedTachyon, line 274: I'm now wondering if it might be better to use `np.empty` instead if we're not using a dedicated placeholder value.

Either way, the array will be filled with some throwaway data, but with empty we're not pretending that this data has any actual meaning.

pseudo-rnd-thoughts: I ran a quick check and this fails `test_flat_space_contains_flat_points` as the flattened version might not be contained within the flattened space. 
But this is just a weird artifact of `flatten` rather than anything actually incorrect with the approach

pseudo-rnd-thoughts (approved):

Issue #106, opened 2022-11-02
[Proposal/Bug Fix] Change truncation to termination in Car Racing after finishing a lap


RedTachyon:
### Proposal

Currently in Car Racing, when the agent finishes a lap, the environment is marked as truncated instead of terminated. This seems like a really odd choice to me.

https://github.com/Farama-Foundation/Gymnasium/blob/cc7d8ddbd43d5a924d79c4ac9bd3b76569f44ceb/gymnasium/envs/box2d/car_racing.py#L557-L561

This was added in openai/gym#2890 alongside an actual fix to the environment logic. I suspect the review focused on the bug fix, and omitted the undiscussed change, so it slipped through the cracks. (BTW now you see why I'm always being annoying about out-of-scope changes in PRs and similar stuff)

Finishing a lap is a very clear example for episode **termination**. You reach a terminal state after making a full loop, and the episode ends. It should never have been marked as truncation. 

The annoying part the environment version was bumped for this (but also for the actual bug), so we'll have to bump it again. But I can't really see any justification for keeping this marked as truncation, which is inconsistent with the entire rationale for what truncation is meant to be (reaching the time limit). The explanation in some of the comments was that finishing the lap shouldn't be treated as a failure, but termination does not imply failure. Failure or success is defined by the reward. Termination says "You're done, nothing more to do". Truncation says "You took too long, try again".

@pseudo-rnd-thoughts @jkterry1 @araffin

Markus28:
I don't think it's an MDP if you change it to be the done signal since the observation doesn't really carry any information about whether a lap has been completed/how much of a lap has been completed. So making it the done signal may be problematic, at least that would be my understanding.

RedTachyon:
> I don't think it's an MDP if you change it to be a termination

I don't really buy this. The observation is a pixel rendering in a radius around the environment, so it's hardly an MDP in the first place.

PR #813, opened 2023-12-05
Change end-of-episode in CarRacing to termination as opposed to truncation

RedTachyon:
There's been some heated debate about this, I'm still open to have my mind changed, but imo with the new termination-truncation split, this is a clear example of a termination as opposed to truncation.

Truncation is meant to represent hitting the "externally imposed" time limit as opposed to actually reaching a terminal state. When the car does the laps that it needs, the episode should be terminated.

To give some more information, I also added an info entry describing the reason for termination (whether or not the lap was finished)

This also requires a version bump, since implementations that correctly handle termination/truncation may converge to different policies


pseudo-rnd-thoughts:
There are a couple of options which we need to plot the training curves for
1. Do nothing, keep the buggy truncation case
2. Fix the truncation testing to check that the agent truncates correctly after 2 (or more) laps (add the truncation lap number as an environment parameter, for later) - https://pypi.org/project/ufal.pybox2d/
3. Change to termination after a single lap

Kallinteris-Andreas:
For reference, this is the initial discussion: https://github.com/Farama-Foundation/Gymnasium/issues/106

Here is my analysis, giving a truncation signal after finishing the LAP would imply that the environment is not over, that there would be more LAPs left to complete, while giving a termination signal would indicate that nothing matters after completing the LAP, for example crushing the car afterwords would be totally fine (realistically that is not going to happen). 

Learning performance wise, I suspect both options would be really close regardless.

pseudo-rnd-thoughts:
@RedTachyon Any chance you can run some training plots for this to measure the change? 
Can you update the changelog for the environment, otherwise good with me

RedTachyon:
Re: changelong - do you mean in the docstring? That's already done. The docs site page (I think) gets automatically generated from that.

Re: experiments - probably too much effort for the benefit. The impact would likely be visible in the qualitative behavior of the agent towards the very end of the race, and that's contingent on the implementation correctly accounting for the termination/truncation split

pseudo-rnd-thoughts (approved):

PR #818, opened 2023-12-06
Change the functional API to include explicit params

RedTachyon:
See: https://discord.com/channels/961771112864313344/1029089576398114846/1165900249798287391

This introduces an explicit params object to customize envs within a jitted context. Still sorta WiP, but I want to see full CI


pseudo-rnd-thoughts (approved):
Why did we remove the example code from CartPole etc otherwise good to merge

PR #830, opened 2023-12-09
Add `padding_type` parameter to `FrameSkipObservation` and change default to "reset" to preserve backwards comp

### LINKED ISSUE #829: [Question] Does FrameStackObservation work as expected?
This is the current code of the reset function:

https://github.com/Farama-Foundation/Gymnasium/blob/ede2ed1af2146bd9d3a3129418eb929636c78d96/gymnasium/wrappers/stateful_observation.py#L396-L399

Should the code no be like this:
```python

        obs, info = super().reset(seed=seed, options=options)
        for _ in range(self.stack_size):
            self._stacked_obs.append(obs)  # <--- change here!
        
```

jamartinh:
# Description

FrameStacking as initially (from long time ago) designed filled the buffer with the initial observation from env.reset, I.e. repeated many times as the stacking number.

In the re-implementation the assumption changed to fill it with zeroes, but this not only breaks many codes, it breaks "what is expected" and its proper documentation design.

I have fixed this by using its proper definition:

```
Note:
        - After :meth:`reset` is called, the frame buffer will be filled with the initial observation.
          I.e. the observation returned by :meth:`reset` will consist of `num_stack` many identical frames.

```

Fixes #829

Also, I have modified the call to super() with self.env as this is the right way for working with wrappers. 
If super() used then unexpected errors can arise when using multiple wrappers.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

### Screenshots

Before:
```
Original env:
[ 1.25 -0.   -0.   -0.    0.   -0.    0.    0.   -0.   -0.    0.  ]

Stacked env:
[[ 0.    0.    0.    0.    0.    0.    0.    0.    0.    0.    0.  ]
 [ 0.    0.    0.    0.    0.    0.    0.    0.    0.    0.    0.  ]
 [ 0.    0.    0.    0.    0.    0.    0.    0.    0.    0.    0.  ]
 [ 0.    0.    0.    0.    0.    0.    0.    0.    0.    0.    0.  ]
 [ 1.25 -0.    0.    0.    0.    0.    0.   -0.   -0.    0.    0.  ]]


```

After:
```
Original env:
[ 1.25  0.   -0.   -0.    0.   -0.    0.   -0.   -0.   -0.    0.  ]

Stacked env:
[[ 1.25 -0.   -0.   -0.    0.    0.    0.    0.    0.   -0.   -0.  ]
 [ 1.25 -0.   -0.   -0.    0.    0.    0.    0.    0.   -0.   -0.  ]
 [ 1.25 -0.   -0.   -0.    0.    0.    0.    0.    0.   -0.   -0.  ]
 [ 1.25 -0.   -0.   -0.    0.    0.    0.    0.    0.   -0.   -0.  ]
 [ 1.25 -0.   -0.   -0.    0.    0.    0.    0.    0.   -0.   -0.  ]]

```

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gymnasium/wrappers/stateful_observation.py
+++ b/gymnasium/wrappers/stateful_observation.py
@@ -371,7 +371,7 @@ class FrameStackObservation(
         Returns:
             Stacked observations, reward, terminated, truncated, and info from the environment
         """
-        obs, reward, terminated, truncated, info = super().step(action)
+        obs, reward, terminated, truncated, info = self.env.step(action)
         self._stacked_obs.append(obs)
 
         updated_obs = deepcopy(
@@ -393,11 +393,10 @@ class FrameStackObservation(
         Returns:
             The stacked observations and info
         """
-        obs, info = super().reset(seed=seed, options=options)
-        for _ in range(self.stack_size - 1):
-            self._stacked_obs.append(self.zero_obs)
-        self._stacked_obs.append(obs)
-
+        obs, info = self.env.reset(seed=seed, options=options)
+        self._stacked_obs = deque(
+            [obs for _ in range(self.stack_size)], maxlen=self.stack_size
+        )
         updated_obs = deepcopy(
             concatenate(
                 self.env.observation_space, self._stacked_obs, self._stacked_array
--- a/gymnasium/wrappers/stateful_observation.py
+++ b/gymnasium/wrappers/stateful_observation.py
@@ -302,38 +302,80 @@ class FrameStackObservation(
     No vector version of the wrapper exists.
 
     Note:
-        - After :meth:`reset` is called, the frame buffer will be filled with the initial observation.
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for pointing this out. 
I implemented with zero as the padding observation due to this is the standard atari implementation / usage. 
However I agree that this breaks the prior implementation.

Therefore, I would propose the user having the ability to select the pad observation in `__init__` with the options being "zero", "initial", and custom

pseudo-rnd-thoughts:
@jamartinh I have made some changes to your implementation, before I merge, are you happy with the changes and new tests

jamartinh:
Good, thanks!

pseudo-rnd-thoughts (approved):

PR #844, opened 2023-12-14
fix mujoco_v5 observation_space constructor typing issue

Kallinteris-Andreas:
# Description

fixes a type hint oversight

## Type of change

- [X] Type hint only change

# Checklist:


pseudo-rnd-thoughts (commented):
In what case, can't the users pass the observation space to `MujocoEnv`?
I'm just alarmed that this could cause users to accidently not have an observation space for their environment

Kallinteris-Andreas:
Ok, let me elaborate on what happens,

with MuJoCo-v4 and older environments, the observation space was predefined in the environment class

but with MuJoCo-v5 the observation space is set based on the model used (which may be different from the default because of the `xml_file` argument), which in computed after the model is loaded
https://github.com/Farama-Foundation/Gymnasium/blob/219d1242e8eade5385d2ff909e9d7fe2028375c5/gymnasium/envs/mujoco/ant_v5.py#L303-L327

pseudo-rnd-thoughts (approved):
Fair, I guess that if the observation space is still none then the `PassiveEnvChecker` should catch this for users

PR #856, opened 2023-12-22
Remove `pyright` version limit for pre-commit

Kallinteris-Andreas:
# Description

introduces a couple of errors in `registration.py`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,6 +65,6 @@ repos:
         language: node
         pass_filenames: false
         types: [python]
-        additional_dependencies: ["pyright@1.1.342"]
+        additional_dependencies: ["pyright"]
         args:
           - --project=pyproject.toml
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for updating this, I had forgotten about this limitation

I'm surprised by the type changes required as `wrappers` and `vector_kwargs` should not be able to `None`. 
@Kallinteris-Andreas I have removed the dependency of pyright so it should autoupdate over time without the requirement of further PRs.  
However this can cause new type issues possibly in the future so we will have to see

Kallinteris-Andreas:
Is it a good idea to have `pyright` not pinned, but everything else in `pre-commit` pinned?

pseudo-rnd-thoughts:
I was considering this, pyright is updated on a very regular basis with bug fixes (which we would prefer to have) while the other pre-commit modules are not so regular or as important. I'm happy to revert if you think this is unwise

PR #863, opened 2023-12-28
Change space and force calculation

adrianmay:
# Description

The goal is to endow CartPole with a continuous-valued action. It seems odd that it was ever constrained to give full gas all the time. A control system should try to minimise its use of fuel but this policy renders that impossible. Also, this digitisation presents neural networks with a problem that's irrelevant and obfuscating to most research goals. 

I notice that two environments already have continuous and digitised action variants, but in two different styles, and it's hard to see what the digitsed variants offer that isn't a strict subset of what the continuous ones do, begging the question why the project would want the digitised variants hanging around.
 
The current patch simply switches the input space to a 1D box ranging from 0 to 1 and interprets the value such that 0 and 1 are interpreted the same as before: full gas to the left or right respectively. To apply no force, one must now send 0.5. Existing client code would have suffered an exception if it had sent any value besides 0 or 1, so this change is guaranteed not to break existing experiments. The fact that actions can't stray outside of the range 0-1 means that no new extreme observations could have been introduced. 

So while backward compatibility has been preserved, if an experiment wants to use the new continuous action, then it must use a version of Gymnasium subsequent to the present patch.

In this light, you might think the patch can be accepted as is (but for documentation updates and maybe new tests.) On the other hand, you might think it should be called CartPole-v2, or CartPoleContinuous (and in either case perhaps that 0 should mean no force while negative/positive values mean push left/right), or you might reject the whole concept in which case I'd have to direct anybody interested in my stuff to use my cloned repo, or make a new third party environment, both of which would seem like a lot of baggage for such a tiny and non-breaking change.

Please let me know your thoughts.

# Checklist:

### DIFF
--- a/gymnasium/envs/classic_control/cartpole.py
+++ b/gymnasium/envs/classic_control/cartpole.py
@@ -131,7 +131,11 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
             dtype=np.float32,
         )
 
-        self.action_space = spaces.Discrete(2)
+        self.action_space = spaces.Box(
+           np.array([0.0], dtype=np.float32),
+           np.array([1.0], dtype=np.float32),
+           dtype=np.float32)
+
         self.observation_space = spaces.Box(-high, high, dtype=np.float32)
 
         self.render_mode = render_mode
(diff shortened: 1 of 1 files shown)


Kallinteris-Andreas:
https://gymnasium.farama.org/environments/mujoco/inverted_pendulum/ already exist

The implementation if based on the original paper by Sutton and Burton https://ieeexplore.ieee.org/document/6313077

pseudo-rnd-thoughts:
Thanks @Kallinteris-Andreas I was not aware of that. Therefore, closing as this implementation already exists

adrianmay:
Sorry for the radio silence. It's been a busy Christmas and I since figured
out that I have to do some other research before reaching for any control
task. So it'll be a while before I get to pay with the game engine version.
I guess it'll do the trick though. Thanks for your advice, Adrian.

PR #895, opened 2024-01-29
fix cartpole argument examples formating

Kallinteris-Andreas:
previously:
https://gymnasium.farama.org/main/environments/classic_control/cart_pole/
image


pseudo-rnd-thoughts:
Complete in #898

Issue #851, opened 2023-12-20
[Question] `FuncEnv` why is the `key` not argument for all member functions


Kallinteris-Andreas:
### Question

`FuncEnvs`'s
- initial
- transition
do have a `key` argument

but 
- observation
- reward
- terminal
do not have a `key` argument

which could be useful for environments with noisy observation and noisy rewards (while it is less common than initial and transition)

@RedTachyon

RogerJL:
I see no reason for `terminal` to be noisy.

Having `key` in all methods will require generating a new one before **all** calls, and `split` takes time...
My thought was that `param` could be use to pass it for those that are in need, but then I guess `transition`(?) should be able to output a new modified `params`

Issue #861, opened 2023-12-23
[Question] [`FuncEnv`] `params` are not used by `__init__` & `render_init` & `render_close`


Kallinteris-Andreas:
### Question

- `__init__` uses `options` instead of `params`
- `render_init` uses `kwargs` instead of `params`
- `render_close` uses nothing (granted, I do not see a need to for it to include `params`)

RedTachyon:
This is somewhat similar to #851, so I'll address them together.

Basically, there's a tradeoff between being consistent and being efficient. In case of the RNG argument, if a certain function isn't expected to use RNG, we can omit it for brevity, or include it for consistency (and for the edge cases that might use it)

I think in both cases, I tried to be a bit too clever going for efficiency, and it might instead be better to keep it consistent instead - every single method would be expected to have both `rng` and `params`? This seems a bit weird for things like `render_close`, so I'm not sure. But we definitely should have some clear principles, e.g. maybe omit `rng` just for rendering and info-related functions? (which are all supposed to convey some sort of information about the environment "from the outside")

As for the specific questions in this issue - `__init__` options serve a different purpose than `params` everywhere else, so it's good imo to keep them named differently. `render_init` initially had `**kwargs` because it's a thing that's unlikely to be jitted anyways, plus it's hard to predict what users might want to put there, but with `params` introduced, it's probably better to just use that. `render_close` probably doesn't need anything, but we should keep it consistent. And it might need something, e.g. we know if we need to close a pygame window based on whether it was created (based on the params)

Kallinteris-Andreas:
When we are making an Environment API, it should support all environments (other than perhaps very weird edge cases) so:

## rng
- `observation` & `reward` `terminal` should all have an `rng` argument, which, even if rare, some environments may need them
- `transition_info` & `state_info` should not have one, because they are deterministic by nature
- `render_image` & `render_init` `render_close` should not have one, as they should be a deterministic rendering (it would be worrying if an environment needs to present nondeterministic information to the user)
- `transform` & `get_default_params` unless you are psycho, they should be deterministic

## params
- `render_init` `**kwargs` should be replaced with `params`
- `render_close` add `params`

RogerJL:
I have no need of `rng` in `observation`, completely deterministic, and calls to `split` function to generate new are among the slowest in my loop.

state - global state (`transform` could split new `rngs` saved here if needed)

With that `rng` would only be needed to initial

Kallinteris-Andreas:
[...]

PR #900, opened 2024-02-01
Add `params` and `rng` argument to all `FuncEnv` member functions

Kallinteris-Andreas:
# Description

Fixes https://github.com/Farama-Foundation/Gymnasium/issues/851, https://github.com/Farama-Foundation/Gymnasium/issues/861

## Type of change

# Checklist:


tests/envs/functional/test_core.py
```diff
@@ -12,23 +12,25 @@ def __init__(self, options: Optional[Dict[str, Any]] = None):
...
-    def terminal(self, state: np.ndarray) -> bool:
+    def terminal(self, state: np.ndarray, rng: Any) -> bool:
         return state[1] > 0
 
 
 def test_api():
     env = BasicTestEnv()
     state = env.initial(None)
-    obs = env.observation(state)
+    obs = env.observation(state, None)
```
RedTachyon, line 33: Can we make the rng argument optional the same way as params?

RedTachyon (changes requested):
At a glance seems alright, except it would be nice if the `rng` argument was also optional. Can you check if that works well, and if so, add tests for all variants of calling the functions? (implicit None, explicit None, explicit non-trivial arguments)

tests/envs/functional/test_core.py
Kallinteris-Andreas: How should `rng` default be handled?
If an environment's needs `transition` member function requires `rng`, what should be done?

In the case of parameters, there is clearly a default

Kallinteris-Andreas: @RedTachyon

Kallinteris-Andreas:
@RedTachyon check the abobe comment Thanks

Kallinteris-Andreas:
@RedTachyon

RedTachyon:
Sorry about the delay. I was a bit concerned about potential performance implications of this, since jax jit can be a bit moody sometimes with its optimizations, but from a quick test it seems to not cause any issues. We might still have to revisit performance when it comes up, but it's not a priority at the moment, and we'll see how big of a deal it is in "real" environments

So, uh, LGTM

RedTachyon (approved):

PR #913, opened 2024-02-06
pin `mujoco <= 3.1.1`

Kallinteris-Andreas:
I will work with the mujoco devs tomorrow to fix the bug, for now this should work


pseudo-rnd-thoughts:
Are we sure this isn't a bug within MuJoCo?

RogerJL:
Is the below build error caused by this problem?

[Failed build](https://github.com/RogerJL/Gymnasium/actions/runs/7806176036/job/21291981482
)

Kallinteris-Andreas:
@RogerJL yes, you may ignore it

pseudo-rnd-thoughts (approved):
Merging as a immediate patch

PR #922, opened 2024-02-13
refactor BipedalWalker heuristics: extract heuristic in a separate function

OliEfr:
# Description

- refactor the heuristic used for the bipedal walker in a seperate class / function.
- this allows for greater re-usability if one wants to use this function elsewhere, e.g. to generate offline training data as it was the case for me.
- further, there is no explicit reason why the heuristics should be included in the training loop

## Type of change

- [x] Refactor (non-breaking change)

# Checklist:

- [ ] N/A I have commented my code, particularly in hard-to-understand areas
- [ ] N/A I have made corresponding changes to the documentation
- [ ] N/A I have added tests that prove my fix is effective or that my feature works

**Please let me know about any changes you would like to see**

### DIFF
--- a/gymnasium/envs/box2d/bipedal_walker.py
+++ b/gymnasium/envs/box2d/bipedal_walker.py
@@ -782,13 +782,7 @@ class BipedalWalkerHardcore:
         )
 
 
-if __name__ == "__main__":
-    # Heurisic: suboptimal, have no notion of balance.
-    env = BipedalWalker()
-    env.reset()
-    steps = 0
-    total_reward = 0
-    a = np.array([0.0, 0.0, 0.0, 0.0])
+class BipedalWalkerHeuristics:
     STAY_ON_ONE_LEG, PUT_OTHER_DOWN, PUSH_OFF = 1, 2, 3
     SPEED = 0.29  # Will fall forward on higher speed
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, the changes look good

Kallinteris-Andreas:
@OliEfr i know this has been merged, but can you tell me what validation steps were taken for this PR?
can you run [`check_environments_match`](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/utils/env_match.py) for the old and new version (at the same time) for 100k steps?

Thanks!

pseudo-rnd-thoughts:
@Kallinteris-Andreas This doesn't modify the environment. It is just heuristic for acting in the environment and very uncommonly used I believe. Though it would be helpful to confirm that the heuristics achieve the same score given the same seed

OliEfr:
Hi @Kallinteris-Andreas Yes, I didnt touch the environment itself. I also did not change the implementaion for the heuristics. I just moved the code from the environment-interaction-loop to a seperate function. This way the heuristic can be used in other projects and its just more senseful.

As there are more heuristics for other environments, I think it would make sense to refactor all the heuristics in a seperate file for future work.

Issue #921, opened 2024-02-13
[Proposal] Update cython dependency to cython 3


evan0greenup:
### Proposal

Currently, the dependency specifier on `cython` is `cython < 3`.  It would be good to port `gymnasium` to cython 3. 

### Motivation

User whose project need both cython 3 and gymnasium would have problem. And cython 3 fix a lot of bugs.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

Kallinteris-Andreas:
`cython<3` to use the old `mujoco-py` binaries for the `v2` and `v3` of those requirements, if you are not using them you can safely use `cython>=3`

im-Kitsch:
I think we all admit that mujoco-py and mujoco-series "-v2" environments are annoying and not as good as new version. But the problem is that they are still widely used since the research cummunity still uses mujoco-py frequently for online or offline RL.  To validate this, just search RL related paper randomly in this year or check the new issues in mujoco-py repo. To compare with new/old baselines/benchmarks people have to use old "-v2" version.

Moreover, the dependency for cython could be simplely updated by this PR https://github.com/openai/mujoco-py/pull/782. So it would be really really very very appreciated to adjust mujoco-py to cython>3. 

I disklike Mujoco-py, it's really annoying but I think many people still have to use it for a long time.

pseudo-rnd-thoughts:
Mujoco-py is hosted and maintained by OpenAI, not Farama, therefore, there is nothing that we can do to apply the PR you link above. 
We understand your annoyance with v2 and plan to remove them in the future but this isn't a problem and if you don't download `gymnasium[mujoco-py]` then this issue shouldn't effect you

pseudo-rnd-thoughts:
Only "gymnasium[mujoco-py]" is limited to `cython < 3`, therefore, for users needing cython 3 then they cannot use mujoco-py (from pypi) or must build from source with the linked solution. 
Therefore, closing as not a gymnasium issue

PR #926, opened 2024-02-16
mark `mujoco-py` environments as explicity deprecated

Kallinteris-Andreas:
# Description
This PR simply changes the documentation to explicitly state that `mujoco-py` are deprecated

since this [issue](https://github.com/Farama-Foundation/Gymnasium/issues/921), it is clear that the inclusion of the old MuJoCo environments causes issues to user who do not interact with it

I am not sure about when, the removal of `mujoco-py` environments will happen, probably after 1.0, maybe with the release of `py313`?

Fixes # (issue)

## Type of change
- [X] Documentation only change (no code changed)

### DIFF
--- a/docs/environments/mujoco.md
+++ b/docs/environments/mujoco.md
@@ -79,8 +79,8 @@ Gymnasium includes the following versions of the environments:
 | ------- | --------------- | ------------------------------------------------ |
 | `v5`    | `mujoco=>2.3.3` | Recommended (most features, the least bugs)      |
 | `v4`    | `mujoco=>2.1.3` | Maintained for reproducibility                   |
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
I agree, at this point the number of projects using mujoco-py and gymnasium is very low. 

For v1.0.0, I think we keep the environment but after that, we cut them maybe.

pseudo-rnd-thoughts:
It would make sense to move to Gymnasium-robotics I believe, @Kallinteris-Andreas how do you feel about that?

Kallinteris-Andreas:
depends on how compatible it is with `py3.13`

Issue #872, opened 2024-01-11
[Bug Report] Various rendering issues in 'cartpole.py'


RogerJL:
### Describe the bug

1. Accepts any string as 'render_mode' then assumes everything else than 'human' to be "rgb_array"
2. line 512-513: test that 'self.clocks' is None, then sets 'self.clock'...
3. line 525-526: transposes x but does not pick the to the environment corresponding data

### Code example

```shell
__init__


        if self.clocks is None:
            self.clock = [pygame.time.Clock() for _ in range(self.num_envs)]

        for state, screen, clock in zip(self.state, self.screens, self.clocks):
            x = self.state.T
```

### System info

Gymnasium installed from source
version 1.0.0rc1
Windows
Python 3.11

### Additional context

Issues corrected in the HumanVectorRenderingPR4

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

Kallinteris-Andreas:
```py
>>> gymnasium.make("CartPole-v1", render_mode="NON_VALID")
/home/master-andreas/gym/Gymnasium/gymnasium/envs/registration.py:735: UserWarning: WARN: The environment is being initialised with render_mode='NON_VALID' that is not in the possible render_modes (['human', 'rgb_array']).
  logger.warn(
<TimeLimit<OrderEnforcing<PassiveEnvChecker<CartPoleEnv<CartPole-v1>>>>>
>>> gymnasium.make("Ant-v5", render_mode="NON_VALID")
/home/master-andreas/gym/Gymnasium/gymnasium/envs/registration.py:735: UserWarning: WARN: The environment is being initialised with render_mode='NON_VALID' that is not in the possible render_modes (['human', 'rgb_array', 'depth_array']).
  logger.warn(
<TimeLimit<OrderEnforcing<PassiveEnvChecker<AntEnv<Ant-v5>>>>>
```

RogerJL:
OK, fixing

RogerJL:
BTW noticed

registration.py:722
            logger.warn(
                "You are trying to use 'human' rendering for an environment that doesn't natively support it. "
                "The HumanRendering wrapper is being applied to your environment."
            )
Should that really be a warning? Feels like info level to me. Asking for "human", knowing of automatic HumanRendering, getting "human"...

Kallinteris-Andreas:
[...]

PR #937, opened 2024-02-24
CartPole: fix vector rendering (#872)

RogerJL:
Also corrected some bugs in vectorized Cartpole

# Description

This fixes some, probably breaking, issues in Chartpole
1) Testing if clocks is not set, created clock if not, resulting in a new clock array created every turn.
2) Looping over states but assigned x with full all states

Fixes # (872)

## Type of change

# Checklist:

Tests redesigned, will arrive in HumanRenderer for Vector environments

### DIFF
--- a/gymnasium/envs/classic_control/cartpole.py
+++ b/gymnasium/envs/classic_control/cartpole.py
@@ -519,7 +519,7 @@ class CartPoleVectorEnv(VectorEnv):
                     for _ in range(self.num_envs)
                 ]
         if self.clocks is None:
-            self.clock = [pygame.time.Clock() for _ in range(self.num_envs)]
+            self.clocks = [pygame.time.Clock() for _ in range(self.num_envs)]
 
         world_width = self.x_threshold * 2
         scale = self.screen_width / world_width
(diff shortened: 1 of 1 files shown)


gymnasium/envs/classic_control/cartpole.py
```diff
@@ -531,9 +531,9 @@ def render(self):
         if self.state is None:
             return None
 
-        for state, screen, clock in zip(self.state, self.screens, self.clocks):
-            x = self.state.T
-
+        for state, screen, clock, x in zip(
+            self.state, self.screens, self.clocks, self.state.T
```
pseudo-rnd-thoughts, line 535: Why does this need both `state` and `state.T`? This seems extremely weird.
I'm guessing that one of them is incorrect and should be removed

pseudo-rnd-thoughts:
Testing the PR, it doesn't actually allow human rendering for `num_envs>1`; therefore, I think the easiest solution is to remove `human` rendering from vector environments and use the `vector.HumanRendering` wrapper proposed in https://github.com/Farama-Foundation/Gymnasium/pull/880

Issue #927, opened 2024-02-16
[Proposal] Improve `check_env`


Kallinteris-Andreas:
### Proposal

Hey, [one issue](https://github.com/Farama-Foundation/Gymnasium-Robotics/issues/207) in [Gymansium-Robotics](https://github.com/Farama-Foundation/Gymnasium-Robotics) was discovered, which could have been avoided if `check_env` was improved 

# Critique 1
`utils.check_env.data_equivilance` makes `all_close` comparisons, instead of equality comparison, 
Line 57 at:
https://github.com/Farama-Foundation/Gymnasium/blob/806d8afe9358d9ff2b092f4a4d7ccf0b320690a8/gymnasium/utils/env_checker.py#L33-L63

## proposed solution
add an `exact` argument to `data_equivilance`
and it should be used as `True` when `env.spec.nondeterministic=False`.

Note: all `gymnasium` & `gymnasium-robotics` environments pass with `exact=True` 

# Critique 2
It does not test that `env.reset()` properly resets the internal state, it just checks that initial observation is the same

## proposed solution
for now, we can not have a solution that works across environments

# Critique 3
It does test `env.step()`
## proposed solution
add a simple `check_step` test function, which assumes that `reset(seed)` is deterministic (tested before running this function) and asserts that step is deterministic.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

Kallinteris-Andreas:
@pseudo-rnd-thoughts regarding Critique 1
1) is it a good idea to change the default `exact` argument of `data_equivilance` to  `True`?
2) or change the make function call to include `exact=True`  (at least inside `env_checker.py`)?

pseudo-rnd-thoughts:
Sorry I didn't respond when this was originally posted.
I'm worried about using `allequal` rather than `allclose` due to float point numbers particularly for robotics environments.

An alternative solution for `check_env` is to use both `allclose` and `allequal`. 
For a given observation, we first check if `allequal`, if False then run `allclose`, if all close fails then we raise an error about that otherwise we just log to the users the non-equalness as an environment can be deterministic to e-8 but not equivalent. 

I know from talking to the Mujoco devs that due to physics changes / optimisations that mujoco versions are not deterministic and can produce different observations however within e-10 or so I believe.

Kallinteris-Andreas:
[...]

PR #953, opened 2024-03-07
`env_checker` to use exact data_equivilance

Kallinteris-Andreas:
# Description

fixes https://github.com/Farama-Foundation/Gymnasium/issues/927
## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gymnasium/utils/env_checker.py
+++ b/gymnasium/utils/env_checker.py
@@ -99,7 +99,7 @@ def check_reset_seed_determinism(env: gym.Env):
             ), "The observation returned by `env.reset(seed=123)` is not within the observation space."
             if env.spec is not None and env.spec.nondeterministic is False:
                 assert data_equivalence(
-                    obs_1, obs_2
+                    obs_1, obs_2, exact=True
                 ), "Using `env.reset(seed=123)` is non-deterministic as the observations are not equivalent."
             assert (
                 env.unwrapped._np_random.bit_generator.state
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
Like @RedTachyon said, there are cases where floating point makes exact a very difficult standard for projects to require. 
I would prefer to reduce the `allclose` `atol` or `rtol` values than change to `allequal`

Kallinteris-Andreas:
As I previously said, the operations in `check_env` performs operations deterministically, so there is no concern of FP arithmetic ordering errors

RedTachyon:
Nitpick:

image

(yes, this uses nan's - although those would also fail approximate equality tests)

RedTachyon:
Can't we do like a two-tiered check? If they're not `allclose`, then throw an error because that's a big issue (like what happens now). If it is `allclose` but isn't `==`, then raise a warning.

Alternatively, maybe we can just add a switch in `check_env` controlling how strict the equality checks are.

pseudo-rnd-thoughts:
I agree with @RedTachyon on the two-tier testing

Kallinteris-Andreas:
I'm sorry, I do not understand the benefit of a two-tier testing System.  All deterministic environments.  Should fail if they cannot pass `allequal` 

note: NaN It is already tested before the determinism checks.

pseudo-rnd-thoughts:
From my perspective, we can take the `check_env` function in two directions.
The first is a general testing function for most environments that will generally test possible values but not provide a guarantee. To test determinism, we only test the first reset and step values, ignoring any roll out, meaning that in many ways the function is not that strict. 

The second is a comprehensive testing tool that covers everything about the function. 

In my opinion, we should test the essential core elements of the API comprehensively, while the non-essential elements (i.e., determinism) should be more lacks. Particularly for determinism that is very difficult to guarantee for some environment with external simulators, etc and as Gymnasium is a general API we should try to find a middle ground on this.  

Therefore, a two-tier equivalence system could help provide a middle ground to purely alert users if strict equivalence doesn't hold but not failure that would prevent the rest of the testing suite to complete. 

For non-essential items, I don't want the testing to fail and the user disregard the function when later parts of the function would show essential elements actually failing if they continued using it

gymnasium/utils/env_checker.py
```diff
@@ -99,7 +99,7 @@ def check_reset_seed_determinism(env: gym.Env):
             ), "The observation returned by `env.reset(seed=123)` is not within the observation space."
             if env.spec is not None and env.spec.nondeterministic is False:
                 assert data_equivalence(
```
pseudo-rnd-thoughts, line 101: Add warning case and remove exact=True in this case

gymnasium/utils/env_checker.py
```diff
@@ -189,13 +189,26 @@ def check_step_determinism(env: gym.Env, seed=123):
...
+    if not data_equivalence(obs_0, obs_1, exact=True):
+        logger.warn("step observation is not deterministic.")
     assert data_equivalence(obs_0, obs_1), "step observation is not deterministic."
+
+    if not data_equivalence(rew_0, rew_1, exact=True):
+        logger.warn("step reward is not deterministic.")
     assert data_equivalence(rew_0, rew_1), "step reward is not deterministic."
-    assert data_equivalence(term_0, term_0), "step terminal is not deterministic."
+
+    assert data_equivalence(term_0, term_0, True), "step terminal is not deterministic."
```
pseudo-rnd-thoughts, line 201: `exact=True`

pseudo-rnd-thoughts (changes requested):
Could the warning and error messages be a bit different otherwise it could be difficult for users to understand what is happening

pseudo-rnd-thoughts (approved):

Issue #790, opened 2023-11-24
[Bug Report] `CartPole`'s reward function constantly returns 1 (even when it falls)


Kallinteris-Andreas:
### Describe the bug

The `CartPole` environment provides `reward==1` when the pole "stands"  and `reward==1` when the pole has "fallen".

The [old gym documentation](https://github.com/openai/gym/blob/e8f29806033861f196cbc679788b985f4b40c4b7/gym/envs/classic_control/cartpole.py) mentioned that this was the behavior, and so does the current [documentation](https://gymnasium.farama.org/v0.29.0/environments/classic_control/cart_pole/), indicating that this is the desired behavior, but I can find no evidence that this was the design goal.

The argument could be made that "state-of-the-art algorithms (e.g. `DQN`, `A2C`, `PPO`) can easily solve this environment anyway, so why bother?", which is true that the environment is very easy to solve, but it is still important to examine why, and what the consequences are in these learning algorithms and potential future ones.

The reason the algorithms are able to learn a policy, despite the fact that the environment is effectively reward-free, is because of the sampling bias introduced by the number of actions taken per given policy.
If the agent has a "good policy", then on average it will be alive longer than if it has a "bad policy", and this will cause the "good policy" to be sampled more often on average by the `train()` subroutine of the learning algorithm, and therefore a "good policy" will be reinforced more than a "bad policy" (instead of the "good policy" being reinforced and the "bad policy" being reduced, which is what normally happens in RL).

This would mean that RL algorithms that are not affected by this sampling bias (or are much less affected by it) would not be able to learn a policy for the `CartPole-v1` environment,
And the CartPole environment is where many RL algorithms are tested during the prototyping phase (since it is probably assumed that if a learning algorithm cannot solve CartPole, it is probably a dud).
Therefore, if an algorithm that was developed was less affected by this sampling bias, it might have failed the prototyping phase because of the wrong implementation of `Gymnasium/CartPole`.

# Quick Performance Analysis
![CartPole](https://github.com/Farama-Foundation/Gymnasium/assets/30759571/f99844ed-0220-4a9a-b3f0-5b203573bf0b)

- `DQN` seems to be benefit significantly from fixing this bug.
- `A2C` seem to benefit slightly from fixing this bug.
- `PPO` does not seem to be affected by this bug, as it can already easily solve the problem.

Note: The shaded area is the `(min, max)` episode length of each algorithm.

# Suggestion
## Suggestion 1 (the one I recommend)
[...]

PR #958, opened 2024-03-08
cartpole v2

Kallinteris-Andreas:
# Description
adds `CartPole-v2` with the original reward function from sutton, barto 
note: can use the previous reward function with `sutton_barto_reward=False`

Fixes https://github.com/Farama-Foundation/Gymnasium/issues/790

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/envs/classic_control/cartpole.py
+++ b/gymnasium/envs/classic_control/cartpole.py
@@ -56,16 +56,19 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
        if the pole angle is not in the range `(-.2095, .2095)` (or **±12°**)
 
     ## Rewards
+    In v2 the reward matches the original reward, `0` for every non-terminating step taken, and `-1` for terminating steps.
 
+    In v1 and v0 the rewards is:
     Since the goal is to keep the pole upright for as long as possible, a reward of `+1` for every step taken,
-    including the termination step, is allotted. The threshold for rewards is 500 for v1 and 200 for v0.
+    including the termination step, is allotted. 
+    
+    ## Reward Threshold
+    The threshold for rewards is 0 for v2, 500 for v1 and 200 for v0.
 
     ## Starting State
-
     All observations are assigned a uniformly random value in `(-0.05, 0.05)`
 
     ## Episode End
-
     The episode ends if any one of the following occurs:
 
     1. Termination: Pole Angle is greater than ±12°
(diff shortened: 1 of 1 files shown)


Kallinteris-Andreas:
@pseudo-rnd-thoughts 
I have changed the CartPole code and the documentation, can you review it and if it passes review I can move into updating the vector and jax versions

RedTachyon:
Why do we want this to be a v2, as opposed to just adding the optional argument to the constructor and letting users use it?

And in any case, please write the reward computation a bit more explicitly, converting booleans to floats and manipulating them is clever and all that, but unreadable and annoying to maintain in the future. A ternary operator would probably work nicely here.


I didn't participate in the original issue, and I'm overall not convinced this is necessary since both rewards are perfectly valid (albeit a bit different), but I won't oppose adding it as an optional setting. However, if we make it the v2 version, there are two problems:
- The already messy fact that we have v0 and v1 concurrently, becomes even more messy due to having v0, v1 and v2
- We are sorta officially stating that v2 is the "correct" version, which I'm not fully convinced of (though I might need to read the original thread in more detail)

Kallinteris-Andreas:
As far as I can tell, new arguments have not been added mid-environment version

Also I believe that the previous version was technical incorrect (even if it worked.)

pseudo-rnd-thoughts:
Environment versions have been used to indicate that a bug in the previous version has been fixed, which would change the dynamic, so users should be alerted to this change. 

Therefore, as this is an addition that can't affect previous users, this should not be a version change, particularly as a feature is turned off by default

Kallinteris-Andreas:
Now it simply adds the new argument. I have redone the pull request. And does not bump the environment version

gymnasium/envs/classic_control/cartpole.py
```diff
@@ -84,9 +86,12 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
     <TimeLimit<OrderEnforcing<PassiveEnvChecker<CartPoleEnv<CartPole-v1>>>>>
     >>> env.reset(seed=123, options={"low": -0.1, "high": 0.1})  # default low=-0.05, high=0.05
     (array([ 0.03647037, -0.0892358 , -0.05592803, -0.06312564], dtype=float32), {})
-
```
pseudo-rnd-thoughts, line 87: You need to re-add this for the doctest

gymnasium/envs/classic_control/cartpole.py
```diff
@@ -56,16 +56,18 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
        if the pole angle is not in the range `(-.2095, .2095)` (or **±12°**)
 
     ## Rewards
-
     Since the goal is to keep the pole upright for as long as possible, a reward of `+1` for every step taken,
```
pseudo-rnd-thoughts, line 59: Since the goal is to keep the pole upright for as long as possible, by default, a reward of `+1` is given for every step taken, including the termination step. The default reward threshold is 500 for v1 and 200 for v0 due to the time limit on the environment. 

If `sutton_barto_reward=True`, then a reward of `0` is awarded for every non-terminating step and `-1` for the terminating step. As a result, the reward threshold is 0 for v0 and v1.

gymnasium/envs/classic_control/cartpole.py
```diff
@@ -56,16 +56,18 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
...
-    including the termination step, is allotted. The threshold for rewards is 500 for v1 and 200 for v0.
+    including the termination step, is allotted.
 
-    ## Starting State
+    If `sutton_barto_reward=True` then a reward of `0` is awarded for every non-terminating step and `-1` for the terminating step.
+
+    ## Reward Threshold
```
pseudo-rnd-thoughts, line 64: remove for comment above

pseudo-rnd-thoughts (approved):
With the two comments

Kallinteris-Andreas:
I cannot understand why the testing is failing.

Kallinteris-Andreas:
I realized after merging that we need to update the vector and jax versions also

PR #959, opened 2024-03-08
Randomize LunarLander wind generation at reset to gain statistical independence between episodes

### LINKED ISSUE #954: [Proposal] Randomize LunarLander wind generation at reset to gain statistical independence between episodes
### Proposal

In `gymnasium/envs/box2d/lunar_lander.py`, randomly draw a new `wind_idx` and `torque_idx` when the environment is reset through `reset()`. The cleanest fix would be to move the lines
```
        self.wind_idx = np.random.randint(-9999, 9999)
        self.torque_idx = np.random.randint(-9999, 9999)
```
From the `__init__()` method to the `reset()` method. This would ensure that the environment is statistically independent between episodes, which it is currently not. I believe it is important to have statistical independence between episodes since researchers are expecting this behavior, e.g., for Monte Carlo simulation.

### Motivation

When simulating the lunar lander environment with wind enabled, it seems that the variables `wind_idx` and `torque_idx`, which are only drawn randomly in the `__init__()` call, greatly affect the environment. Though the variables continuously increase with every step after initialization, some initial choices generally create an easier environment than others. I encountered this problem when evaluating landing/crashing performances for a trained reinforcement learning policy. In one example, I got the following statistics from 5000 episodes:
```
Landing rate: 81.62% +- 1.80%
Crash rate: 10.54% +- 1.43%
```
Where the `+-` indicates a 99.99% confidence interval. Now, in another example (where the environment was initialized again), the following statistics were obtained:
```
Landing rate 94.84% +- 1.03%
Crash rate 2.02% +- 0.65%
```
The only plausible explanation for this behavior is that **each episode is not statistically independent**. I tested changing the code to draw independent `wind_idx` and `torque_idx` each time the environment is reset (as described in the proposal), and **now I get the same landing and crash rate for every simulation**. My best explanation for this behavior is the fact that randomly drawing `wind_idx` and `torque_idx` permanently affects the phase offsets of the wind and torque variables `wind_mag` and `torque_mag`, which may or may not align favorably with the average length of each episode. 

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

TobiasKallehauge:
# Description

This request changes the way `gymnasium/envs/box2d/lunar_lander.py` randomly draws a new `wind_idx` and `torque_idx` so new indexes are drawn randomly whenever the environment is reset rather than only at initialization. This ensures that the environment is statistically independent between episodes, which it is currently not. Changed the version from `v2` to `v3` and added a unit test to check that the seed is correctly working in the new version. 

Fixes #954

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/envs/box2d/lunar_lander.py
+++ b/gymnasium/envs/box2d/lunar_lander.py
@@ -179,7 +179,7 @@ class LunarLander(gym.Env, EzPickle):
      The recommended value for `turbulence_power` is between 0.0 and 2.0.
 
     ## Version History
-    - v3: Reset wind and turbulence offset (`C`) whenever the environment is reset to ensure statistical independence between consecutive episodes.
+    - v3: Reset wind and turbulence offset (`C`) whenever the environment is reset to ensure statistical independence between consecutive episodes (related [GitHub issue](https://github.com/Farama-Foundation/Gymnasium/issues/954)).
     - v2: Count energy spent and in v0.24, added turbulence with wind power and turbulence_power parameters
     - v1: Legs contact with ground added in state vector; contact with ground give +10 reward points,
           and -10 if then lose contact; reward renormalized to 200; harder initial random push.
--- a/gymnasium/envs/box2d/lunar_lander.py
+++ b/gymnasium/envs/box2d/lunar_lander.py
@@ -179,6 +179,7 @@ class LunarLander(gym.Env, EzPickle):
      The recommended value for `turbulence_power` is between 0.0 and 2.0.
 
     ## Version History
+    - v3: Reset wind and turbulence offset (`C`) whenever the environment is reset to ensure statistical independence between consecutive episodes.
     - v2: Count energy spent and in v0.24, added turbulence with wind power and turbulence_power parameters
     - v1: Legs contact with ground added in state vector; contact with ground give +10 reward points,
           and -10 if then lose contact; reward renormalized to 200; harder initial random push.
--- a/gymnasium/envs/box2d/lunar_lander.py
+++ b/gymnasium/envs/box2d/lunar_lander.py
@@ -254,8 +254,6 @@ class LunarLander(gym.Env, EzPickle):
         self.turbulence_power = turbulence_power
 
         self.enable_wind = enable_wind
-        self.wind_idx = np.random.randint(-9999, 9999)
-        self.torque_idx = np.random.randint(-9999, 9999)
 
         self.screen: pygame.Surface = None
         self.clock = None
@@ -402,6 +400,11 @@ class LunarLander(gym.Env, EzPickle):
             ),
             True,
         )
+        
+
+        if self.enable_wind: # Initialize wind pattern based on index
+            self.wind_idx = np.random.randint(-9999, 9999)
+            self.torque_idx = np.random.randint(-9999, 9999)
 
         # Create Lander Legs
         self.legs = []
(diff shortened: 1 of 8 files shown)


gymnasium/envs/box2d/lunar_lander.py
```diff
@@ -403,6 +401,10 @@ def reset(
             True,
         )
 
+        if self.enable_wind:  # Initialize wind pattern based on index
+            self.wind_idx = np.random.randint(-9999, 9999)
```
Kallinteris-Andreas, line 405: `self.np_random.` As Mark said, in ththie issue

Kallinteris-Andreas (changes requested):
We will also need Unit Testing to validate the changes.

gymnasium/envs/box2d/lunar_lander.py
TobiasKallehauge: Yes, sorry - I did not understand before. It has been fixed now.

TobiasKallehauge: I cannot get the all the doctests to run on my machine (also not on the main branch). Is there a command to only test the updated files?
pytest

pseudo-rnd-thoughts: Don't worry, this passes all the CI testing

TobiasKallehauge:
I committed the version change to the documentation and updated the default version used for tests. Let me know if something else should be changed.

Kallinteris-Andreas:
I want to see a unit test validating the change.

also Bump the version in registration

pseudo-rnd-thoughts:
@Kallinteris-Andreas What do you mean? 

@TobiasKallehauge Could you add a new test in `tests/envs/test_env_implementations.py` that using different `reset(seed)` will cause `wind_idx` and `torque_idx` to change. (Probably use known seeds to avoid future randomness issues)

You need to change `gymnasium/envs/__init__.py` on the `lunarlander` registration to be `v3` not `v2`

TobiasKallehauge:
I added the unit test now, and the seed works as expected - I hope the test complies with your standards. 

There was still an error in the previous commit for the version due to changing the version number but maybe the new one will pass after changing the tests

TobiasKallehauge:
There seems to be a few more places where the version number is referenced:
- `gymnasium/wrappers/rendering.py`
- `docs/tutorials/gymnasium_basics/vector_envs_tutorial.py`

Right now `gymnasium/wrappers/rendering.py` is throwing an error. There are also some markdown files where it is references:

- `docs/index.md`
- `docs/introduction/migration_guide.md`
- `docs/introduction/basic_usage.md`

I will update the version in all these places (including in the markdown files) - let me know if this is wrong

Kallinteris-Andreas (changes requested):
Overall this looks good to me.

gymnasium/envs/box2d/lunar_lander.py
```diff
@@ -179,6 +179,7 @@ class LunarLander(gym.Env, EzPickle):
      The recommended value for `turbulence_power` is between 0.0 and 2.0.
 
     ## Version History
+    - v3: Reset wind and turbulence offset (`C`) whenever the environment is reset to ensure statistical independence between consecutive episodes.
```
Kallinteris-Andreas, line 182: Can you add a link to the related GitHub issue See, MuJoC, ant for example. https://gymnasium.farama.org/main/environments/mujoco/ant/#version-history

TobiasKallehauge: Of course - its done

pseudo-rnd-thoughts (approved):
Thanks for the PR @TobiasKallehauge and your rapid responses

TobiasKallehauge:
Thanks to you as well! I am happy to contribute

Issue #950, opened 2024-03-05
[Question] Problems with Collisions in Pusher-v4


HernandezEduin:
### Question

I'm testing out the RL training with cleanRL, but I noticed in the video provided below that the robotic arm goes through both the table and the object it is supposed to be pushing.

The code below also has a similar behaviour:

```
import gymnasium as gym
env = gym.make("Pusher-v4", render_mode="human")
observation, info = env.reset(seed=39)
for _ in range(1000):
   action = env.action_space.sample()  # this is where you would insert your policy
   observation, reward, terminated, truncated, info = env.step(action)

   if terminated or truncated:
      observation, info = env.reset()

env.close()
```
Any idea what is going on or what might be missing?

---
System info
---
Installed via pip version 0.29.1
```
Python 3.8.12
Gymnasium 0.29.1
Mujoco 3.1.2
```
---

https://github.com/Farama-Foundation/Gymnasium/assets/86664284/0e28d81f-8633-4a78-a716-4f244dd1ff32

Kallinteris-Andreas:
1) does this also happen with `mujoco==2.3.7`
2) can you produce the video with a different angle

HernandezEduin:
1. Yes, same issue with Mujoco==2.3.7.
2. Is there a way to modify the angle with the Gymnasium record_video?

Kallinteris-Andreas:
2.
Yes, you need to change the `azimuth` `elevation` camera properties 

```py
import gymnasium as gym
import numpy as np

DEFAULT_CAMERA_CONFIG = {
    "trackbodyid": -1,
    #"distance": 4.0,
    "distance": 3.0,
    "azimuth": 135.0,
    "elevation": -22.5,
}

env = gym.make("Pusher-v5", default_camera_config=DEFAULT_CAMERA_CONFIG, render_mode="human")
```

HernandezEduin:
2. The changing of camera option is not present for Gymnasium 0.29.1. If I upgrade to gymnasium==1.0.0a1, the option becomes available along with Pusher-v5, but I'm no longer able to run CleanRL due to compatibilities issues of this version of Gym and Stable-Baselines3.

Kallinteris-Andreas:
You could try this:

```py
import gymnasium

DEFAULT_CAMERA_CONFIG = {
    "trackbodyid": -1,
    #"distance": 4.0,
    "distance": 3.0,
    "azimuth": 135.0,
    "elevation": -22.5,
}

[...]

PR #963, opened 2024-03-12
Raise error for `pusher-v4` environment when `mujoco>=3`

Kallinteris-Andreas:
# Description

https://github.com/Farama-Foundation/Gymnasium/issues/950 noticed that pusher-v4 had collision issues which was isolated to a bug fixed in mujoco==3.0.0
This collision issue has a major implement on performance, therefore, raising an error for pusher-v4 when mujoco>=3
The issue will be fixed in pusher-v5

### DIFF
--- a/gymnasium/envs/mujoco/pusher_v4.py
+++ b/gymnasium/envs/mujoco/pusher_v4.py
@@ -1,3 +1,4 @@
+import mujoco
 import numpy as np
 
 from gymnasium import utils
@@ -22,7 +23,9 @@ class PusherEnv(MujocoEnv, utils.EzPickle):
     }
 
     def __init__(self, **kwargs):
+        assert mujoco.__version__ < "3.0.0", "`Pusher-v4` is only supported on `mujuco<3`."
         utils.EzPickle.__init__(self, **kwargs)
+        
         observation_space = Box(low=-np.inf, high=np.inf, shape=(23,), dtype=np.float64)
         MujocoEnv.__init__(
             self,
--- a/gymnasium/envs/mujoco/pusher_v4.py
+++ b/gymnasium/envs/mujoco/pusher_v4.py
@@ -25,7 +25,7 @@ class PusherEnv(MujocoEnv, utils.EzPickle):
     def __init__(self, **kwargs):
         assert (
             mujoco.__version__ < "3.0.0"
-        ), "`Pusher-v4` is only supported on `mujuco<3`."
+        ), "`Pusher-v4` is only supported on `mujuco<3`, for more information https://github.com/Farama-Foundation/Gymnasium/issues/950."
         utils.EzPickle.__init__(self, **kwargs)
 
         observation_space = Box(low=-np.inf, high=np.inf, shape=(23,), dtype=np.float64)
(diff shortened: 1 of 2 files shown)


Kallinteris-Andreas:
What do we do for testing?

pseudo-rnd-thoughts:
> What do we do for testing?

Remove from the set of testing environments

We could add a specialist CI case where mujoco < 3 is installed just in case. 
But with v5 fixing the issue, I don't imagine many people will use it. 

Is there anyway that we can check that a similar bug doesn't exist in other environments?

gymnasium/envs/mujoco/pusher_v4.py
```diff
@@ -22,7 +23,11 @@ class PusherEnv(MujocoEnv, utils.EzPickle):
     }
 
     def __init__(self, **kwargs):
+        assert (
+            mujoco.__version__ < "3.0.0"
+        ), "`Pusher-v4` is only supported on `mujuco<3`."
```
pseudo-rnd-thoughts, line 28: I would add a link to the issue for users to understand why

pseudo-rnd-thoughts (approved):
@Kallinteris-Andreas Were you planning on anymore changes?

Kallinteris-Andreas:
I have no idea why this is passing CI

Kallinteris-Andreas:
Did not realize you commitef to it, makes sense now

tests/envs/mujoco/test_mujoco_v5.py
```diff
@@ -331,15 +331,17 @@ def test_reward_sum(version: str):
         env_conf("HumanoidStandup", True, False, False, "superset"),
         env_conf("InvertedDoublePendulum", True, True, False, "superset"),
         env_conf("InvertedPendulum", False, True, False, "superset"),
-        env_conf("Pusher", False, True, False, "keys-superset"),
+        env_conf("Pusher", True, True, False, "keys-superset"),  # pusher-v4
```
pseudo-rnd-thoughts, line 334: Why was this changed?

tests/envs/mujoco/test_mujoco_v5.py
```diff
@@ -584,17 +586,18 @@ def test_model_object_count(version: str):
...
-    assert env.model.nq == 11
-    assert env.model.nv == 11
-    assert env.model.nu == 7
-    assert env.model.nbody == 13
-    if mujoco.__version__ >= "3.1.2":
-        assert env.model.nbvh == 8
-    assert env.model.njnt == 11
-    assert env.model.ngeom == 21
-    assert env.model.ntendon == 0
+    if not (version == "v4" and mujoco.__version__ >= "3.0.0"):
```
pseudo-rnd-thoughts, line 589: I think this is wrong, I think it should be version is v4 and mujoco < 3

pseudo-rnd-thoughts (changes requested):

tests/envs/mujoco/test_mujoco_v5.py
Kallinteris-Andreas: Same thing

pseudo-rnd-thoughts: No, as this should run for pusher v4 and mujoco < 3 which this won't

Issue #993, opened 2024-04-03
[Proposal] Remove `mujoco-py` based environments at May 2025


Kallinteris-Andreas:
### Proposal

`mujoco-py` is an old unmaintained version of MuJoCo bindings (`mujoco` is the new version), and its inclusion in the gymnasium project is causing some issues (the environments that still use that library are v2 and v3 of the MuJoCo environments).

At May 2022 the first `mujoco` based environments were released (`Ant-v4` and other MuJoCo-v4 environments) and they have been a big improvement, but the old environments are still used for reproducibility of old experiments, so we are giving a 3-year period from the release of new environments (1 year after this post) for which we will still support them and include them in the repository, afterward they will be moved to a separate repository and will not be supported. 

The `mujoco-py` environments are considered deprecated until their removal

### Motivation

1) `Cython>=3` is not supported with `mujoco-py` why restricts other libraries that can be used https://github.com/Farama-Foundation/Gymnasium/issues/921

### Alternatives

There is the option to create a `mujoco-py` soft fork `PyPI` package that just update the build script, but i am choosing to not follow it as an economization of development time

https://github.com/openai/mujoco-py/pull/782

that being said if someone i willing to assist it might happen. 

### Additional context

Note this also applies https://robotics.farama.org/ environments

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
While I think its a reasonable suggestion to move the mujoco-py environments into another repo such that users can still access them if necessary, `gymnasium-robotics` would be the most obvious location. 

But I don't totally understand the issue for keeping them in the repo? 
The problem with cpython is a semi-permentant issue with mujoco-py no matter the location, therefore, this is more a problem for users than the repo imho. 

The biggest motivation would be to encourage users to migrate to v4 / v5 through making them no longer part of the easy install

ReHoss:
+1 since `mujoco-py` triggers exceptions like below:

```
Exception: 
You appear to be missing MuJoCo.  We expected to find the file here: /home/mpanchen/.mujoco/mujoco210
```

Also appearing in https://github.com/Farama-Foundation/Gymnasium/issues/896

RogerJL:
Prior to release of 1.0 would be the perfect time to remove this - no API promises on alfa/beta...
If someone needs backward compatibility they can use old git version of Gymnasium IMHO.

[...]

PR #994, opened 2024-04-04
use different `mujoco-py` build options(under testing)

Kallinteris-Andreas:
https://github.com/Kallinteris-Andreas/mujoco-py is a soft fork `mujoco-py` that simply changes the build flags

if this gets merged or a similar solution is found, we can delay https://github.com/Farama-Foundation/Gymnasium/issues/993

```py
>>> import gymnasium
>>> import cython
>>> cython.__version__
'3.0.10'
>>> env = gymnasium.make("Ant-v2")
/home/master-andreas/temp/Gymnasium/gymnasium/envs/registration.py:518: DeprecationWarning: WARN: The environment Ant-v2 is out of date. You should consider upgrading to version `v5`.
  logger.deprecation(
/home/master-andreas/temp/Gymnasium/gymnasium/envs/mujoco/mujoco_py_env.py:217: DeprecationWarning: WARN: This version of the mujoco environments depends on the mujoco-py bindings, which are no longer maintained and may stop working. Please upgrade to the v5 or v4 versions of the environments (which depend on the mujoco python bindings instead), unless you are trying to precisely replicate previous works).
  logger.deprecation(
```

# Advantages
- removes the requirement for  `cython <3`

# Downsides
- if the user runs `pip gymnasium[mujoco-py]` or `pip gymnasium[all]` with `mujoco-py` pre-installed, it will re-install `mujoco-py` potentially changing the version they had installed
- other unknown downsides I can not foresee

# Alternatives
make a `pypi` package for this soft fork `mujoc-py-updated`???, but I would rather not do that

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,8 +39,8 @@ accept-rom-license = ["autorom[accept-rom-license] ~=0.4.2"]
 box2d = ["box2d-py ==2.3.5", "pygame >=2.1.3", "swig ==4.*"]
 classic-control = ["pygame >=2.1.3"]
 classic_control = ["pygame >=2.1.3"]  # kept for backward compatibility
-mujoco-py = ["mujoco-py@https://github.com/Kallinteris-Andreas/mujoco-py"]
-mujoco_py = ["mujoco-py@https://github.com/Kallinteris-Andreas/mujoco-py"]       # kept for backward compatibility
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
I would prefer to include this is as an "unofficial" solution such that we link to your repo / pypi for cython 3 but this is not the default solution

PR #998, opened 2024-04-05
Re-enable `check_env` for JAX environments

Kallinteris-Andreas:
# Description


pseudo-rnd-thoughts:
I did this intentionally as if a user creates one of the jax environments then they will get "obs not in observation_space" as the observation spaces don't support jax arrays

pseudo-rnd-thoughts:
This was actually just disabling the passive environment checker applied on make not `check_env`

PR #1000, opened 2024-04-06
`Env.metadata` docstring add note for `jax` and numpy environments

Kallinteris-Andreas:


pseudo-rnd-thoughts (approved):
Good idea, I have expanded it to provide an example metadata

PR #1004, opened 2024-04-10
Add testing for Python 3.12

pseudo-rnd-thoughts:
# Description

With Tensorflow and PyTorch updated, we can add testing for Python 3.12 in CI


pseudo-rnd-thoughts:
Shimmy or ale-py is missing python 3.12 support

PR #1013, opened 2024-04-12
Add `wrappers.vector.HumanRendering`

pseudo-rnd-thoughts:
# Description

Rather than add human rendering to every vector environment, this wrapper hopes to add human style rendering as a wrapper for environments to use.
Completes https://github.com/Farama-Foundation/Gymnasium/pull/880 and fixes cartpole rendering in https://github.com/Farama-Foundation/Gymnasium/issues/872

### DIFF
--- a/gymnasium/envs/classic_control/cartpole.py
+++ b/gymnasium/envs/classic_control/cartpole.py
@@ -339,13 +339,13 @@ class CartPoleEnv(gym.Env[np.ndarray, Union[int, np.ndarray]]):
 
 class CartPoleVectorEnv(VectorEnv):
     metadata = {
-        "render_modes": ["human", "rgb_array"],
+        "render_modes": ["rgb_array"],
         "render_fps": 50,
     }
 
     def __init__(
         self,
-        num_envs: int = 2,
+        num_envs: int = 1,
         max_episode_steps: int = 500,
         render_mode: Optional[str] = None,
     ):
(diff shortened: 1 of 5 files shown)


pseudo-rnd-thoughts:
@TimSchneider42 Could you test this PR and see if it works for you?

TimSchneider42:
I had to make an adjustment to make it work:
```diff
diff --git a/gymnasium/wrappers/vector/rendering.py b/gymnasium/wrappers/vector/rendering.py
index c4f691a44..1d7463152 100644
--- a/gymnasium/wrappers/vector/rendering.py
+++ b/gymnasium/wrappers/vector/rendering.py
@@ -131,7 +131,9 @@ class HumanRendering(VectorWrapper):
             assert (cols * subenv_size[0] * scaling_factor == self.screen_size[0]) or (
                 rows * subenv_size[1] * scaling_factor == self.screen_size[1]
             )
-            return rows, cols, scaling_factor
+            self.num_rows = rows
+            self.num_cols = cols
+            self.scaled_subenv_size = (int(subenv_size[0] * scaling_factor), int(subenv_size[1] * scaling_factor))
 
             assert self.num_rows * self.num_cols >= self.num_envs
             assert self.scaled_subenv_size[0] * self.num_cols <= self.screen_size[0]
```

Apart from this, it worked fine.

One thing that is confusing to me is that human rendering can no longer be enabled with the render_mode argument but rather requires the user to apply the wrapper manually. We could consider wrapping the classic control environments automatically if the render mode is human.

Also, is there a reason why this wrapper defines the render function to always return none? We could just leave it undefined so that the user can still call the original render function if they require a RGB image.

Best,
Tim

PR #1014, opened 2024-04-12
mujoco tests, disable nbvh checks

Kallinteris-Andreas:
# Description
since we do not guarantee identical rendering behavior between `mujoco` versions ISO state, there is no benefit to keeping this check


pseudo-rnd-thoughts (commented):
Is it not helpfully to understand the underlying changes to mujoco. Given that it is only one of the environments this affect I wouldn't remove it

PR #1018, opened 2024-04-15
Fixed `wrappers.vector.RecordEpisodeStatistics` episode length computation from new autoreset api

### LINKED ISSUE #1017: [Bug Report] wrappers.vector.RecordEpisodeStatistics computes episode length incorrectly
### Describe the bug

Hi,

`wrappers.vector.RecordEpisodeStatistics` computes the episode lengths incorrectly. It still adheres to the old API, in which the final observations of episodes were never returned. This causes the wrapper to be off by one in all episodes except for the first one.

For example in this case:
```python
import gymnasium as gym

env = gym.wrappers.vector.RecordEpisodeStatistics(gym.make_vec(id="CartPole-v1", num_envs=1, max_episode_steps=3))
env.reset(seed=0)

for i in range(15):
    print(f"Step {i}")
    *_, info = env.step(env.action_space.sample())
    if "episode" in info:
        print(f"Length: {info['episode']['l'][0]}")
```

`max_episode_steps` is set to 3, so we would expect all episodes to be exactly 3 steps long (it is not possible to end the episode faster than that in CartPole-v1). However, the output of this script is
```
Step 0
Step 1
Step 2
Length: 3.0
Step 3
Step 4
Step 5
Step 6
Length: 4.0
Step 7
Step 8
Step 9
Step 10
Length: 4.0
Step 11
Step 12
Step 13
Step 14
Length: 4.0
```

It is easy to get confused here, as the first step of each episode is not a real step but the output of the reset function and, thus, does not count as a step.

Best,
Tim

### Code example

_No response_

### System info

_No response_

### Additional context

_No response_

### Checklist

TimSchneider42:
# Description

Fixed wrappers.vector.RecordEpisodeStatistics episode length computation.

Fixes #1017

## Type of change

# Checklist:

### DIFF
--- a/tests/wrappers/vector/test_record_episode_statistics.py
+++ b/tests/wrappers/vector/test_record_episode_statistics.py
@@ -35,7 +35,7 @@ def test_record_episode_statistics(num_envs, env_id="CartPole-v1", num_steps=100
     assert data_equivalence(wrapper_vector_obs, vector_wrapper_obs)
     assert data_equivalence(wrapper_vector_info, vector_wrapper_info)
 
-    for _ in range(num_steps):
+    for t in range(1, num_steps + 1):
         action = wrapper_vector_env.action_space.sample()
         (
             wrapper_vector_obs,
@@ -52,20 +52,22 @@ def test_record_episode_statistics(num_envs, env_id="CartPole-v1", num_steps=100
             vector_wrapper_info,
         ) = vector_wrapper_env.step(action)
 
-        data_equivalence(wrapper_vector_obs, vector_wrapper_obs)
-        data_equivalence(wrapper_vector_reward, vector_wrapper_reward)
-        data_equivalence(wrapper_vector_terminated, vector_wrapper_terminated)
-        data_equivalence(wrapper_vector_truncated, vector_wrapper_truncated)
+        assert data_equivalence(wrapper_vector_obs, vector_wrapper_obs)
+        assert data_equivalence(wrapper_vector_reward, vector_wrapper_reward)
+        assert data_equivalence(wrapper_vector_terminated, vector_wrapper_terminated)
+        assert data_equivalence(wrapper_vector_truncated, vector_wrapper_truncated)
 
         if "episode" in wrapper_vector_info:
-            assert "episode" in vector_wrapper_info
-
             wrapper_vector_time = wrapper_vector_info["episode"].pop("t")
             vector_wrapper_time = vector_wrapper_info["episode"].pop("t")
             assert wrapper_vector_time.shape == vector_wrapper_time.shape
             assert wrapper_vector_time.dtype == vector_wrapper_time.dtype
 
-        data_equivalence(wrapper_vector_info, vector_wrapper_info)
+            vector_wrapper_info["episode"].pop("_l")
+            vector_wrapper_info["episode"].pop("_r")
+            vector_wrapper_info["episode"].pop("_t")
+
+        assert data_equivalence(wrapper_vector_info, vector_wrapper_info)
 
     wrapper_vector_env.close()
     vector_wrapper_env.close()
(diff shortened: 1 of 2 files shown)


tests/wrappers/vector/test_record_episode_statistics.py
```diff
@@ -56,16 +60,41 @@ def test_record_episode_statistics(num_envs, env_id="CartPole-v1", num_steps=100
         data_equivalence(wrapper_vector_reward, vector_wrapper_reward)
         data_equivalence(wrapper_vector_terminated, vector_wrapper_terminated)
         data_equivalence(wrapper_vector_truncated, vector_wrapper_truncated)
+        data_equivalence(wrapper_vector_info, vector_wrapper_info)
```
pseudo-rnd-thoughts, line 63: I believe this isn't possible due to the episode time taken will vary between the two

This is the reason for the sequent code

A comment would be helpful

tests/wrappers/vector/test_record_episode_statistics.py
```diff
@@ -56,16 +60,41 @@ def test_record_episode_statistics(num_envs, env_id="CartPole-v1", num_steps=100
...
+            == dones[t]
+        )
 
+        if "episode" in wrapper_vector_info:
             wrapper_vector_time = wrapper_vector_info["episode"].pop("t")
             vector_wrapper_time = vector_wrapper_info["episode"].pop("t")
             assert wrapper_vector_time.shape == vector_wrapper_time.shape
             assert wrapper_vector_time.dtype == vector_wrapper_time.dtype
 
-        data_equivalence(wrapper_vector_info, vector_wrapper_info)
```
pseudo-rnd-thoughts, line 68: Once we pop `t`, why can we not just do `data_equivalence` for the two infos?
Why did this test fail previously?

tests/wrappers/vector/test_record_episode_statistics.py
TimSchneider42: Strangely, this test did not fail for me. I will move it back to where it was originally. Not sure why I moved it in the first place.

TimSchneider42:
I moved the info dict comparison back to its original place

tests/wrappers/vector/test_record_episode_statistics.py
pseudo-rnd-thoughts: I messed on the testing of this, it should be `assert data_equivalence(...)`

The new testing code should be 
```python
        assert data_equivalence(wrapper_vector_obs, vector_wrapper_obs)
        assert data_equivalence(wrapper_vector_reward, vector_wrapper_reward)
        assert data_equivalence(wrapper_vector_terminated, vector_wrapper_terminated)
        assert data_equivalence(wrapper_vector_truncated, vector_wrapper_truncated)

        if "episode" in wrapper_vector_info:
            assert "episode" in vector_wrapper_info

            wrapper_vector_time_taken = wrapper_vector_info["episode"].pop("t")
            vector_wrapper_time_taken = vector_wrapper_info["episode"].pop("t")

            assert wrapper_vector_time_taken.shape == vector_wrapper_time_taken.shape
            assert wrapper_vector_time_taken.dtype == vector_wrapper_time_taken.dtype

            vector_wrapper_info["episode"].pop("_l")
            vector_wrapper_info["episode"].pop("_r")
            vector_wrapper_info["episode"].pop("_t")

        assert data_equivalence(wrapper_vector_info, vector_wrapper_info)
```
This should now test it correctly.

I added `max_episode_steps` to the `make_vec` such that episode ends for this to actually be tested (update to main, this was previously broken)

TimSchneider42: Oh, that is why it did not fail when I moved the info dict comparison! Do you want me to fix the test, or are you going to do it?

pseudo-rnd-thoughts: Could you update the test to my suggested code above, I think it is simplier than your code and achieves the same thing.

TimSchneider42: I totally agree that your version is simpler. For some reason, I thought the test did not consider the episode lengths at all, but after looking at it more carefully, I now understand what it actually does. I just changed the code accordingly.

pseudo-rnd-thoughts: Yeah, it does that implicitly with the info check at the end

PR #1023, opened 2024-04-16
Fix missing closing single quote in environment creation tutorial

turbotimon:
# Description

Add missing quote to the end of the string in code example.

## Type of change


pseudo-rnd-thoughts (approved):
Thanks for spotting that and the PR

PR #1080, opened 2024-06-10
Cliffwalking should terminate when running into cliff

jugheadjones10:
# Description

The Cliffwalking environment as inspired from Richard and Sutton's RL book does not currently terminate when the agent walks into the Cliff. However, this is not the intended behaviour as outlined by Sutton. Running into the cliff should be considered a terminal state with a -100 reward.

I know this to be the case because I tried re-creating the graphs in the RL book. Only by terminating at the cliff can I reproduce the results by Sutton.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/envs/toy_text/cliffwalking.py
+++ b/gymnasium/envs/toy_text/cliffwalking.py
@@ -10,7 +10,6 @@ from gymnasium import Env, spaces
 from gymnasium.envs.toy_text.utils import categorical_sample
 from gymnasium.error import DependencyNotInstalled
 
-
 UP = 0
 RIGHT = 1
 DOWN = 2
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Thanks for the PR, good spot. 
For the future, what page number in the book or link to the website page where it says this?

jugheadjones10:
Hi @pseudo-rnd-thoughts,
This is in page number 132 of Sutton's RL book. Here's a screenshot of the page:
image

pseudo-rnd-thoughts:
Technically the environment does what the text says

> "Reward is −1 on all transitions except those into the region marked “The Cliff.” Stepping into this region incurs a reward of −100 and sends the agent instantly back to the start."

And on [line](https://github.com/Farama-Foundation/Gymnasium/blob/b58207fdb723a881d653b2241f6804f40af56ade/gymnasium/envs/toy_text/cliffwalking.py#L167) if the new position is in the `_cliff` then the *new* position is `self.start_state_index` therefore, "sending the agent instantly back to the start"

Therefore, I'm closing the issue, feel free to message if you spot an issue I missed

PR #1086, opened 2024-06-14
Check the determinism of `env.reset(seed=42); env.reset()`

### LINKED ISSUE #1084: [Bug Report] The env checker does not detect when the environment is not deterministic after the first reset.
### Describe the bug

From the [doc](https://gymnasium.farama.org/api/env/#gymnasium.Env.reset)

> Therefore, `reset()` should (in the typical use case) be called with a seed right after initialization and then never again.

This implies that the reset with a seed, followed by a reset without a seed, must be deterministic.
However, when an environment does not respect this rule, the environment checker does not fail.

### Code example

```python
import gymnasium as gym
from gymnasium.utils.env_checker import check_env
import random


class MyEnv(gym.Env):
    def __init__(self):
        super().__init__()
        self.observation_space = gym.spaces.Discrete(5)
        self.action_space = gym.spaces.Discrete(5)

        super().reset(seed=seed)
        if seed is not None:
            obs = self.np_random.integers(5)
        else:
            # generate a random observations, but not based on the inner rng -> bad => non deterministic
            obs = int(random.random() * 5)
        return obs, {}

    def step(self, action):
        obs = self.np_random.integers(5)
        return obs, 0, False, False, {}


register = gym.register(
    id="MyEnv-v0",
    entry_point=MyEnv,
    max_episode_steps=1000,
)

check_env(gym.make("MyEnv-v0"))  # Passes but:

env = gym.make("MyEnv-v0")

env.reset(seed=0)
obs1, _ = env.reset()

env.reset(seed=0)
obs2, _ = env.reset()

assert obs1 == obs2  # Fails
```

### System info

gymnasium 0.29.1 (also tested with 1.0.0a2)
MacOS

### Additional context

Context: realized panda-gym was in the case described above: https://github.com/qgallouedec/panda-gym/issues/94

### Checklist

qgallouedec:
# Description

Fixes #1084 (see the issue for the description)

## Type of change

# Checklist:

-> No need, already documented

- [ ] I have added tests that prove my fix is effective or that my feature works -> 

-> No, but catching the non-determinism of `env.reset` was not part of the test, so I let this for further improvement coverage.
You can still check that the new feature works with the code provided in #1084 (the error is now catched by the env checker):

```log
  File "/Users/quentingallouedec/Gymnasium/t.py", line 32, in <module>
    check_env(gym.make("MyEnv-v0"))  # Passes but:
  File "/Users/quentingallouedec/Gymnasium/gymnasium/utils/env_checker.py", line 412, in check_env
    check_reset_seed_determinism(env)
  File "/Users/quentingallouedec/Gymnasium/gymnasium/utils/env_checker.py", line 116, in check_reset_seed_determinism
    assert data_equivalence(
AssertionError: Using `env.reset(seed=123)` then `env.reset()` is non-deterministic as the observations are not equivalent.
```

### DIFF
--- a/gymnasium/utils/env_checker.py
+++ b/gymnasium/utils/env_checker.py
@@ -91,33 +91,51 @@ def check_reset_seed_determinism(env: gym.Env):
             assert (
                 env.unwrapped._np_random is not None
             ), "Expects the random number generator to have been generated given a seed was passed to reset. Most likely the environment reset function does not call `super().reset(seed=seed)`."
-            seed_123_rng = deepcopy(env.unwrapped._np_random)
-
-            obs_2, info = env.reset(seed=123)
+            seed_123_rng_1 = deepcopy(env.unwrapped._np_random)
+            
+            obs_2, info = env.reset()
             assert (
                 obs_2 in env.observation_space
+            ), "The observation returned by `env.reset()` is not within the observation space."
+
+            obs_3, info = env.reset(seed=123)
+            assert (
+                obs_3 in env.observation_space
             ), "The observation returned by `env.reset(seed=123)` is not within the observation space."
+            seed_123_rng_3 = deepcopy(env.unwrapped._np_random)
+
+            obs_4, info = env.reset()
+            assert (
+                obs_4 in env.observation_space
+            ), "The observation returned by `env.reset()` is not within the observation space."
+
             if env.spec is not None and env.spec.nondeterministic is False:
                 assert data_equivalence(
-                    obs_1, obs_2
+                    obs_1, obs_3
                 ), "Using `env.reset(seed=123)` is non-deterministic as the observations are not equivalent."
-                if not data_equivalence(obs_1, obs_2, exact=True):
+                assert data_equivalence(
+                    obs_2, obs_4
+                ), "Using `env.reset(seed=123)` then `env.reset()` is non-deterministic as the observations are not equivalent."
+                if not data_equivalence(obs_1, obs_3, exact=True):
                     logger.warn(
                         "Using `env.reset(seed=123)` observations are not equal although similar."
                     )
+                if not data_equivalence(obs_2, obs_4, exact=True):
+                    logger.warn(
+                        "Using `env.reset(seed=123)` then `env.reset()` observations are not equal although similar."
+                    )
 
             assert (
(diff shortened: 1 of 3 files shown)


qgallouedec:
Tests pass locally, not sure why they fail here. If you can take a look @pseudo-rnd-thoughts

pseudo-rnd-thoughts:
Strange this fails for me locally

qgallouedec:
It must be good now.

pseudo-rnd-thoughts:
Could you provide an example environment that fails the previous version that this PR fixes.
I was imagining, we loop over several seed values rather than just copy and paste the code which is what seems to happen

qgallouedec:
See the example environment from #1084

```python
import gymnasium as gym
from gymnasium.utils.env_checker import check_env
import random


class MyEnv(gym.Env):
    def __init__(self):
        super().__init__()
        self.observation_space = gym.spaces.Discrete(5)
        self.action_space = gym.spaces.Discrete(5)

        super().reset(seed=seed)
        if seed is not None:
            obs = self.np_random.integers(5)
        else:
            # generate a random observations, but not based on the inner rng -> bad => non deterministic
            obs = int(random.random() * 5)
        return obs, {}

    def step(self, action):
        obs = self.np_random.integers(5)
        return obs, 0, False, False, {}


register = gym.register(
    id="MyEnv-v0",
    entry_point=MyEnv,
    max_episode_steps=1000,
)

check_env(gym.make("MyEnv-v0"))
```


Before: no error, now:

After:
```
  File "/Users/quentingallouedec/Gymnasium/t.py", line 32, in <module>
    check_env(gym.make("MyEnv-v0"))
  File "/Users/quentingallouedec/Gymnasium/gymnasium/utils/env_checker.py", line 412, in check_env
    check_reset_seed_determinism(env)
  File "/Users/quentingallouedec/Gymnasium/gymnasium/utils/env_checker.py", line 116, in check_reset_seed_determinism
    assert data_equivalence(
AssertionError: Using `env.reset(seed=123)` then `env.reset()` is non-deterministic as the observations are not equivalent.
```

pseudo-rnd-thoughts (approved):
Thanks for the example, using this I better understand the PR and why the looping idea that I made doesn't make sense. 
LGTM

Issue #161, opened 2022-11-27
[Proposal] Add transitional probabilities to Taxi and Cliff Walking toy text environments


axb2035:
### Proposal

Only Frozen Lake in the toy text grid world environments implements transitional probabilities. 

Taxi is supposed to have it based on the previous documentation but has never been implemented, always returning 1.0. 

At the same time cliff walking could also be set up to use transitional probabilities using the same approach.

### Motivation

Adding transitional probabilities to taxi will close out a TODO that has been on the list for a long time. It will also bring the environment in line with the source paper, The Fickle Taxi Task - Section 7.1 of Hierarchical Reinforcement Learning with the MAXQ Value Function Decomposition (https://www.jair.org/index.php/jair/article/view/10266/24463).

For cliff walking, it presents and opportunity to add depth to the environment and since it uses the same approach would not add significantly more time or risk.

### Pitch

Taxi
Add transitional probability into taxi toy text environment:
- leverage approach from frozen_lake to supply a transitional probability of 0.8 direction intended, 0.1 left and 0.1 right of intended direction for movement actions.
- the paper proposes that, once the taxi has picked up the passenger and moved one square away from the passenger's source location, the passenger changes their destination location with probability 0.3.
- for taxi transition probabilities for pick up and drop off actions remain 1.0.
- add arguments to enable/disable features: 
      - `is_rainy = True | False` to enable transitional probabilities on taxi movement, defaults to `False`.
      - `fickle_passenger = True | False` to enable the passenger to change their destination once picked up, defaults to `False`.

Cliff walking
Add transitional probability into cliff walking toy text environment by leverage approach from frozen_lake to supply a transitional probability of 0.3 direction intended, 0.3 left and 0.3 right of intended direction for movement actions.
- add arguments to enable/disable transitional probabilities.  
      - `is_slippery = True | False` to enable transitional probabilities on player movement, defaults to `False`.

For both:
- Update unit tests.
- Update documentation.
- Increment versions in registry.

### Alternatives

1. Do nothing. Misses an opportunity to make the toy_text environments consistent and more useful for beginner RL practitioners.

2. Remove transitional probability from taxi and/or cliff walking. In either case `prob` would be removed from the info returned. Removes the need to complete taxi work and will simplify any ongoing maintenance.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

axb2035:
[...]

PR #1087, opened 2024-06-19
Add `is_slippery` option for cliffwalking environment

CloseChoice:
# Description

Supports #161  (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gymnasium/envs/toy_text/cliffwalking.py
+++ b/gymnasium/envs/toy_text/cliffwalking.py
@@ -1,7 +1,7 @@
 from contextlib import closing
 from io import StringIO
 from os import path
-from typing import Optional
+from typing import Optional, List, Tuple, Any, Union
 
 import numpy as np
 
@@ -15,6 +15,13 @@ UP = 0
 RIGHT = 1
 DOWN = 2
 LEFT = 3
+
+POSITION_MAPPING = {
+    UP: [-1, 0],
+    RIGHT: [0, 1],
+    DOWN: [1, 0],
+    LEFT: [0, -1]
+}
 
 
 class CliffWalkingEnv(Env):
@@ -32,6 +39,9 @@ class CliffWalkingEnv(Env):
 
     Adapted from Example 6.6 (page 132) from Reinforcement Learning: An Introduction
     by Sutton and Barto [<a href="#cliffwalk_ref">1</a>].
+
+    The cliff can be chosen to be slippery (disabled by default) so the player may move perpendicular
+    to the intended direction sometimes (see <a href="#is_slippy">`is_slippery`</a>).
 
     With inspiration from:
     [https://github.com/dennybritz/reinforcement-learning/blob/master/lib/envs/cliff_walking.py](https://github.com/dennybritz/reinforcement-learning/blob/master/lib/envs/cliff_walking.py)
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts:
@CloseChoice Could you update the PR with the project main

CloseChoice:
> @CloseChoice Could you update the PR with the project main

just done that.

pseudo-rnd-thoughts (changes requested):
Thanks for that @CloseChoice 

Is this equivalent to the old version when `is_slippery=False`? 

Would it be possible to include in the `reset` and `step` `info` a `probability` key that is equal to the probability of the outcome occuring?

CloseChoice:
> Thanks for that @CloseChoice
> 
> Is this equivalent to the old version when `is_slippery=False`?
> 
> Would it be possible to include in the `reset` and `step` `info` a `probability` key that is equal to the probability of the outcome occuring?

Yes, old version and `is_slippery=False` are identical. 

Isn't the probability (or `prob`) key already implemented on main, see [here](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/envs/toy_text/cliffwalking.py#L183) and [here](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/envs/toy_text/cliffwalking.py#L192). Or am I missing something here?

pseudo-rnd-thoughts:
I meant `is_slippery=True`

Sorry, I didn't realise that `prob` also existed. Could you add a test that checks that the `prob` == 1 for `is_slippery=False` and < 1 for `is_slippery=True`

Small note: v1 was actually due to the reward threshold being changed (https://github.com/openai/gym/pull/2205 and https://github.com/openai/gym/pull/2315) but kept the comment about `is_slippy` with a note that this was at `v1.0`

CloseChoice:
> I meant `is_slippery=True`
> 
> Sorry, I didn't realise that `prob` also existed. Could you add a test that checks that the `prob` == 1 for `is_slippery=False` and < 1 for `is_slippery=True`
> 
> Small note: v1 was actually due to the reward threshold being changed ([openai/gym#2205](https://github.com/openai/gym/pull/2205) and [openai/gym#2315](https://github.com/openai/gym/pull/2315)) but kept the comment about `is_slippy` with a note that this was at `v1.0`

I think I tested this already, see [here](https://github.com/Farama-Foundation/Gymnasium/pull/1087/files#diff-8cca4b46cf7f212d00f4c9b92ea7aad851362ee52fc5f82fb98b529bddd58127R99). Do you want this more explicitly, or in some other fashion?

Is there a todo for me regarding the version? Somehow I think I did not register v1 properly but I don't know how to do that, would be great if you could point me in the right direction

pseudo-rnd-thoughts:
> I think I tested this already, see [here](https://github.com/Farama-Foundation/Gymnasium/pull/1087/files#diff-8cca4b46cf7f212d00f4c9b92ea7aad851362ee52fc5f82fb98b529bddd58127R99). Do you want this more explicitly, or in some other fashion?

Sorry, I somehow missed that

> suggests that there is a misunderstanding. The implementation on main of this environment is not slippery, meaning each action can just lead to one state (with transition prob 1). But this PR adds the slippery option. To keep things backwards compatible the default for `is_slippery` is `False` which reproduces the current behaviour on main.

Your right, I've updated the title, thanks for clarifying

pseudo-rnd-thoughts (approved):
Looks good to merge

PR #1094, opened 2024-06-27
Add compatibility with numpy 2.0

RedTachyon:
Numpy broke some things with the 2.0 release: https://numpy.org/doc/stable/release/2.0.0-notes.html

The tentative policy is that Gymnasium will support both numpy 1.x and 2.x, at least for the foreseeable future. Of course there will be complications everywhere in downstream projects that have other dependencies, but the changes are minor enough that we should be able to handle it here. There will also be minor incompatibilities here, like conflicts with pytorch, and the fact that numpy 2.0 doesn't support python 3.8. Nothing we can do about that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,8 +18,8 @@ jobs:
       - uses: actions/checkout@v4
       - run: |
           docker build -f bin/all-py.Dockerfile \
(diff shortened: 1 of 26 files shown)


pseudo-rnd-thoughts (approved):
Looks good to me

PR #1099, opened 2024-06-29
[Proposal] Default metadata in BaseMujocoEnv

### LINKED ISSUE #1059: [Proposal] Default metadata in BaseMujocoEnv
### Proposal

The current code for BaseMujocoEnv requires the env metadata dictionary to have fixed, pre-specified values. While this may be useful for future API changes, it doesn't seem very useful at the moment.

------------
        assert self.metadata["render_modes"] == [
            "human",
            "rgb_array",
            "depth_array",
        ], self.metadata["render_modes"]
        if "render_fps" in self.metadata:
            assert (
                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
------------

I propose to set the metadata attribute in BaseMujocoEnv to the current fixed values

-------------
self.metadata["render_modes"] == [
            "human",
            "rgb_array",
            "depth_array",
        ]
self.metadata["render_fps"] = int(np.round(1.0 / self.dt))
-------------

This will enable classes derived from MujocoEnv to overwrite the dictionary (if required, in the future), but it will not force them to write an explicit metadata dictionary if the default values suffice, reducing redundancy and making derived classes more compact and readable.

### Motivation

_No response_

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

spiglerg:
# Description

The current code for BaseMujocoEnv requires the env metadata dictionary to have fixed, pre-specified values. While this may be useful for future API changes, it doesn't seem very useful at the moment.

Fixes #1059 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/envs/mujoco/ant_v4.py
+++ b/gymnasium/envs/mujoco/ant_v4.py
@@ -17,7 +17,7 @@ class AntEnv(MujocoEnv, utils.EzPickle):
             "rgb_array",
             "depth_array",
         ],
-        "render_fps": 20,
+        "render_fps": 30,
     }
 
     def __init__(
--- a/gymnasium/envs/mujoco/ant_v4.py
+++ b/gymnasium/envs/mujoco/ant_v4.py
@@ -17,7 +17,7 @@ class AntEnv(MujocoEnv, utils.EzPickle):
             "rgb_array",
             "depth_array",
         ],
-        "render_fps": 30,
+        "render_fps": 20,
     }
 
     def __init__(
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -88,7 +88,7 @@ class MujocoEnv(gym.Env):
                 "rgb_array",
                 "depth_array",
             ],
-            "render_fps": int(np.round(1.0 / self.dt))
+            "render_fps": int(np.round(1.0 / self.dt)),
         }
 
         if observation_space is not None:
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -90,8 +90,13 @@ class MujocoEnv(gym.Env):
 
         self.frame_skip = frame_skip
 
-        if "render_fps" not in self.metadata:
+        if "render_fps" in self.metadata:
+            assert (
+                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
+            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
+        else:
             self.metadata["render_fps"] = int(np.round(1.0 / self.dt))
+
 
         if observation_space is not None:
             self.observation_space = observation_space
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -37,14 +37,6 @@ def expand_model_path(model_path: str) -> str:
 class MujocoEnv(gym.Env):
     """Superclass for MuJoCo based environments."""
 
-    metadata = {
-        "render_modes": [
-            "human",
-            "rgb_array",
-            "depth_array",
-        ]
-    }
-
     def __init__(
         self,
         model_path: str,
@@ -90,14 +82,14 @@ class MujocoEnv(gym.Env):
 
         self.frame_skip = frame_skip
 
-        if "render_fps" in self.metadata:
-            assert (
-                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
-            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
-        else:
-            # Make a copy of the dictionary to avoid modifying the class variable
-            self.metadata = self.metadata.copy()
-            self.metadata["render_fps"] = int(np.round(1.0 / self.dt))
+        self.metadata = {
+            "render_modes": [
+                "human",
+                "rgb_array",
+                "depth_array",
+            ],
+            "render_fps": int(np.round(1.0 / self.dt))
+        }
 
         if observation_space is not None:
             self.observation_space = observation_space
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -82,15 +82,13 @@ class MujocoEnv(gym.Env):
 
         self.frame_skip = frame_skip
 
-        assert self.metadata["render_modes"] == [
-            "human",
-            "rgb_array",
-            "depth_array",
-        ], self.metadata["render_modes"]
-        if "render_fps" in self.metadata:
-            assert (
-                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
-            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
+        self.metadata["render_modes"] = [
+                    "human",
+                    "rgb_array",
+                    "depth_array",
+                ]
+        self.metadata["render_fps"] = int(np.round(1.0 / self.dt))
+
         if observation_space is not None:
             self.observation_space = observation_space
         self._set_action_space()
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -37,6 +37,14 @@ def expand_model_path(model_path: str) -> str:
 class MujocoEnv(gym.Env):
     """Superclass for MuJoCo based environments."""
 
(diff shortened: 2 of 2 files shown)


Kallinteris-Andreas:
i will be inactive till 11/07/2024, after which i will review it

gymnasium/envs/mujoco/ant_v4.py
```diff
@@ -17,7 +17,7 @@ class AntEnv(MujocoEnv, utils.EzPickle):
             "rgb_array",
             "depth_array",
         ],
-        "render_fps": 20,
```
Kallinteris-Andreas, line 20: why is this changed

gymnasium/envs/mujoco/mujoco_env.py
```diff
@@ -82,15 +90,9 @@ def __init__(
...
-        assert self.metadata["render_modes"] == [
-            "human",
-            "rgb_array",
-            "depth_array",
-        ], self.metadata["render_modes"]
-        if "render_fps" in self.metadata:
-            assert (
-                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
-            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
+        if "render_fps" not in self.metadata:
```
Kallinteris-Andreas, line 93: Here is the functionality for verifying that `render_fps` equals the expected value is removed (for v4 or older based environments that define the  `render_fps` themselfs)

Kallinteris-Andreas:
From my testing, everything appears to work
```py
import gymnasium
from gymnasium.envs.mujoco.mujoco_env import MujocoEnv
assert MujocoEnv.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array']}

env0 = gymnasium.make("Ant-v5")
assert MujocoEnv.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array']}
assert env0.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 20}

env1 = gymnasium.make("Humanoid-v5")
assert MujocoEnv.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array']}
assert env0.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 20}
assert env1.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 67}

env2 = gymnasium.make("Hopper-v4")
assert MujocoEnv.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array']}
assert env0.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 20}
assert env1.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 67}
assert env2.metadata == {'render_modes': ['human', 'rgb_array', 'depth_array'], "render_fps": 125}
```

gymnasium/envs/mujoco/ant_v4.py
spiglerg: I calculate that dynamically in the constructor, since it depends on self.dt and before it was forcing it to always have this value anyway:

Previous code:

assert (
                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'

gymnasium/envs/mujoco/mujoco_env.py
spiglerg: Hmm right, but if render_fps is in metadata and is different than int(np.round(1.0 / self.dt)), the assertion fails. We can maintain the previous behavior by always overriding render_fps with the computed value for the moment then, perhaps?

Kallinteris-Andreas: No, it should fail if `render_fps` is explicitly defined to something else, otherwise render issues will arise

spiglerg: Hmm ok!
Then I can modify it to set this computed value *if render_fps is not in the metadata*, and run the assertion if it is?

Kallinteris-Andreas: yes

spiglerg: Should I also update the individual Mujoco environments to remove the metadata -where it matches the new automatically generated one- ?

Kallinteris-Andreas: Just leave all the v4 and older environment as is

spiglerg: ok!

Kallinteris-Andreas:
run `pre-commit run --all-files` check contribute.md for clarification

spiglerg:
There was a problem with a line in the file not committed. Now pre-commit tests run locally.

spiglerg:
Apologies; the last problem was due to `frames_skip' afftecting self.dt. I am fixing and running tests locally. I will push a fix as soon as possible.

gymnasium/envs/mujoco/mujoco_env.py
```diff
@@ -82,15 +90,15 @@ def __init__(
...
-            "rgb_array",
-            "depth_array",
-        ], self.metadata["render_modes"]
         if "render_fps" in self.metadata:
             assert (
                 int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
             ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
+        else:
+            # Make a copy of the dictionary to avoid modifying the class variable
+            self.metadata = self.metadata.copy()
```
Kallinteris-Andreas, line 99: if we need, wouldn't be easier to do
```py
    metadata = {
        "render_modes": [
            "human",
            "rgb_array",
            "depth_array",
        ],
    "render_fps": int(np.round(1.0 / self.dt))
    }
    ```

Comment (spiglerg) [reply gymnasium/envs/mujoco/mujoco_env.py:99]: Yes, that was my original intention. ;)
However, if metadata is initialized as self.metadata in the constructor (as in this approach), but a derived class still implements the older way with metadata as static class variable, the outcome may be unexpected for the user (hence the use of assertions in the latest code, in case for some reason a mismatch happens).

I can test that and add a pytest to check for correct behavior in case a derived class re-defines the metadata as static (sub)class variable, as is currently done.
In any case, the current main repository does not allow for any derived class to have any different metadata, so any metadata written explicitly by a user (for a new derived class) must be identical to this value).

It would also help to remove all 'metadata' from the existing Mujoco environments to improve clarity, and indeed only define it once in the base class. Or perhaps to change them from class variables to object variables initialized in the constructor (i.e. "render_fps" can be always removed, since there is only one valid value;  "render_modes" can be set to replace the default array, e.g.

self.metadata["render_modes"] = [new-modes]

 eploiting the fact that the default metadata would always contain the two keys, "render_modes" and "render_fps").

Comment (spiglerg) [reply gymnasium/envs/mujoco/mujoco_env.py:99]: ** This applying to MujocoEnv and derived classes. 
gym.Env also (ab)uses static class variables, including metadata (although it is not strict like MujocoEnv in the choice of "render_fps") ?

In that case, if metadata is kept a class variable as in gym.Env, then dynamic copying (metadata.copy()) and updating of "render_fps" the constructor may be the only alternative.

Comment (spiglerg) [reply gymnasium/envs/mujoco/mujoco_env.py:99]: A possible (although ugly) modification may be as follows:

```
        # Back-compatibility: this only applies for derived classes that define `metadata`
        # as class variable, instead of modifying it in their constructor.
        if hasattr(self, 'metadata') and ("render_fps" in self.metadata):
            assert (
                int(np.round(1.0 / self.dt)) == self.metadata["render_fps"]
            ), f'Expected value: {int(np.round(1.0 / self.dt))}, Actual value: {self.metadata["render_fps"]}'
        self.metadata = {
            "render_modes": [
                "human",
                "rgb_array",
                "depth_array",
            ],
            "render_fps": int(np.round(1.0 / self.dt))
        }
```

However, since the current behavior is to terminate if the "render_fps" of the derived class isn't correct (for Mujoco environments), any existing environment should already satisfy the assertion.

If the problem is users that to (incorrectly) change it in the derived class, perhaps we could add a check at renderer-level?

spiglerg: By the way, "render_fps" is only used either explicitly by a few environments that handle their own rendering, or as argument to be explicitly passed by the user to the `save_video' utility function.

It does not seem to be a problem to just set the correct value (for Mujoco envs) without checks on later modifications by the users. Perhaps we can add a comment in the code to explain that Mujoco environments should leave the default "render_fps" as is, since there is no use in ever modifying it.

pseudo-rnd-thoughts:
@Kallinteris-Andreas What is happening with this PR?

PR #1102, opened 2024-07-01
Add missing `NotImplementedError` for `VectorEnv.step` 

jjshoots:
# Description

Fixes some broken typehints.

### DIFF
--- a/gymnasium/vector/vector_env.py
+++ b/gymnasium/vector/vector_env.py
@@ -127,7 +127,7 @@ class VectorEnv(Generic[ObsType, ActType, ArrayType]):
     def reset(
         self,
         *,
-        seed: int | list[int] | None = None,
+        seed: int | None = None,
         options: dict[str, Any] | None = None,
     ) -> tuple[ObsType, dict[str, Any]]:  # type: ignore
         """Reset all parallel environments and return a batch of initial observations and info.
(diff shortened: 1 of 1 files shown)


gymnasium/vector/vector_env.py
```diff
@@ -127,7 +127,7 @@ class VectorEnv(Generic[ObsType, ActType, ArrayType]):
     def reset(
         self,
         *,
-        seed: int | None = None,
+        seed: int | list[int] | None = None,
```
pseudo-rnd-thoughts, line 130: This was an intentional decision that is overwritten by the `AsyncVectorEnv` and `SyncVectorEnv` 
but not for `VectorEnv`

gymnasium/vector/vector_env.py
```diff
@@ -190,6 +192,7 @@ def step(
             >>> infos
             {}
         """
+        raise NotImplementedError(f"{self.__str__()} step function is not implemented.")
```
pseudo-rnd-thoughts, line 195: This does seem helpful to include

jjshoots: I'm not sure how the formatter didn't catch this function without a body. Is that normal?

gymnasium/vector/vector_env.py
jjshoots: Ah gotcha, will revert.

gymnasium/vector/vector_env.py
pseudo-rnd-thoughts: It is because the docstring counts as a function component

gymnasium/vector/vector_env.py
pseudo-rnd-thoughts: Thanks

pseudo-rnd-thoughts (approved):

PR #1107, opened 2024-07-02
Fix non-used device argument in `to_torch` conversion

### LINKED ISSUE #1106: [Bug Report] `NumpyToTorch` wrapper ignores device upon reset
## Bug
[NumpyToTorch](https://github.com/Farama-Foundation/Gymnasium/blob/57136fbaeec340c023ae6939348ac2e7cfe3b3cb/gymnasium/wrappers/numpy_to_torch.py#L102) wrapper ignores `device` when the environment is reset. I created a PR that should fix it.

## Feature request
I would like to further ask if it was possible to change the wrapper's behavior to return tensorized `reward`, `truncated`, `terminated` for non-vectorized environments (or at least add an option to be able to conditionally return their tensor versions)? This is because it is more consistent and more convenient to handle such inputs, for example, when creating tensor buffers:

```python
for i, env_out in enumerate(env_outs[:-1]):
    buffer[i] = torch.cat([buffer[i], env_out.unsqueeze(0)])
```
I tried applying `vectors.numpy_to_torch.NumpyToTorch` wrapper on a non-vector environment which would do the trick, however, there is an assertion that requires the environment to be necessarily a vector env.

### Code example

```shell
import gymnasium as gym
from gymnasium.wrappers.numpy_to_torch import NumpyToTorch

env = NumpyToTorch(gym.make("CartPole-v1"), device="cuda")
print(env.reset()[0].device) # device(type='cpu')
```

### System info

`python=3.11`, `gymnasium=1.0.0a2`

### Additional context

_No response_

### Checklist

mantasu:
# Description

A small fix that propagates the `device` argument when converting from an _Iterable_ of `numpy` or `jax` objects to an _Iterable_ or `torch.Tensor`.

Fixes #1106 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/jax_to_torch.py
+++ b/gymnasium/wrappers/jax_to_torch.py
@@ -119,9 +119,9 @@ def _jax_iterable_to_torch(
     if hasattr(value, "_make"):
         # namedtuple - underline used to prevent potential name conflicts
         # noinspection PyProtectedMember
-        return type(value)._make(jax_to_torch(v) for v in value)
+        return type(value)._make(jax_to_torch(v, device) for v in value)
     else:
-        return type(value)(jax_to_torch(v) for v in value)
+        return type(value)(jax_to_torch(v, device) for v in value)
 
 
 class JaxToTorch(gym.Wrapper, gym.utils.RecordConstructorArgs):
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, sadly I don't think we can test this as much as we would wish as the CI server only has a CPU but might be worth adding tests in the future that are ignored by the device unless a gpu is connected

PR #1112, opened 2024-07-07
Fix some small bugs in gymnasium.vector

xihuai18:
# Description

## Type of change

### Screenshots

- Avoid unnecessary reset in sync_vector_env.py
- Fix some typos and type hints in async_vector_env.py

# Checklist:

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -9,7 +9,7 @@ from copy import deepcopy
 from enum import Enum
 from multiprocessing import Queue
 from multiprocessing.connection import Connection
-from typing import Any, Callable, Sequence
+from typing import Any, Callable, Sequence, Dict, Tuple
 
 import numpy as np
 
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -648,7 +648,7 @@ def _async_worker(
     env_fn: callable,
     pipe: Connection,
     parent_pipe: Connection,
-    shared_memory: bool,
+    shared_memory: multiprocessing.Array | Dict[str, Any] | Tuple[Any, ...],
     error_queue: Queue,
 ):
     env = env_fn()
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -697,7 +697,7 @@ def _async_worker(
                 break
             elif command == "_call":
                 name, args, kwargs = data
-                if name in ["reset", "step", "close", "set_wrapper_attr"]:
+                if name in ["reset", "step", "close", "_setattr", "_check_spaces"]:
                     raise ValueError(
                         f"Trying to call function `{name}` with `call`, use `{name}` directly instead."
                     )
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Please could you explain the PR, I don't understand the change

xihuai18:
I think `self._autoreset_envs` records whether the sub-environments should be reset in the next step. However, after the `reset` method of SyncVectorEnv instances is manually called, `self._autoreset_envs` is not set to all `False`, leading to some sub-environments will be *reset twice* consecutively.

I also noticed that in `_aysnc_worker` of `async-vector_env.py`, the `autoreset` flag is set to `False` if the `reset` method is manually called.

So I think `self._autoreset_envs` should be set to all `False` at the end of the `reset` method of SyncVectorEnv.

If I am wrong, ignore this PR.

xihuai18:
In addition, there may be some typos in `_async_worker` of `async_vector_env.py` and I fix them.

pseudo-rnd-thoughts (changes requested):
Could you add a test for these features to show them as an issue

pseudo-rnd-thoughts:
@xihuai18 Your correct about the autoreset for `SyncVectorEnv` however the other two changes I don't believe are correct. 
Could you revert those changes then it should be good to merge

xihuai18:
I'd like to explain my changes:
1. exceptions for `_call`
https://github.com/Farama-Foundation/Gymnasium/blob/1878f6dd7d69439dc770289a9000eed6a52b4a43/gymnasium/vector/async_vector_env.py#L700-L726

Line 725 shows that all accepted commands are `reset`, `step`, `close`, `_call`, `_setattr`, `_check_spaces`, so corresponding exceptions are changed in Line 702. For example, one can not `_call` `_setattr`, but should use the `_setattr` method directly.

2. type hints for `shared_memory`
https://github.com/Farama-Foundation/Gymnasium/blob/1878f6dd7d69439dc770289a9000eed6a52b4a43/gymnasium/vector/async_vector_env.py#L669-L672

`shared_memory` are used for writing in the obtained obserations via `write_to_shared_memory` function (if it is not `None`). Thus it should be a `multiprocessing.Array` or other nested types, as defined in https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/vector/utils/shared_memory.py.

xihuai18:
I have passed the pre-commit now.

pseudo-rnd-thoughts:
Ahh you are right on `shared_memory`, I had missed the line below and thought it must have been a boolean

For the error, it is around the function / attr being called not the command. The list you have updated is the list of commands but not if the function should be called. 
Saying that, I think the list should be `"reset", "step", "close"` and with `"set_wrapper_attr"` users should just use `"set_attr"` the error doesn't support a different function name.

xihuai18:
> Ahh you are right on `shared_memory`, I had missed the line below and thought it must have been a boolean
> 
> For the error, it is for the function / attr name not the command, i.e., `envs.call("reset")` is disabled. The list you have updated is the list of commands but not if the function should be called. Saying that, I think the list should be `"reset", "step", "close"` and with `"set_wrapper_attr"` users should just use `"set_attr"` the error doesn't support a different function name.

I think the current version is right according to your review?

pseudo-rnd-thoughts (approved):

PR #1117, opened 2024-07-12
naming returned values from step()

yhirshhorn:
I've added a suggestion to name the values returned from step(), since whenever i try to learn gym I have tons of troubles with this method and it's returned values. The documentation at gymnasium.farama.org is good, but I still manage to get confused and hit walls. Besides, I believe self documenting code is way superior to documentation. I just don't know if there is any penalty to naming the values like I suggest here. If there is no, or if it is a negligible penalty, then will you please consider this for the sake of other head-bangers such as myself?


Kallinteris-Andreas:
The API is standardized, and is used by 1000s of RL projects, it can not be changed.

Not that this would be a good idea regardless of that, 

the intended way to use the step function is this:
```py
observation, reward, terminated, truncated, info = env.step(action)
```
for more info, refer to the documentation https://gymnasium.farama.org/main/introduction/basic_usage/

PR #1126, opened 2024-07-23
Add (A)syncVectorEnv support for sub-envs with different obs spaces

reginald-mclean:
# Description

In order to add Meta-World environments to use the Sync/Async Gymnasium methods, a change needed to be made in order to batch observations that have different low/high values.

This change allows the user to specify one of 3 modes for batching observations:
- 'same': current method of having the same observation space for all environments
- 'different': method that allows for different observation spaces to be batched together, given the observation spaces are the same size 
- gym.Space object: allows the user to input a specific method not covered by the 2 above modes

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -159,7 +159,7 @@ class AsyncVectorEnv(VectorEnv):
                     [env.observation_space for env in self.env_fns]
                 )
             else:
-                raise ValueError
+                raise ValueError("Need to pass in mode for batching observations")
         self.action_space = batch_space(self.single_action_space, self.num_envs)
 
         dummy_env.close()
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -14,7 +14,7 @@ from typing import Any, Callable, Sequence
 
 import numpy as np
 
-from gymnasium import logger
+from gymnasium import logger, Space
 from gymnasium.core import ActType, Env, ObsType, RenderFrame
 from gymnasium.error import (
     AlreadyPendingCallError,
@@ -98,6 +98,7 @@ class AsyncVectorEnv(VectorEnv):
             ]
             | None
         ) = None,
+        observation_mode: str or Space = 'same',
     ):
         """Vectorized environment that runs multiple environments in parallel.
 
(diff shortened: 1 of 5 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for the PR @reginald-mclean 

Could you add explicit testing of different and custom observation space. I recommend using the GenericTestEnv for testing with different obs spaces

Second, we need to think a bit more about what single_observation_space means as batch space of the single might not be equivalent to the full obs space. How should users interpret single in relation to full? 

This might not be a big issue compared to single action space which users commonly sample for random actions, single obs space is never sampled in the same way

For custom space, I'm tempted to add testing (e.g. can sub env obs samples be concatenated together) but if users use it then it is their risk and they should get appropriate errors we hope

gymnasium/vector/utils/batched_spaces.py
```diff
@@ -21,7 +21,7 @@
 
 
 @singledispatch
-def batch_differing_spaces(spaces: list[Space]):
+def batch_differing_spaces(spaces: "list[Space]"):
```
pseudo-rnd-thoughts, line 24: You can revert this change with `from __future__ import annotation`

gymnasium/vector/sync_vector_env.py
```diff
@@ -100,6 +102,8 @@ def __init__(
                 self.observation_space = batch_differing_spaces(
                     [env.observation_space for env in self.envs]
                 )
+            else:
+                raise ValueError
```
pseudo-rnd-thoughts, line 106: Add message

gymnasium/vector/async_vector_env.py
```diff
@@ -155,7 +158,8 @@ def __init__(
                 self.observation_space = batch_differing_spaces(
                     [env.observation_space for env in self.env_fns]
                 )
-
+            else:
+                raise ValueError
```
pseudo-rnd-thoughts, line 162: Add error message

pseudo-rnd-thoughts (changes requested):

pseudo-rnd-thoughts (changes requested):
Could you add testing of the feature

pseudo-rnd-thoughts:
I have a PR to add to this, the easiest way of doing this is to merge this PR into a different branch, push my changes to the branch then merge the batch back into main

pseudo-rnd-thoughts (approved):

Issue #1122, opened 2024-07-18
[To investigate] Check that the new autoreset order works correctly in vectorized jax envs


RedTachyon:
I don't have time to debug properly right now, but I'm running some experiments with 1.0.0a2 and vectorized `phys2d/Pendulum-v0`.

The loop is roughly:

```python
for t in trange(200 * 5):

    actions = ...

    next_obs, reward, term, trunc, info = env.step(actions)

    done = np.logical_or(term, trunc).any()

    replay_buffer.add(obs, actions[:, jnp.newaxis], next_obs)


    obs = next_obs
    print(t, done)

    # This is the questionable code
    if done:
        obs, *_ = env.step(actions)
        obs, *_ = env.step(actions)
```

The time limit is set to 200, so I'm trying to set it up so that I get a consistent 200 samples for each episode. 
So steps 0-199 (inclusive) should be from episode 1, steps 200-399 from episode 2, and so on.

There's no termination condition, so all sub-environments will truncate in sync.

In the "questionable" part, if I don't have any additional `step`s, I get two consecutive steps with `done==True`. With one additional `step`, I get a single `done==True`, but there seems to be a misalignment after the first episode, since `done`s occur at steps 199, 400, 601, etc.

When I switch to a sync vector env implementation of the regular Pendulum, I get the "aligned" behavior with only one extra `step` (as opposed to two in the code above), which is consistent with my understanding of how the new autoreset order should work. So it's likely that FunctionalJaxVectorEnv isn't properly adapted. Quite possibly my fault, as off-by-one-errors are still my personal nemesis.

amiii123malviya:
Hello,

Is this issue still active?
I would like to work on this.

Best Regards.

pseudo-rnd-thoughts:
Hey @amiii123malviya, #1130 should have fixed this

amiii123malviya:
Okay, Thankyou so much for your time .

Best Regards

PR #1130, opened 2024-07-27
Fix the off-by-one error in FunctionalJaxVectorEnv

RedTachyon:
See #1122 

FunctionalJaxVectorEnv wasn't adapted to the new autoreset order

### DIFF
--- a/gymnasium/envs/functional_jax_env.py
+++ b/gymnasium/envs/functional_jax_env.py
@@ -137,7 +137,7 @@ class FunctionalJaxVectorEnv(gym.vector.VectorEnv):
 
         self.steps = jnp.zeros(self.num_envs, dtype=jnp.int32)
 
-        self.autoreset_envs = jnp.zeros(self.num_envs, dtype=jnp.bool_)
+        self.prev_done = jnp.zeros(self.num_envs, dtype=jnp.bool_)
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
Quick comment on the test, could you seed the action space and should it not be either both truncate on the same step or neither do. It only checks if one of them do.
Can we add a test for terminate as well truncate?
Otherwise I think the auto reset is now correct, thanks for fixing it. 
It would be helpful to have a check to confirm implementation between (a)sync vector env and jax vector env

RedTachyon:
> seed the action space

Done

> sync/async

Done


> termination vs truncation

This test specifically checks this one failure case on the simplest scenario, i.e. a fixed-length episode which always truncates after 200 steps. Pendulum never terminates, so `term | trunc == trunc`. There's probably something to be said in terms of the code clarity vs redundancy, but that's a fairly subtle tradeoff. A more general case would need some more nuanced handling, but I estimate it would be significantly more complex, to the point where it's probably not worth the effort at this moment.

pseudo-rnd-thoughts (approved):
Thanks for the changes looks good to me. 
I didn't realise that Pendulum doesn't have a termination case.

PR #1175, opened 2024-09-25
add `python==3.13` and drop `python==3.8`

Kallinteris-Andreas:


README.md
```diff
@@ -28,7 +28,7 @@ To install the base Gymnasium library, use `pip install gymnasium`
 
 This does not include dependencies for all families of environments (there's a massive number, and some can be problematic to install on certain systems). You can install these dependencies for one family like `pip install "gymnasium[atari]"` or use `pip install "gymnasium[all]"` to install all dependencies.
 
-We support and test for Python 3.8, 3.9, 3.10, 3.11 and 3.12 on Linux and macOS. We will accept PRs related to Windows, but do not officially support it.
```
Kallinteris-Andreas, line 31: This was removed because we have a badge at the top of the README that contains the information, can be reverted if you do not want it

RedTachyon:
Is there anything forcing us to break compatibility with 3.8? My general intuition is that we can keep it around for now, and once we need any features that were added in 3.9, we drop it without much worry in the next release

Kallinteris-Andreas:
Generally we drop support for python versions once they become unsupported by the python developers.

RedTachyon:
This isn't some dogma that we must always follow. My point is a bit more from first principles - there's a difference between actively and passively supporting some version. 
With 1.0, we will actively support python>=3.9,<=3.13, as in we make sure that Gymnasium works with each of those versions.
On the other hand, I want to avoid deliberately "unsupporting" 3.8, i.e. we passively support 3.8 for as long as it makes sense. 

FWIW I just checked and I'm pretty sure this PR actually does it the way I'm proposing: 

https://github.com/Farama-Foundation/Gymnasium/blob/50bf685cf0ccf506e67ebd0bd2144ec6fef46774/pyproject.toml#L11


If we *really* want to drop 3.8, we'd bump this to 3.9, so that users cannot (without some weird workarounds) install gymnasium on 3.8. I don't think that makes sense to do if we know that Gymnasium actually runs alright on 3.8.

On the other hand, if we want to keep this "passive" support, removing it from the test suite makes it... awkward. Let's say in 1.1.0 we want to add a feature which uses, say, dict merging `{1: 2} | {3: 4}`, which was added in 3.9. We shouldn't hesitate to add this - but at that point we should also stop supporting 3.8, even passively, i.e. we actively mark it as incompatible by changing the aforementioned line in pyproject.toml. But if we don't have 3.8 in the test suite, then we won't know when this happens, so we might end up in a weird state where someone (for whatever reason) using 3.8, by default installs a version of Gymnasium that doesn't actually work.

All this to say - we should add 3.13 to the test matrix, and keep a policy that 3.8 compatibility can be broken, but consciously. There is no reason that I see to actually drop support right now, so let's keep it to maintain wider compatibility. We definitely shouldn't do anything halfway, i.e. allow it to be installed but untested.

pyproject.toml
```diff
@@ -13,14 +13,14 @@ authors = [{ name = "Farama Foundation", email = "contact@farama.org" }]
 license = { text = "MIT License" }
 keywords = ["Reinforcement Learning", "game", "RL", "AI", "gymnasium"]
 classifiers = [
-    "Development Status :: 4 - Beta",  # change to `5 - Production/Stable` when ready
```
Kallinteris-Andreas, line 16: this is done because 1.0 release

Kallinteris-Andreas:
Most libraries such as `numpy` also stop supporting python versions as they become unsupported, so some issues may be caused long term, I am not opposed to making an exception for `1.0` specifically to have 3.8 supported, also depends on when 1.0 is actually being released

pseudo-rnd-thoughts:
This would break what @RedTachyon is suggesting, but if we completely dropping 3.8 then we should update pyupgrade in pre-commit to `--py39-plus` which causes a significant number of changes, mainly reducing the number of type imports and to `pyright` in `pyproject.toml` with `pythonVersion="3.9"`

pseudo-rnd-thoughts:
Python 3.13 has been delayed by a week due to unexpected performance regressions fyi

Kallinteris-Andreas:
Waiting for `pytorch` as usual

pseudo-rnd-thoughts:
Mujoco `3.2.5` doesn't support 3.13 yet - https://github.com/google-deepmind/mujoco/issues/2232

h-vetinari:
You're also going to have to move to Cython 3.0, because that's the minimum version necessary for CPython 3.13 support, see
https://github.com/Farama-Foundation/Gymnasium/blob/13230f4a120336b45c880b644585056d31f4c956/pyproject.toml#L42
https://github.com/Farama-Foundation/Gymnasium/blob/13230f4a120336b45c880b644585056d31f4c956/pyproject.toml#L63

saran-t:
MuJoCo 3.2.6 has prebuilt wheels for Python 3.13.

pseudo-rnd-thoughts:
Thanks @saran-t 

We are still waiting on TensorStore that flax uses - https://github.com/google/tensorstore/issues/204
@Kallinteris-Andreas Could we test without the missing modules just to check that everything works. If we can, then we could cut a release with Python 3.13 and specify that this might not work for users that use modules X, Y, Z

amacati:
Is there anything still blocking 3.13 adoption? Tensorstore now has a wheel for 3.13 (see [here](https://github.com/google/tensorstore/commit/69d7d0573b4b093e60f41d1d57cf3a3a9039d6ea)) and Python 3.8 (and 3.9) support was dropped in #1333 for other reasons.

amacati:
Is mujoco-py blocking the transition because of incompatibility issues?

PR #1176, opened 2024-09-26
CI test

Kallinteris-Andreas:


pseudo-rnd-thoughts:
I've made several PRs around the doctest https://github.com/Farama-Foundation/Gymnasium/pull/1144 and https://github.com/Farama-Foundation/Gymnasium/pull/1153 reverting it back.
I'm confused, which style does it want us to use

pseudo-rnd-thoughts:
@Kallinteris-Andreas you pushes to the wrong or

PR #1154, opened 2024-09-03
add FlyCraft to third party environments

GongXudong:
# Description

Add a fixed-wing UAV environment FlyCraft to the third party environments. 

## Type of change

- [x] Documentation only change (no code changed)

PR #1180, opened 2024-09-28
Revert #1154 (Fix NumPy DocTests)

pseudo-rnd-thoughts:
# Description

Reverts #1153 which had reverted #1144, why the NumPy string formats are not working is strange and unexplained as I don't believe it is directly related to NumPy releases


Kallinteris-Andreas:
Hello Darkness, my old friend

Kallinteris-Andreas:
ok, so it appears to pass CI for `numpy<2` but fail for `numpy>=2` (from what i can tell)


image

pseudo-rnd-thoughts:
Ohh, doctest is disabled for 3.8 which is why it passed. 

The reason I closed the PR is that reverted the changes, the PR failed on all the changes. 
So I reran the main CI that had just failed on doctest and it passed just two days without any changes. 

This really is a mystery to understand what is causing the problem.

Issue #1118, opened 2024-07-13
[Proposal] Allow to specify dtype for Discrete 


younik:
### Proposal

Add dtype argument to spaces.Discrete (similar to MultiDiscrete and Box).

### Motivation

Currently the dtype is fixed to numpy.int64. However, often Discrete spaces are much smaller, resulting in a waste of memory.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/Gymnasium/issues) in the repo

pseudo-rnd-thoughts:
Sure, if you can update all the functions and update the example testing spaces with dtypes

JDRanpariya:
I would like to work on the issue.

Can you guide me on what other functions needs to be updated if we default it to int8 or int32 in constructor? 

With example testing spaces do you mean tests/spaces/test_discrete.py? There is no explicit dtype checks in multidiscrete test, can you tell what needs to be done there?

greyentity101:
This feature has already been implemented and merged upstream — `Discrete` now accepts a `dtype` argument (default `np.int64`), matching `MultiDiscrete` and `Box`.

- Implemented in #1467 (*Add dtype parameter to Discrete*), with follow-ups fixing `contains` for small dtypes (#1616, #1648).
- Covered by the parametrized `test_dtype` in `tests/spaces/test_discrete.py` and a dtype example in `tests/spaces/utils.py` (e.g. `Discrete(n=4, dtype=np.int32, start=1)`).

Example:
```python
>>> from gymnasium.spaces import Discrete
>>> space = Discrete(3, dtype=np.int8)
>>> space.sample().dtype
dtype('int8')
```

Suggest closing this issue.

PR #1197, opened 2024-10-06
Add `Discrete(dtype)` parameter

JDRanpariya:
# Description

Add support for dtype in Discrete Space.

Fixes # (issue)
#1118 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -37,7 +37,7 @@ class Discrete(Space[np.int64]):
 
         Args:
             n (int): The number of elements of this space.
-            dtype: This should be some kind of integer type.
+            dtype: The space type, for example, ``int``, ``np.int64``, ``np.int32``, or ``np.uint8``.
             seed: Optionally, you can use this argument to seed the RNG that is used to sample from the ``Dict`` space.
             start (int): The smallest element of this space.
         """
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -27,6 +27,7 @@ class Discrete(Space[np.int64]):
     def __init__(
         self,
         n: int | np.integer[Any],
+        dtype: str | type[np.integer[Any]] = np.int64,
         seed: int | np.random.Generator | None = None,
         start: int | np.integer[Any] = 0,
     ):
@@ -36,6 +37,7 @@ class Discrete(Space[np.int64]):
 
         Args:
             n (int): The number of elements of this space.
+            dtype: This should be some kind of integer type.
             seed: Optionally, you can use this argument to seed the RNG that is used to sample from the ``Dict`` space.
             start (int): The smallest element of this space.
         """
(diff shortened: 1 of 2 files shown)


gymnasium/spaces/discrete.py
```diff
@@ -84,13 +99,13 @@ def sample(self, mask: MaskNDArray | None = None) -> np.int64:
                 np.logical_or(mask == 0, valid_action_mask)
             ), f"All values of a mask should be 0 or 1, actual values: {mask}"
             if np.any(valid_action_mask):
-                return self.start + self.np_random.choice(
```
pseudo-rnd-thoughts, line 87: You should be able to pass a type to `choice` or use `asdtype(self.dtype)`

gymnasium/spaces/discrete.py
```diff
@@ -84,13 +99,13 @@ def sample(self, mask: MaskNDArray | None = None) -> np.int64:
...
-                    np.where(valid_action_mask)[0]
+                return self.start + self.dtype.type(
+                    self.np_random.choice(np.where(valid_action_mask)[0])
                 )
             else:
                 return self.start
 
-        return self.start + self.np_random.integers(self.n)
+        return self.start + self.dtype.type(self.np_random.integers(self.n))
```
pseudo-rnd-thoughts, line 108: See comment above

pseudo-rnd-thoughts (changes requested):
Could you move the changes then good to go

gymnasium/spaces/discrete.py
JDRanpariya: unfortunately that returns an integer and it doesn't have `asdtype` or `astype`.

gymnasium/spaces/discrete.py
JDRanpariya: Here I was able to use `.astype`

gymnasium/spaces/discrete.py
JDRanpariya: Let me know if you have any other idea, otherwise would have to go with `self.dtype.type()`

pseudo-rnd-thoughts (changes requested):
I forgot to request testing,

```python
@pytest.mark.parametrize("dtype, sample_dtype", [
    (int, np.int64),
    (np.int64, np.int64),
    (np.int32, np.int32),
    (np.uint8, np.uint8),
])
def test_dtype(dtype, sample_dtype):
    space = Discrete(n=5, dtype=dtype, start=2)

    sample = space.sample()
    sample_mask = space.sample(mask=np.array([0, 1, 0, 0, 0], dtype=np.int8))
    print(f'{sample=}, {sample_mask=}')
    print(f'{type(sample)=}, {type(sample_mask)=}')
    assert isinstance(sample, sample_dtype), type(sample)
    assert isinstance(sample_mask, sample_dtype), type(sample_mask)


@pytest.mark.parametrize("dtype", [
    str,
    np.float32,
    np.complex64,
])
def test_dtype_error(dtype):
    with pytest.raises(ValueError, match="Invalid Discrete dtype"):
        Discrete(4, dtype=dtype)
```
Could you add this to `tests/spaces/test_discrete.py`

gymnasium/spaces/discrete.py
pseudo-rnd-thoughts: no worries, `self.dtype.type` is fine

JDRanpariya (commented):
Have added the tests, thanks!

tests/spaces/test_discrete.py
```diff
@@ -32,3 +33,36 @@ def test_sample_mask():
...
+
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        str,
+        np.float32,
+        np.complex64,
+    ],
+)
+def test_dtype_error(dtype):
```
Kallinteris-Andreas, line 66: Also Test dtype equals none

Kallinteris-Andreas: Yes, feel free to create a PR for multi-discrete testing next.

pseudo-rnd-thoughts (changes requested):
I just realised that we need to update the `contains` to work with other dtype.
@JDRanpariya Could you update `contains` and add to `test_dtype`, checking that the samples are contained within the space. 
It might be worth checking if failure cases as well including `int` for `np.int64` dtype and empty shape arrays

JDRanpariya:
Hmm, there is no test for `contains', 'to_jsonable` and 'from_jsonable`, would have to know the expected behavior and write tests for all three of them. 

Would look into it, shortly. I would like to spend sometime looking at whole code base for a while.

KumarAnmol1234 (approved):

PR #1204, opened 2024-10-11
Use `pipx` for `pre-commit` and `PyPi` in CI

Kallinteris-Andreas:
Due to changes to Ubuntu we have to use `pipx` instead of `pip` (https://peps.python.org/pep-0668/)

Fixes https://github.com/Farama-Foundation/Gymnasium/pull/1203

From testing, the following does not work:
```yaml
      - run: pipx install pre-commit
      - run: pre_commit --version
      - run: pre_commit run --all-files
```
or
```yaml
      - run: python -m venv test_env
      - run: source test_env/bin/activate
      - run: python -m pre_commit --version
      - run: python -m pre_commit run --all-files
```
or
```yaml
      - run: pipx install pre-commit
      - run: pipx run pre_commit --version
      - run: pipx run pre_commit run --all-files
```


pseudo-rnd-thoughts (approved):
Ok, this is going to require all farama projects to update I believe. 
Thanks for sorting this out @Kallinteris-Andreas

PR #1206, opened 2024-10-12
documentation: replace "Jax" with "NumPy" and make capitalization uniform

enjoh:
# Description

Replaces all the leftover "Jax" with "NumPy" and makes capitalization of NumPy uniform within the file.

## Type of change


pseudo-rnd-thoughts (approved):
Thanks for spotting those

PR #1211, opened 2024-10-16
Add Beartype as partial replacement of PyRight

pseudo-rnd-thoughts:
# Description

This is purely an exploratory PR to see if [beartype](https://github.com/beartype/beartype), a runtime type checker, (don't worry we only run this at test time) helps identify type problems more effectively than PyRight


.github/workflows/run-pytest.yml
```diff
@@ -5,38 +5,54 @@ permissions:
...
         numpy-version: ['>=1.21,<2.0', '>=2.0']
-        exclude:
-          - python-version: '3.8'  # numpy>=2.0 requires Python>=3.9
-            numpy-version: '>=2.0'
     steps:
       - uses: actions/checkout@v4
       - run: |
           docker build -f bin/all-py.Dockerfile \
-            --build-arg PYTHON_VERSION="${{ matrix.python-version }}" \
+            --build-arg PYTHON_VERSION="3.12" \
```
Kallinteris-Andreas, line 54: You are fixing some issues on python 3.12, But those changes might be breaking for Python 3.8, 

Typings should be tested on the lowest supported version.

PR #1229, opened 2024-10-23
Add mujoco rgbd rendering

### LINKED ISSUE #1226: [Proposal] Add RGB-D Render Mode to MuJoCo Environments
### Proposal

My proposal is to add a new `render_mode` to MuJoCo environments for when RGB **and** Depth images are required as observations, e.g. to create point clouds.

(related issue: #727)

### Motivation

Currently one can achieve this by calling `MujocoEnv.mujoco_renderer.render` twice with both `render_mode=rgb_array` and `render_mode=depth_array` respectively. However, internally rgb and depth are always rendered and only one is returned based on the `render_mode`. Therefore, offering to return both rgb and depth in one step saves the effort of rendering both images twice.

### Pitch

'rgbd_array' should be added as valid render_mode that returns both the rgb and the depth image either separately or as a 4 channel image (which entails conversion from uint8 to float32 and back for rgb).

### Alternatives

_No response_

### Additional context

I've already implemented a solution in my [fork](https://github.com/DavidPL1/Gymnasium/tree/add-mujoco-rgbd-rendering) that returns both images separately.

Here's a comparison of execution times produced by the following script:
```python
import timeit
import numpy as np

setup_render_twice = """
import gymnasium as gym
env = gym.make('Pusher-v5', render_mode='rgb_array')
env.reset()
"""

render_twice = """
rgb = env.get_wrapper_attr('mujoco_renderer').render('rgb_array')
depth = env.get_wrapper_attr('mujoco_renderer').render('depth_array')
"""

setup_render_once = """
import gymnasium as gym
env = gym.make('Pusher-v5', render_mode='rgbd_array')
env.reset()
"""

render_once = """
rgb, depth = env.render()
"""



times_twice = timeit.repeat(setup=setup_render_twice, stmt=render_twice, number=10, repeat=10000)
print(f"Render twice:\n\tmean: {np.mean(times_twice):.4f}+-{np.std(times_twice):.4f}\n\tmin: {np.min(times_twice):.4f}\n\tmax: {np.max(times_twice):.4f}\n")

times_once = timeit.repeat(setup=setup_render_once, stmt=render_once, number=10, repeat=10000)
print(f"RGB-D:\n\tmean: {np.mean(times_once):.4f}+-{np.std(times_once):.4f}\n\tmin: {np.min(times_once):.4f}\n\tmax: {np.max(times_once):.4f}\n")
```

Results:
```
Render twice:
        mean: 0.0301+-0.0018
        min: 0.0283
        max: 0.0633

RGB-D:
        mean: 0.0136+-0.0010
        min: 0.0127
        max: 0.0246
```

### Checklist

DavidPL1:
# Description

Adds a new `render_mode` to MuJoCo environments for when rgb and depth images are required as observations, e.g. to create point clouds.

Fixes #1226 

## Type of change

# Checklist:

# Open Questions
- Currently this change was applied to all v5 and v4 environments. Is that correct or should I create v6 envs?
- In case just changing v5 and v4 is fine, should this feature be added to the version history docstrings of the respective envs?


gymnasium/utils/passive_env_checker.py
```diff
@@ -292,6 +292,34 @@ def _check_render_return(render_mode, render_return):
             logger.warn(
                 f"Depth-array rendering should return a numpy array with two axes, got {render_return.ndim}"
             )
+    elif render_mode == "rgbd_array":
```
pseudo-rnd-thoughts, line 295: I don't think we need this code as I don't mind many other projects will use it

pseudo-rnd-thoughts (changes requested):
Looks good to me other than the comment. 
@Kallinteris-Andreas could you review as well

gymnasium/utils/passive_env_checker.py
DavidPL1: good point. I'll remove it

DavidPL1:
@pseudo-rnd-thoughts 
Looks like `Run PyTest / build-all (3.8, >=1.21,<2.0) ` for some reason got stuck at `apt-get -y update` can you trigger it again?

pseudo-rnd-thoughts:
I can't from my phone but will be home in an hour or so and can do then

gymnasium/envs/mujoco/mujoco_rendering.py
```diff
@@ -281,6 +281,10 @@ def render(
                         seg_ids[geom.segid + 1, 1] = geom.objid
                 rgb_img = seg_ids[seg_img]
 
+            if render_mode == "rgbd_array":
```
Kallinteris-Andreas, line 284: This section (L259-L289) was not written to support 3 render types, it should be re-written to be clearer, into 2 stages, first stage collects the images, and the second stage returns the correct images.
```py
if render_mode in ["depth_array", "rgbd_array"]:
    depth_img = depth_arr.reshape(self.viewport.height, self.viewport.width)
if render_mode in ["rgb_array", "rgbd_array"]:
    ...
```

```py
if render_mode == "rgb_array":
    return ...
if render_mode == "depth_array":
    return ...
if render_mode == "rgbd_array":
    return ...
```

Kallinteris-Andreas (changes requested):
1) need to update the index page  (https://gymnasium.farama.org/environments/mujoco/ ) to document the available `render_mode`s

2) add a test in `test_mujoco_rendering.py` named `test_rgbd_array`, where we assert the output of `test_rgbd_array` is a proper combination of `rgb_array` and `depth_array`

3) I am not convinced that the correct output format is a tuple of `(rgb_image, depth_image)` while it is the most human-readable format, it is not ideal for CNNs and would require an additional `renderWrapper` (for the cases where it used as `addRenderObservation` wrapper)
edit: RGB-D CNNs do not appear to be "mature", see "RGB-D Object Recognition Using Deep Convolutional Neural Networks" as an example the data is preprocessed anyway, so long as the decision is done intentionally, I am fine with the current choice. @DavidPL1 you may be more familiar with depth CNNs what is your oppinon

Thanks

DavidPL1:
> 1. need to update the index page  (https://gymnasium.farama.org/environments/mujoco/ ) to document the available `render_mode`s
> 
> 2. add a test in `test_mujoco_rendering.py` named `test_rgbd_array`, where we assert the output of `test_rgbd_array` is a proper combination of `rgb_array` and `depth_array`
> 
Will do after we settle on the type 

> 3. I am not convinced that the correct output format is a tuple of `(rgb_image, depth_image)` while it is the most human-readable format, it is not ideal for CNNs and would require an additional `renderWrapper` (for the cases where it used as `addRenderObservation` wrapper)
> edit: RGB-D CNNs do not appear to be "mature", see "RGB-D Object Recognition Using Deep Convolutional Neural Networks" as an example the data is preprocessed anyway, so long as the decision is done intentionally, I am fine with the current choice. @DavidPL1 you may be more familiar with depth CNNs what is your oppinon

I'm am also not too familiar with RGB-D CNNs myself. After a quick read on the topic, researchers seem to perform all kinds of processing steps for alternative representations (voxels, pointclouds, ...). Usually RGB-D cameras provide access to two separate streams and datasets also provide the modalities separated.
Still, from my point of view both ways, tuple or 4D-image, would be fine. And custom wrappers should be used to get a favored representation.

I opted for a tuple because I thought it achieves the same thing as the rendering twice alternative (which I have already seen), returning the same types as the individual images. Also I'm using an existing Wrapper combining the two images into a point cloud.
Though I must admit that the naming `rgbd_array` might mislead into thinking a single array is returned.

Kallinteris-Andreas:
The `Open3D` python library seems to store the rgbd data on a tuple https://www.open3d.org/docs/latest/tutorial/Basic/rgbd_image.html#RGBD-images
I can not figure out how `OpenCV` structures the storage of rgbd images

A tuple of rgb and depth should be fine

DavidPL1:
judging by the function calls of e.g. [rgbd::warpFrame](https://docs.opencv.org/3.4/d2/d3a/group__rgbd.html#gac0db6aeba01fa17ec2c69694497926f0) taking both as separate inputs, OpenCV seems to also store image and depth separately.

Then I'll stick with the tuple implementation. `rbgd_tuple` also sounds good to me.

Kallinteris-Andreas:
@pseudo-rnd-thoughts All I want from you is to tell us if you are okay with the name of the new `render_mode` being "`rgbd_tuple`"

DavidPL1:
> 2. add a test in `test_mujoco_rendering.py` named `test_rgbd_tuple`, where we assert the output of `test_rgbd_tuple` is a proper combination of `rgb_array` and `depth_array`

See https://github.com/Farama-Foundation/Gymnasium/pull/1229/commits/a72f44d5cffe9ad51ebdcd7652a5880fc18c6a05 or do you want to add something more to the test?

tests/envs/mujoco/test_mujoco_rendering.py
```diff
@@ -94,6 +99,27 @@ def test_camera_id(render_mode: str):
...
+        assert (depth_a != depth_c).any()
+
+    elif render_mode != "human":
         assert (env_a.render() == env_b.render()).all()
         assert (env_a.render() != env_c.render()).any()
+
+
+def test_rgbd_tuple():
+    """Assert that rgbd_tuple is the proper combination of rgb and depth images as tuple"""
+    env = gymnasium.make("Ant-v5", camera_id=0, render_mode="rgbd_tuple").unwrapped
```
Kallinteris-Andreas, line 118: render once with `rgb_array` and `depth_array` and assert that the resulting images are the same

pseudo-rnd-thoughts (approved):
`rgbd_tuple` sounds good to me.

Kallinteris-Andreas (approved):

PR #1231, opened 2024-10-24
Add uv to install pip deps in workflows

ffelten:
Building the docker container now takes ~1min40, was 2min40 before.


pseudo-rnd-thoughts (approved):

AdrienBolling:
uv is the best

PR #1239, opened 2024-11-04
Fix Train Agent doc, need env param

satur9nine:
# Description

The first parameter of the BlackjackAgent example needs to be the environment.

This will save time for those copying this code as they learn while reading along.

## Type of change

- [X] Documentation only change (no code changed)


pseudo-rnd-thoughts (approved):
Thanks, fixed it

PR #1240, opened 2024-11-07
StickyAction wrapper can repeat the old action for more than 1 step

sparisi:
The StickyAction wrapper now takes an optional argument to allow the old action to be repeated for more than 1 steps (default is 1). The original behavior is unchanged. 

# Description

No fix, this is an extra feature. It increases the difficulty of sticky actions and the "non-Markovianity" of the environment (the more steps the action is repeated, the more in the past the agent should look to predict the next state). It can be useful for RL of non-Markov decision processes.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/stateful_action.py
+++ b/gymnasium/wrappers/stateful_action.py
@@ -18,7 +18,8 @@ class StickyAction(
     """Adds a probability that the action is repeated for the same ``step`` function.
 
     This wrapper follows the implementation proposed by `Machado et al., 2018 <https://arxiv.org/pdf/1709.06009.pdf>`_
-    in Section 5.2 on page 12.
+    in Section 5.2 on page 12, and adds the possibility to repeat the action for
+    more than one step.
 
     No vector version of the wrapper exists.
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
@sparisi If I understand correctly, this will limit the number of repeat actions that can be taken.

pseudo-rnd-thoughts:
Ahh I understand better now thanks for the description
Originally, the implement was just a single sticky action step where, given a probability, there was a chance of taking the last action. 
In Atari, this is paired with FrameSkip, i.e., the agent takes X steps with a 0.25% of taking the last action taken.  

ALE implements two options for frameskip of a deterministic, v5 with 4 frames and v0 randomly between 2 and 5. 

This PR is added frameskip for a random value up to X which is equivalent to the v0 style randomly between 1 and X. 
Am I understanding correctly? 
If I am, I would shift it to the ALE style implement with either a deterministic number of frames, `int`, or a stochastic number of frames with `tuple[int, int]`

sparisi:
@pseudo-rnd-thoughts 
I see. 
I think the difference with FrameSkip is that the agent, when actions are repeated, does still try to execute an action. With FrameSkip, instead, the agent is not aware of the repeats. 
The main reason I made this wrapper is to study how non-Markovianity affects classic RL. 

"This PR is added frameskip for a random value up to X" 
Actually, the number of steps the action is repeated is deterministic. What's random is when the action starts being repeated (the randomness in StickyAction). 

Now I have added the possibility to have stochastic repeats within a range. When the agent starts a series of repeats, the duration is randomly determined within the range passed as argument to the wrapper.

pseudo-rnd-thoughts:
@sparisi I came back the PR after a couple of days and hopefully better understand what you are talking about with the difference with this vs frameskip. 
I've updated the code to change the variable names and added a test that checks the expected actions match the executed actions. 
Could you check the code, then I would be happy to merge

sparisi:
@pseudo-rnd-thoughts 
Checked and tried on an environment, all good. Thanks!

pseudo-rnd-thoughts (approved):

PR #1242, opened 2024-11-08
Fix parameter name in tutorial code

bpesquet:
# Description

There is something off in the blackjack tutorial code. The current version uses a `deque_size` parameter which doesn't  exist. I renamed it `buffer_length`, hoping this is the right parameter to use.

In order to test my code, I had to add a missing dependency (seaborn) to `pyproject.toml`.

I also gitignored the `/.venv` subfolder, which I think is quite frequently used for storing virtual envs.

## Type of change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ __pycache__/
 # Virtualenv
 /env
 /venv
+/.venv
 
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts (approved):
Thanks for fixing that

PR #1253, opened 2024-11-15
Fix discrete CarRacing-v3

Sebastian-Griesbach:
# Description

CarRacing-v3 in discrete mode first converts the action to a float value before checking if the action is part of the discrete action-space. This causes an error to be raised.

Minimal example to reproduce this error:

```
import gymnasium as gym
env = gym.make("CarRacing-v3", continuous=False)
_ = env.reset()
action = env.action_space.sample()
_ = env.step(action)
```

This problem can be fixed simply by moving the float conversion to after the `if self.continuous:` check such that for the discrete case `else:` the action will still be an integer. And the `InvalidAction` error is not raised.

# Checklist:

- [NA] I have commented my code, particularly in hard-to-understand areas
- [NA] I have made corresponding changes to the documentation
- [NA] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gymnasium/envs/box2d/car_racing.py
+++ b/gymnasium/envs/box2d/car_racing.py
@@ -541,8 +541,8 @@ class CarRacing(gym.Env, EzPickle):
     def step(self, action: Union[np.ndarray, int]):
         assert self.car is not None
         if action is not None:
-            action = action.astype(np.float64)
             if self.continuous:
+                action = action.astype(np.float64)
                 self.car.steer(-action[0])
                 self.car.gas(action[1])
                 self.car.brake(action[2])
--- a/tests/envs/test_env_implementation.py
+++ b/tests/envs/test_env_implementation.py
@@ -346,27 +346,22 @@ def test_cartpole_vector_equiv():
     envs.close()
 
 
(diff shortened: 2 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for spotting this issue, could you add a test to check that this works, probably in `tests/envs/test_env_implementations.py`
I believe that Lunar Lander has a similar `continuous=False`, could you add this to the test. You just need to pass valid and invalid action and check that the error occurs

Sebastian-Griesbach:
I added a test that checks if continuous actions fail in the discrete setting and discrete actions don't raise any error.

pseudo-rnd-thoughts (approved):
Thanks for making that change @Sebastian-Griesbach

Kallinteris-Andreas:
For how long has this bug existed, do we need to list it in the environment version history

https://github.com/Farama-Foundation/Gymnasium/blob/e73245912087d47b538dcdb45fa9a9d185b805c5/gymnasium/envs/box2d/car_racing.py#L192-L196

pseudo-rnd-thoughts:
I dont think this changes environment behaviour

PR #1254, opened 2024-11-18
fix typo in docs/api/wrappers/table.md

Yuning30:
# Description

Fix an error in the docs: the wrapper checks if ```step``` and ```render``` are called before ```reset```, rather than previous ```render```.

## Type of change


pseudo-rnd-thoughts (approved):
Thanks for spotting that

PR #1277, opened 2024-12-16
Update mujoco_env.py xml file reading

KarahanS:
# Description

Change: Avoid calling `from_xml_path` and instead first read the file and then call `from_xml_string` with the content. 

Issue: Calling `from_xml_path` gives the following error (although the file exists):
`ValueError: ParseXML: Error opening file 'C:\Users\karab\Desktop\Archive\Tübingen Archive\ML-4350 Reinforcement Learning\code\env\lib\site-packages\gymnasium\envs\mujoco\assets\half_cheetah.xml': No such file or directory`

It may be because, MuJoCo's C++ backend might handle file paths differently than Python. (See https://github.com/google-deepmind/mujoco/issues/257#issuecomment-1111139463)

Python: 3.9.13
gymnasium: 1.0.0
mujoco: 3.2.6

## Type of change

### DIFF
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -126,7 +126,9 @@ class MujocoEnv(gym.Env):
         """
         Initialize MuJoCo simulation data structures `mjModel` and `mjData`.
         """
-        model = mujoco.MjModel.from_xml_path(self.fullpath)
+        with open(self.fullpath, 'r') as f:
+            xml_content = f.read()
+        model = mujoco.MjModel.from_xml_string(xml_content)
         # MjrContext will copy model.vis.global_.off* to con.off*
         model.vis.global_.offwidth = self.width
         model.vis.global_.offheight = self.height


Kallinteris-Andreas:
do you use Microsoft's WSL?

KarahanS:
I have WSL installed in my Windows machine, but I didn't run the code from the WSL.

Kallinteris-Andreas:
try with WSL

KarahanS:
I'm using mujoco 3.2.6. 
I think requiring users to switch to WSL unnecessarily complicates the usage - that's why I opened the PR.

Kallinteris-Andreas:
I just want to know if it's a Microsoft Windows issue Or the power shall issue.

pseudo-rnd-thoughts:
I recently got my windows PC working and tested this PR with the following script where `test_model.xml` is a copy of `ant.xml` in the same folder

```python
import gymnasium as gym
import os

env = gym.make("Ant-v5")
env.reset()
env.close()

env = gym.make("Ant-v5", xml_file=os.getcwd() + "\\test_model.xml")
env.reset()
env.close()
```

And I got no issue using `mujoco==3.2.5`
Therefore, I'm going to close the PR but we can re-open it if anyone can provide an example script to test with that raises the error

pseudo-rnd-thoughts:
ok, I randomly ran some other code and can replicate this issue but with `3.1.6`
@Kallinteris-Andreas should we update to support both versions? Currently Gymnasium-robotics uses `<3.2` which is why I refound the issue

Kallinteris-Andreas:
1) Robotics will support 3.2 next release
2) this fix causes issues with loading models on other repos (meta robotics and third party), because it doed not support multi file loading

PR #1281, opened 2024-12-23
Fixup for 'Training An Agent' page

chr0nikler:
# Description

"Training An Agent" code does not match visual outputs. Short term solution is to change visualization code to match outputs. Long-term proposal to change page to be specifically about agent training and let users go to tutorial for black jack, or do away with page entirely. If keeping page, could turn into a 2-arm bandit e-greedy bandit.

Fixes # (issue)

## Type of change

### Screenshots

N/A (viewing markdown changes should suffice but can attach)

# Checklist:

### DIFF
--- a/docs/introduction/train_agent.md
+++ b/docs/introduction/train_agent.md
@@ -155,37 +155,44 @@ You can use `matplotlib` to visualize the training reward and length.
 
 ```python
 from matplotlib import pyplot as plt
-# visualize the episode rewards, episode length and training error in one figure
-fig, axs = plt.subplots(1, 3, figsize=(20, 8))
 
-# np.convolve will compute the rolling mean for 100 episodes
+def get_moving_avgs(arr, window, convolution_mode):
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (commented):
Thanks for the PR, sorry for the late review.
Quick question, why is two of them moving average using "valid" and the final one using "same"?

chr0nikler:
Yeah that middle graph is greatly affected by the start and end windows if they aren't full. Centering (by "same" which is what the python code in the [BlackJack Tutorial](https://github.com/Farama-Foundation/Gymnasium/blob/main/docs/tutorials/training_agents/blackjack_tutorial.py) does) produces this

Figure_1

For reference: "valid" removes boundary effects. "same" produces the output length equal to max of the bigger of the two arrays.

pseudo-rnd-thoughts (approved):
Ok, I think that makes sense

PR #1285, opened 2025-01-02
Update documentation for reward scaling wrappers

### LINKED ISSUE #1272: [Proposal/Question] Incorrect documentation of NormalizeReward wrapper
### Proposal

The documentation for [NormalizeReward](https://gymnasium.farama.org/api/vector/wrappers/#gymnasium.wrappers.vector.NormalizeReward) ([here](https://github.com/Farama-Foundation/Gymnasium/blob/38dab5e05e71f0a087533a79e0aeac3b366bd672/gymnasium/wrappers/stateful_reward.py#L20) in code), is partially incorrect and unclear. It states:

> This wrapper will scale rewards s.t. the discounted returns have a mean of 0 and std of 1.

1. The wrapper does **not** normalize discounted returns to have a mean of 0. Instead, the rewards are merely divided by the running standard deviation of a specific term (see [here](https://github.com/Farama-Foundation/Gymnasium/blob/38dab5e05e71f0a087533a79e0aeac3b366bd672/gymnasium/wrappers/stateful_reward.py#L131)). The description should be changed accordingly.

2. I am also concerned about the discounted returns having a variance of 1. The term we are computing the running variance of is $\sum_{t=0}^T \gamma^{T-t} r_t$, where $T$ is an episode's current time step. I would interpret this as "backwards-discounting". Importantly, this is different from the discounted return ($\sum_{t=0}^\infty \gamma^t r_t$), or even what I would call the discounted sum of previous rewards up until timestep $T$ ($\sum_{t=0}^T \gamma^t r_t$). I assume the latter is what the description means by "discounted returns", which should be clarified if that is the case.
To me, it is also unclear how dividing by this term leads to a unit variance of either the discounted return or the discounted return up to $T$. In fact, I was not able to empirically confirm that it does. In contrast, I was able to confirm  $\text{Var}\left[\sum_{t=0}^T \gamma^{T-t} r_t\right] = 1$. 
The [reference](https://openreview.net/forum?id=r1etN1rtPB) in the wrapper's description does not mention any theoretical properties of this discounting theme. 

3. The description also says "The exponential moving average will have variance $(1 - \gamma)^2$". I do not understand what exactly this means (and shouldn't it depend on the reward?), and would be happy about any explanation or reference that explains this.

I would be happy about any clarifications concerning points 2 and 3. I also suggest adding the relevant references to the documentation.

### Motivation

I am using reward scaling myself, and find Gymnasium to be an important reference implementation. Clear documentation is important to understand the methods that Gymnasium implements.

### Pitch

1. Change the description of the NormalizeReward wrapper to remove incorrect information.
2. Clarify the description (or answer my questions :) ) by adding relevant references. 

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

keraJLi:
# Description

Changes the documentation of reward scaling wrappers. It mainly removes incorrect or unsubstantiated information.
Affected wrappers are [wrappers/stateful_reward.py](https://github.com/Farama-Foundation/Gymnasium/blob/c11ac0501594fb2b5bf3b932698f2b4085bf7414/gymnasium/wrappers/stateful_reward.py) and [wrappers/vector/stateful_reward.py](https://github.com/Farama-Foundation/Gymnasium/blob/c11ac0501594fb2b5bf3b932698f2b4085bf7414/gymnasium/wrappers/vector/stateful_reward.py).

Fixes #1272 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/stateful_reward.py
+++ b/gymnasium/wrappers/stateful_reward.py
@@ -20,21 +20,13 @@ __all__ = ["NormalizeReward"]
 class NormalizeReward(
     gym.Wrapper[ObsType, ActType, ObsType, ActType], gym.utils.RecordConstructorArgs
 ):
-    r"""This wrapper will scale rewards s.t. the discounted returns have a mean of 0 and std of 1.
-
-    In a nutshell, the rewards are divided through by the standard deviation of a rolling discounted sum of the reward.
-    The exponential moving average will have variance :math:`(1 - \gamma)^2`.
+    r"""Normalizes immediate rewards such that their exponential moving average has an approximately fixed variance.
 
     The property `_update_running_mean` allows to freeze/continue the running mean calculation of the reward
     statistics. If `True` (default), the `RunningMeanStd` will get updated every time `self.normalize()` is called.
     If False, the calculated statistics are used but not updated anymore; this may be used during evaluation.
 
     A vector version of the wrapper exists :class:`gymnasium.wrappers.vector.NormalizeReward`.
-
-    Important note:
-        Contrary to what the name suggests, this wrapper does not normalize the rewards to have a mean of 0 and a standard
-        deviation of 1. Instead, it scales the rewards such that **discounted returns** have approximately unit variance.
-        See [Engstrom et al.](https://openreview.net/forum?id=r1etN1rtPB) on "reward scaling" for more information.
 
     Note:
         In v0.27, NormalizeReward was updated as the forward discounted reward estimate was incorrectly computed in Gym v0.25+.
@@ -74,7 +66,6 @@ class NormalizeReward(
         ...     episode_rewards.append(reward)
         ...
         >>> env.close()
-        >>> # will approach 0.99 with more episodes
         >>> np.var(episode_rewards)
         np.float64(0.010162116476634746)
 
@@ -89,7 +80,7 @@ class NormalizeReward(
         gamma: float = 0.99,
         epsilon: float = 1e-8,
     ):
-        """This wrapper will normalize immediate rewards s.t. their exponential moving average has a fixed variance.
+        """This wrapper will normalize immediate rewards s.t. their exponential moving average has an approximately fixed variance.
 
         Args:
             env (env): The environment to apply the wrapper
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (commented):
Thanks for the PR @keraJLi, to clarify what do you mean by `their exponential moving average`?
To me, this isn't clear what the expected mean is or what exactly the rewards are normalised by?

keraJLi:
The rewards are scaled like this:
1. Starting every episode, we take an EMA of the rewards as $\mu_t = r_t + \gamma \mu_{t-1}$, starting with $\mu_0 = 0$. Note that here $\gamma$ does **not** correspond to the discount factor. I simply kept the notation of the original paper.
2. We track the running variance of $\mu$, across episodes. In the limit, it approaches $\mathbb{E}_t[\text{Var}(\mu_t)]$.
3. We divide rewards by this value.

This means
- The reward's expected mean differs between environments and policies since it changes based on the reward variance.
- The reward's variance differs between environments and policies. In rare cases, it might even be increased.
- The same holds for the mean and variance of the episodic return.

Sadly, you cannot even say the EMA has variance one. This is because rewards are divided by the running variance of the EMA **at different time steps**. 

To me, it seems like you can't really draw any general conclusions about the properties of this method without some assumptions (e.g. if your reward's autocorrelation decays exponentially and the episode length distribution is geometric, you get a nice upper bound on the reward variance).

pseudo-rnd-thoughts:
Thanks for the reply @keraJLi

I think your reword of the docstring makes sense now, I'm happy to merge if you want
Would we want to specify that `epsilon` is a `reward scaling` parameter not the `discount factor` as I think I originally believed? 

Looking at the paper again and, in particular, the code implementation, https://openreview.net/pdf?id=r1etN1rtPB#page=11.12
It appears like the paper and Gym/Gymnasium's implementation significantly differ, in particular, that we add the `-rs.mean` that makes the function look more like a traditional normalising function than the paper's version. 
An idea that the paper explicitly says it isn't doing. 

Looking at the baseline repo that I suspect might be the first implementation - https://github.com/openai/baselines/blob/master/baselines/common/vec_env/vec_normalize.py
They don't include the `-rs.mean` in the step, but in reset, it does confusingly.

pseudo-rnd-thoughts (approved):

PR #1288, opened 2025-01-08
Add `wrappers.vector.TransformObs/Action` single obs/action space argument

### LINKED ISSUE #1287: [Bug Report] `single_observation_space` not updated with vectorized `TransformObservation`
### Describe the bug

`gymnasium.wrappers.vector.TransformObservation` does not set the observation space correctly. It has an argument for `observation_space`, which is correct, but it keeps the old `single_observation_space`. `TransformAction` appears to have the same issue.

I believe this can be a simple fix by just adding a `single_observation_space` argument to the constructor.

### Code example

```python
import gymnasium
from gymnasium.wrappers.vector import TransformObservation

env = gymnasium.make_vec("CartPole-v1", num_envs=8, vectorization_mode='sync')

old_obs_space = env.observation_space

env = TransformObservation(
    env,
    func = lambda obs: obs[:, 0:2],
    observation_space = gymnasium.spaces.Box(low=old_obs_space.low[:,0:2], high=old_obs_space.high[:,0:2], shape=(8, 2))
)

print(env.single_observation_space)
print(env.observation_space)
```

Expected output:
```
Box([-4.8               -inf], [4.8               inf], (2,), float32)
Box([[-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]], [[4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]], (8, 2), float32)
```

Actual output:
```
Box([-4.8               -inf -0.41887903        -inf], [4.8               inf 0.41887903        inf], (4,), float32)
Box([[-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]
 [-4.8 -inf]], [[4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]
 [4.8 inf]], (8, 2), float32)
```

### System info

Reproduced on 1.0.0 and on the main branch as of today (commit `fc74bb8`)

### Additional context

_No response_

### Checklist

howardh:
# Description

Fixes #1287

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/vector/vectorize_action.py
+++ b/gymnasium/wrappers/vector/vectorize_action.py
@@ -85,7 +85,7 @@ class TransformAction(VectorActionWrapper):
             # TODO: We could compute single_action_space from the action_space if only the latter is provided and avoid the warning below.
         if self.action_space != batch_space(self.single_action_space, self.num_envs):
             warn(
-                "The action space and the batched single action space don't match as expected."
+                f"For {env}, the action space and the batched single action space don't match as expected, action_space={env.action_space}, batched single_action_space={batch_space(self.single_action_space, self.num_envs)}"
             )
 
         self.func = func
--- a/gymnasium/wrappers/vector/vectorize_action.py
+++ b/gymnasium/wrappers/vector/vectorize_action.py
@@ -73,20 +73,6 @@ class TransformAction(VectorActionWrapper):
         """
         super().__init__(env)
 
-        """
-        self._single_observation_space_error = None
-        self._single_observation_space = self.env.single_observation_space
-        if observation_space is None:
-            if single_observation_space is not None:
-                self.observation_space = batch_space(single_observation_space, self.num_envs)
-        else:
-            self.observation_space = observation_space
-            if single_observation_space is None:
-                # TODO: We could compute this from the observation_space.
-                self._single_observation_space_error = "`single_observation_space` not defined. A new observation space was provided to the TransformObservation wrapper, but not the single observation space."
-            else:
-                self._single_observation_space = single_observation_space
-        """
         self._single_action_space_error = None
         self._single_action_space = self.env.single_action_space
         if action_space is None:
@@ -108,7 +94,7 @@ class TransformAction(VectorActionWrapper):
 
     @property
     def single_action_space(self) -> Space:
-        """The single observation space of the environment."""
+        """The single action space of the environment."""
         if self._single_action_space_error is not None:
             raise AttributeError(self._single_action_space_error)
         return self._single_action_space
--- a/gymnasium/wrappers/vector/vectorize_action.py
+++ b/gymnasium/wrappers/vector/vectorize_action.py
@@ -92,6 +92,7 @@ class TransformAction(VectorActionWrapper):
         if action_space is None:
             if single_action_space is not None:
                 self.action_space = batch_space(single_action_space, self.num_envs)
+                self._single_action_space = single_action_space
         else:
             self.action_space = action_space
             if single_action_space is None:
--- a/gymnasium/wrappers/vector/vectorize_action.py
+++ b/gymnasium/wrappers/vector/vectorize_action.py
@@ -9,6 +9,7 @@ import numpy as np
 
 from gymnasium import Space
 from gymnasium.core import ActType, Env
+from gymnasium.logger import warn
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for the PR,
I think the `__init__` can be significantly simpler, 
```python
if action_space is None:
    if single_action_space is not None:
        self.single_action_space = single_action_space
        self.action_space = batch_space(single_action_space, self.num_envs)
else:
    self.action_space = action_space
    if single_action_space is not None:
        self.single_action_space = single_action_space
        if action_space != batch_space(single_action_space, self.num_envs):
            warn("The action space and the batched single action space don't match as expected.")

self.func = func
```
The vector wrapper will automatically do the variable forwarding so we don't need the new function either

Could you update and change for the transform obs wrapper

howardh:
The extra stuff I'm doing with `self._single_action_space_error` is to preserve the current behaviour. If someone has code that is using the wrapper and they only assign the `action_space` and never use the `single_action_space`, then this allows them to continue as they were with no breaking changes. What you propose does simplify the check, but it will give the user a warning regardless of whether they need the `single_action_space` or not, which I believe is not desired behaviour.

Also, I think you want is this:

```python
if action_space is None:
    if single_action_space is not None:
        self.single_action_space = single_action_space
        self.action_space = batch_space(single_action_space, self.num_envs)
else:
    self.action_space = action_space
    if single_action_space is not None:
        self.single_action_space = single_action_space
if self.action_space != batch_space(self.single_action_space, self.num_envs):
    warn("The action space and the batched single action space don't match as expected.")

self.func = func
```

howardh:
If you're okay with the behaviour change, then I'll make the changes and resubmit.

pseudo-rnd-thoughts:
> If you're okay with the behaviour change, then I'll make the changes and resubmit.

Yes, I think it should be fine

tests/wrappers/vector/test_transform_observation.py
```diff
@@ -0,0 +1,97 @@
...
+        == np.array([110, 95, 110], dtype=np.float32)
+    ).all()
+
+
+def test_warning_on_mismatched_single_observation_space(
+    n_envs: int = 5,
+):
+    vec_env = SyncVectorEnv([create_env for _ in range(n_envs)])
+    # We only specify observation_space without single_observation_space, so single_observation_space inherits its value from the wrapped env which would not match. This mismatch should give us a warning.
+    with pytest.warns(Warning):
```
pseudo-rnd-thoughts, line 89: Add `match=re.esacpe("")` to check that the correct warning is being caught

tests/wrappers/vector/test_transform_action.py
pseudo-rnd-thoughts, line 1: Add warning check

gymnasium/wrappers/vector/vectorize_action.py
```diff
@@ -61,18 +62,31 @@ def __init__(
...
+                self.single_action_space = single_action_space
+                self.action_space = batch_space(single_action_space, self.num_envs)
+        else:
             self.action_space = action_space
+            if single_action_space is not None:
+                self.single_action_space = single_action_space
+            # TODO: We could compute single_action_space from the action_space if only the latter is provided and avoid the warning below.
+        if self.action_space != batch_space(self.single_action_space, self.num_envs):
+            warn(
+                "The action space and the batched single action space don't match as expected."
```
pseudo-rnd-thoughts, line 88: Improve the warning handling for users, `For {env}, the action space and the batched single action space don't match as expected, action_space={env.action_space}, batched single_action_space={
atch_space(self.single_action_space, self.num_envs)}`

Add to the other wrapper as well

pseudo-rnd-thoughts (changes requested):
Thanks for the update, looks good.
If you could add these changes then good to merge

pseudo-rnd-thoughts (approved):
Thanks for the changes, look good

PR #1294, opened 2025-01-18
Improve 'set_wrapper_attr'.

duburcqa:
# Description

I propose to further improve 'set_wrapper_attr' (as a follow-up of PR https://github.com/Farama-Foundation/Gymnasium/pull/1293). 

I added the optional keyword argument 'force'. Its role is to make sure that the attribute exists before setting it, so that no attribute while be create on the top-level wrapper. While this could already be done using 'has_wrapper_attr', but specifying the new optional argument is not only more convenient but also faster. Moreover, adding this extra argument enables to refactor the implementation of 'set_wrapper_attr' as a recursive function calling itself on the lower-level environment (in case of wrappers). This is not only more than twice faster (because `isinstance` is very slow...), but also more robust, as __it allows anybody to derive its own wrapper from `gym.Env` and provides its own implementation of 'set_wrapper_attr'__ (this aspect is very important to me, because my library is doing exactly this). Finally, the implementation is easier to understand and closely align with `get_wrapper_attr`.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/core.py
+++ b/gymnasium/core.py
@@ -273,9 +273,12 @@ class Env(Generic[ObsType, ActType]):
         """Gets the attribute `name` from the environment."""
         return getattr(self, name)
 
-    def set_wrapper_attr(self, name: str, value: Any):
+    def set_wrapper_attr(self, name: str, value: Any, *, force: bool = True):
         """Sets the attribute `name` on the environment with `value`."""
-        setattr(self, name, value)
+        if force or hasattr(self, name):
+            setattr(self, name, value)
+        else:
+            raise AttributeError(f"{self} has no attribute {name!r}")
 
 
 WrapperObsType = TypeVar("WrapperObsType")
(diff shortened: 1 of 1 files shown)


duburcqa:
@pseudo-rnd-thoughts Any thought on this PR ?

pseudo-rnd-thoughts (changes requested):
Apologies for taking a while to look at it @duburcqa

I like the idea, though I dislike returning an error. 
Could we use a `return True` if the attribute has been updated, then we are checking at the "top" wrapper if the lower wrappers have returned True, else set the attribute there? 

This seems like a better way and possibly faster because there is no error checking

duburcqa:
What I like with the implementation I suggest, is especially the ability to raise an exception. I think it adding an attribute in the base environment is rarely the intended objective. Raising an exception is a nice way to catch silly mistakes for the practitioner. Personally I would even prefer to set `force` to `False` by default, but I kept it to `True` for backward compatibility.

pseudo-rnd-thoughts:
> Besides, error CATCHING is slow, but error checking is free of cost (since Python 3.11). Since adding a new attribute is very unlikely, this approach is overall cheaper than `if ... else ...` causes

Oh, I've learnt something new, but aren't we doing error-catching multiple times if none of the wrappers have the attribute. Also return True / False seems more elegant therefore, I still want to change it

pseudo-rnd-thoughts (approved):

duburcqa:
I don't like the idea of silently NOT adding the attribute if the user specifies `force=False` manually. It is just extremely misleading compared to raising an exception, because nobody would expect that the return value should be checked to raise an exception by itself would reading the documentation. 

At this point, having to do the check manually and raise an exception does not bring any improvement compared to calling `has_wrapper_attr` and raising an exception if necessary, which was already supported before this MR.

PR #1296, opened 2025-01-22
Add probability masking to `space.sample`

### LINKED ISSUE #1255: [Proposal] Probability Mask for sampling a space
### Proposal

Currently, in the gymnasium/spaces/discrete module, the **sample()** method of the Discrete class only allows you to choose randomly between valid integers. It would be nice to extend it so that it can also accept a probability mask, where it gives the probability that a specific number can be returned. It could be used as an alternative to masking, because if a move is invalid, we can assign it a probability of zero.

### Motivation

This would be helpful for me, since I'm implementing an Ant Colony Optimization algorithm, and each action needs to be assigned a probability of being chosen.

### Pitch

Add a probability_mask parameter to the sample function, so that the user can choose to either provide a regular mask, probability mask, or neither. A verification should be added to make sure that they only provided one of the mask options, not both. We also need to verify that the probabilities add up to 1.

### Alternatives

Rather than adding a new argument to the sample function, we could also create a new function called probability_sample, that takes the arguments n, mask. Both arguments could be required, since if you're not going to add a mask, you might as well use the original sample method.

### Additional context

If the maintainers of Gymnasium think that this is a good idea, I would love to implement it myself and make a pull request.

### Checklist

mariojerez:
# Description

Adds a probability mask feature (`probability`) to the `sample()` method of all spaces. This allows you to specify the probability of choosing each action. Similarly to the `mask` parameter, the `probability` parameter is a numpy array with the same shape as `n`, the number of elements in the space. Each value in the array describes the probability of the corresponding value being chosen, with 0 meaning it will not be chosen and 1 meaning that it will be chosen. All of the values in the array must sum to 1. `probability` is unsupported for the Box and MultiBinary space.

## Motivation

This is helpful in instances where values need to be chosen at random, such that some values are given higher priority (a higher likelihood) over others. For example, when implementing an Ant Colony Optimization algorithm, each action needs to be assigned a different probability of being chosen.

Fixes #1255

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -61,7 +61,7 @@ class MultiBinary(Space[NDArray[np.int8]]):
         return True
 
     def sample(
-        self, mask: MaskNDArray | None = None, probability: MaskNDArray | None = None
+        self, mask: MaskNDArray | None = None, probability: None = None
     ) -> NDArray[np.int8]:
         """Generates a single random sample from this space.
 
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -7,6 +7,7 @@ from typing import Any, Sequence
 import numpy as np
 from numpy.typing import NDArray
 
+import gymnasium as gym
 from gymnasium.spaces.space import MaskNDArray, Space
 
 
@@ -59,7 +60,9 @@ class MultiBinary(Space[NDArray[np.int8]]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.int8]:
+    def sample(
+        self, mask: MaskNDArray | None = None, probability: MaskNDArray | None = None
+    ) -> NDArray[np.int8]:
         """Generates a single random sample from this space.
 
         A sample is drawn by independent, fair coin tosses (one toss per binary variable of the space).
@@ -68,11 +71,15 @@ class MultiBinary(Space[NDArray[np.int8]]):
             mask: An optional np.ndarray to mask samples with expected shape of ``space.shape``.
                 For ``mask == 0`` then the samples will be ``0`` and ``mask == 1` then random samples will be generated.
                 The expected mask shape is the space shape and mask dtype is ``np.int8``.
+            probability: A probability mask for sampling values from the MultiBinary space, currently unsupported.
 
         Returns:
             Sampled values from space
         """
         if mask is not None:
+            assert (
+                probability is None
+            ), "Only one of `mask` or `probability` can be provided, and `probability` is currently unsupported"
             assert isinstance(
                 mask, np.ndarray
             ), f"The expected type of the mask is np.ndarray, actual type: {type(mask)}"
@@ -90,6 +97,10 @@ class MultiBinary(Space[NDArray[np.int8]]):
                 mask == 2,
                 self.np_random.integers(low=0, high=2, size=self.n, dtype=self.dtype),
                 mask.astype(self.dtype),
+            )
+        elif probability is not None:
+            raise gym.error.Error(
+                f"MultiBinary.sample cannot be provided a probability, actual value: {probability}"
             )
 
         return self.np_random.integers(low=0, high=2, size=self.n, dtype=self.dtype)
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -8,3 +8,4 @@ pygame
 sphinx_github_changelog
 ale_py
 tabulate
+mujoco-py<2.2,>=2.1
--- a/gymnasium/spaces/box.py
+++ b/gymnasium/spaces/box.py
@@ -366,7 +366,7 @@ class Box(Space[NDArray[Any]]):
             )
         elif probability is not None:
             raise gym.error.Error(
-                f"Box.sample cannot be provided a probability, actual value: {probability}"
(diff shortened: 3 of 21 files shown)


pseudo-rnd-thoughts (commented):
Great job on the PR, overall looks good.
I will look at it in more detail this evening but the biggest thing that I notice is checking if the probabilities don't sum of 1. I'm not sure if numpy will do that for us, or if we'll need to add that check.

mariojerez:
Yeah I don't think we need to require their probability to sum to 1 since we normalize the probability mask anyways so that it adds to 1. I could remove this requirement if you'd like. This would also allow the user to input a zeros array, and have it behave the same way that a ``mask`` of zeros does: return ``space.start``.

Good to know about fixing seeds to help with testing.

Those two failed tests didn't come up for me locally when I ran pytest. I think it might be because I used a different version of numpy, since the string representation of numpy attributes are different on my machine.

pseudo-rnd-thoughts:
Yeah, we do testing with NumPy 2.0 and <2.0

I suspect that you might need to ignore the error if `numpy.__version__ < "2.0"`

gymnasium/spaces/multi_binary.py
```diff
@@ -91,6 +98,10 @@ def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.int8]:
                 self.np_random.integers(low=0, high=2, size=self.n, dtype=self.dtype),
                 mask.astype(self.dtype),
             )
+        elif probability is not None:
+            raise gym.error.Error(
```
pseudo-rnd-thoughts, line 102: Why is this unsupported currently? I'm happy to add this if you wish.

pseudo-rnd-thoughts (changes requested):
Impressive PR @mariojerez 

I think this is a question of programming styles, imho, I prefer simplity over minimal lines of code. 
Therefore, personally I think that we should implement the feature like
```python
if mask is not None and probability is not None: 
    raise ValueError
elif mask is not None:
    # logical mask sampling logic
elif probability is not None:
    # probabilistic mask sampling logic
else:
    # uniform sampling logic
```

This should make it easier to debug and understand for new users what is happening. I understand that this will increase the number of lines of code with duplicate error checking. Thoughts? 

There seem to be two other technical questions to ask
1. Should we enforce the sum of probabilities == 1 or do we normalise before applying the probabilities? Personally, I'm in favour of the first as it makes the requirements for users easy to understand. (If we do this, then use `numpy.isclose` with a small error, as floating point summation). 
2. For composite spaces, i.e., Tuple / Dict, do we allow users to mix the logical and probabilistic masks? From implementation simplity, I would say no. Plus logical is a subset of probabilistic, therefore, users can convert all logical cases to probabilistic to solve this. 

Do you agree with my thoughts?

mariojerez:
Thanks @pseudo-rnd-thoughts !

I agree that that template you provided is simpler and easier to understand. In discrete.py for example, here's what I'm thinking it could be changed to:
```
if mask is not None and probability is not None: 
    raise ValueError
elif mask is not None:
    self._validate_mask(
                mask,
                (self.n,),
                np.int8,
                "mask",
            )
    valid_action_mask = self._get_valid_action_mask(mask, "mask")
    # continue sampling logic for mask
elif probability is not None:
    self._validate_mask(
                probability,
                (self.n,),
                np.float64,
                "probability",
            )
    valid_action_mask = self._get_valid_action_mask(probability, "probability")
    # continue sampling logic for probability
else:
    # uniform sampling logic
```


1. I agree that enforcing probabilities == 1 makes it straight forward to understand. Anyways, the user isn't missing out too much from us enforcing it, because they can easily normalize it themselves.

Referring to your comment about doing the isclose check, In `_get_valid_action_mask()` (in discrete.py), I have this check:
```
assert np.isclose(
                np.sum(mask), 1
            ), f"The sum of all values of `probability mask` should be 1, actual sum: {np.sum(mask)}"
```
I think you're telling me to use `np.fsum` instead of `np.sum`, which makes sense. The relative tolerance and absolute tolerance of np.isclose is `rtol=1e-05`, `atol=1e-08` respectively by default. I don't have a problem with these default values, but do you think that I should hard-code them?

2. I agree with you, I think we can restrict them to one type of mask for composite spaces, without serious disadvantages. One thing to consider though is that according to Numpy's documentation, when adding probabilities (through the `p` parameter) to the numpy.random.Generator.choice method, it samples less efficiently. So I could see it benefiting a user to be able to sample more efficiently by using `mask` instead of `probability` whenever they can.

gymnasium/spaces/multi_binary.py
mariojerez: It wasn't obvious to me how to do it at the time, and I decided to move on. It honestly would be a huge help if you did! I'm pretty overwhelmed with classes and other commitments I have this semester.

mariojerez:
Sounds good. I'll make those changes when I get the chance. It may be a couple weeks, I'm really swamped with other work right now. If that timescale is longer than ideal, I don't mind if you make those changes. Otherwise, I'm happy to do it.

pseudo-rnd-thoughts:
Cool, I had some time today and looked over the whole PR and make some changes 
https://github.com/pseudo-rnd-thoughts/Gymnasium/tree/probability_mask

mariojerez:
Awesome, thank you! Should we close this pull request so that you make one from your branch with your changes?

pseudo-rnd-thoughts:
Updated version of this PR https://github.com/Farama-Foundation/Gymnasium/pull/1310

PR #1299, opened 2025-01-28
Add full support for dlpack API

### LINKED ISSUE #1298: [Bug Report] jax_to_torch uses deprecated dlpack.to() and dlpack.from()
### Describe the bug

The current JaxToTorch wrapper converts jax arrays into a DLPack capsule and then constructs a torch.Tensor from it. This way of transferring tensors is deprecated. Tensors now allow direct conversion via the __dlpack__ protocol.

This is only an issue on newer versions of jax and torch. Updating the wrapper would probably require bumping the version numbers of jax and torch.

### Code example

```shell
import gymnasium as gym
import jax.numpy as jp
import torch


class JaxEnv(gym.Env):
    def __init__(self):
        self.action_space = gym.spaces.Box(low=0, high=1, shape=(1,))
        self.observation_space = gym.spaces.Box(low=0, high=1, shape=(1,))

        return jp.array(self.observation_space.sample()), {}

    def step(self, action):
        return jp.array(self.observation_space.sample()), 0, False, False, {}


env = gym.wrappers.JaxToTorch(JaxEnv())
env.step(torch.tensor([0]))  # DeprecationWarning: Calling from_dlpack with a DLPack tensor is deprecated. The argument to from_dlpack should be an array from another framework that implements the __dlpack__ protocol.
```

### System info

- Install via pip
- Gymnasium 1.0.0
- Jax 0.4.38
- torch 2.5.1
- Ubuntu 22.04
- Python 3.11

### Additional context

_No response_

### Checklist

amacati:
# Description

This PR adds support for the modern dlpack API to transfer jax arrays to torch tensors and vice-versa. We now determine at initialization if the installed versions of jax and pytorch meet the API requirements, and select the suitable conversion path based on that flag.

Fixes #1298 

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/jax_to_torch.py
+++ b/gymnasium/wrappers/jax_to_torch.py
@@ -25,18 +25,6 @@ try:
     import jax.numpy as jnp
     from jax import dlpack as jax_dlpack
 
-    def _at_least_version(version: str, target: str) -> bool:
-        """Check if a version string has at least a certain version."""
-        version = version.split("+")[0]  # Remove any local version identifiers
-        version_parts = list(map(int, version.split(".")))
-        target_parts = list(map(int, target.split(".")))
-        # Pad shorter version to match lengths for comparison (e.g., '0.4.8' vs '0.4')
-        version_parts.extend([0] * (len(target_parts) - len(version_parts)))
-        target_parts.extend([0] * (len(version_parts) - len(target_parts)))
-        return tuple(version_parts) >= tuple(target_parts)
-
-    # TODO: Remove when we drop support for older jax and torch versions
-    _FULL_DLPACK_SUPPORT = _at_least_version(jax.__version__, "0.4.16")
 except ImportError:
     raise DependencyNotInstalled(
         'Jax is not installed therefore cannot call `torch_to_jax`, run `pip install "gymnasium[jax]"`'
@@ -46,9 +34,6 @@ try:
     import torch
     from torch.utils import dlpack as torch_dlpack
 
-    _FULL_DLPACK_SUPPORT = _FULL_DLPACK_SUPPORT and _at_least_version(
-        torch.__version__, "1.10.0"
-    )
     Device = Union[str, torch.device]
 except ImportError:
     raise DependencyNotInstalled(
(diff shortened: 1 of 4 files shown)


amacati:
The most problematic issue with the PR is the version check. Versions can be quite complex (see https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers), but I wanted to avoid a dependency on external packages such as [packaging](https://pypi.org/project/packaging/). If you are fine with the additional dependency, this would be the "correct" way of doing it. Otherwise, this implementation may not capture all possible version types. Doing so without additional dependencies would go beyond the scope of this PR.

gymnasium/wrappers/jax_to_torch.py
```diff
@@ -24,6 +24,19 @@
     import jax
     import jax.numpy as jnp
     from jax import dlpack as jax_dlpack
+
+    def _at_least_version(version: str, target: str) -> bool:
```
pseudo-rnd-thoughts, line 28: could we use https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

pseudo-rnd-thoughts (changes requested):
Looks good, my only request is to use the packaging versions as this will handle all the version checking complexity with alphas, betas, post, etc versions

gymnasium/wrappers/jax_to_torch.py
amacati: I think we should first discuss the implications of #1298 before adding another dependency. Since parts of the library are broken with older pytorch versions anyways, simply raising the version requirements and getting rid of the flag seems viable again.

gymnasium/wrappers/jax_to_torch.py
```diff
@@ -60,9 +61,7 @@ def _number_torch_to_jax(value: numbers.Number) -> Any:
 @torch_to_jax.register(torch.Tensor)
 def _tensor_torch_to_jax(value: torch.Tensor) -> jax.Array:
     """Converts a PyTorch Tensor into a Jax Array."""
-    tensor = torch_dlpack.to_dlpack(value)  # pyright: ignore[reportPrivateImportUsage]
-    tensor = jax_dlpack.from_dlpack(tensor)  # pyright: ignore[reportPrivateImportUsage]
-    return tensor
+    return jax_dlpack.from_dlpack(value)  # pyright: ignore[reportPrivateImportUsage]
```
amacati, line 64: One more question: Doing this with a tensor that requires gradients throws an error. I think this is reasonable. Tensors that require gradients should not be transferred to another framework, as we have no guarantees on what will happen with the data. Otherwise, we'd have to use `value.detach()` here. Depending on what you prefer, we should either leave it as is, or add the `detach` + tests.

pseudo-rnd-thoughts: I think we should leave it as is, and if someone has an issue then we can add it

pseudo-rnd-thoughts (approved):
Thanks for looking at this

pseudo-rnd-thoughts:
Ok, so the new jax requirements doesn't support python 3.8. 
@amacati could you just remove 3.8 from the testing, `.github/workflows/run_pytest.yaml`

amacati:
@pseudo-rnd-thoughts Great, done. I also removed the special requirement for numpy >= 2.0, since it was only required by python 3.8.

amacati:
@pseudo-rnd-thoughts Great, done. I also removed the special requirement for numpy >= 2.0, since it was only required by python 3.8.

PR #1306, opened 2025-02-12
Add None to None conversion for JaxToNumpy, JaxToTorch and NumpyToTorch

amacati:
# Description

The `JaxToNumpy`, `JaxToTorch` and `NumpyToTorch` wrappers currently have no conversion registered for `None` values. This becomes an issue e.g. when calling `env.render()` in "human" render modes, which implicitly returns `None`. Consider the following code:

```python
from gymnasium.wrappers.jax_to_numpy import JaxToNumpy


jax_env = GenericTestEnv()
numpy_env = JaxToNumpy(jax_env)
numpy_env.render()
# Exception: No known conversion for Jax type (<class 'NoneType'>) to NumPy registered. Report as
# issue on github.
```
The snippet fails because there is no None type conversion between NumPy, torch and jax. This PR fixes the behavior for all conversion wrappers and adds appropriate tests. The direct propagation of `None` is reasonable because its meaning does not change across the frameworks.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/jax_to_numpy.py
+++ b/gymnasium/wrappers/jax_to_numpy.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import functools
 import numbers
 from collections import abc
+from types import NoneType
 from typing import Any, Iterable, Mapping, SupportsFloat
 
 import numpy as np
@@ -68,6 +69,12 @@ def _iterable_numpy_to_jax(
         return type(value)(numpy_to_jax(v) for v in value)
 
 
+@numpy_to_jax.register(NoneType)
+def _none_numpy_to_jax(value: None) -> None:
+    """Passes through None values."""
+    return value
+
+
 @functools.singledispatch
 def jax_to_numpy(value: Any) -> Any:
     """Converts a value to a numpy array."""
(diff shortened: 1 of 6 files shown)


amacati:
Btw, this might be out of scope, but there are ongoing efforts to unify code dealing with arrays using the [array API](https://data-apis.org/array-api/latest/#). Might be worth to do away with specific NumpyToJax wrappers etc and have a generic "from array API compatible framework to another array API framework" wrapper. To maintain compatibility, one could alias the current wrappers with the generic one. 

Also, while we are at it, it would be good to be able to specify the device both for the environment and the conversion. At the moment, an observation transferred from jax to torch will be cast to the appropriate torch device, but the action (a torch Tensor) will always be cast to a jax Array on the GPU, when users might want to remain on the CPU. This could be fixed by specifying `from_device` and `to_device`.

@pseudo-rnd-thoughts what do you think?

pseudo-rnd-thoughts:
@amacati Thanks for bringing those to my attention. For the Array API, is it an easy wrapper to add? 
For the Device, you mean that in some cases, users might wish to push a GPU array to CPU but to not push the observations from GPU to CPU. If so, then yes, that shows like a helpful features. However, I would want the change to be back compatible

amacati:
Should be pretty easy, yes. I'll put together a PR when I find the time to work on it

PR #1310, opened 2025-02-19
Add probability masking to `space.sample`

### LINKED ISSUE #1255: [Proposal] Probability Mask for sampling a space

pseudo-rnd-thoughts:
# Description

Adds a probability mask feature (`probability`) to the `sample()` method of all spaces. This allows you to specify the probability of choosing each action. Similarly to the `mask` parameter, the `probability` parameter is a numpy array with the same shape as `n`, the number of elements in the space. Each value in the array describes the probability of the corresponding value being chosen, with 0 meaning it will not be chosen and 1 meaning that it will be chosen. All of the values in the array must sum to 1. `probability` is unsupported for the Box space.

## Motivation

This is helpful in instances where values need to be chosen at random, such that some values are given higher priority (a higher likelihood) over others. For example, when implementing an Ant Colony Optimization algorithm, each action needs to be assigned a different probability of being chosen.

Fixes #1255 and the completion of #1296

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -61,7 +61,7 @@ class MultiBinary(Space[NDArray[np.int8]]):
         return True
 
     def sample(
-        self, mask: MaskNDArray | None = None, probability: MaskNDArray | None = None
+        self, mask: MaskNDArray | None = None, probability: None = None
     ) -> NDArray[np.int8]:
         """Generates a single random sample from this space.
 
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -7,6 +7,7 @@ from typing import Any, Sequence
 import numpy as np
 from numpy.typing import NDArray
 
+import gymnasium as gym
 from gymnasium.spaces.space import MaskNDArray, Space
 
 
@@ -59,7 +60,9 @@ class MultiBinary(Space[NDArray[np.int8]]):
         """Checks whether this space can be flattened to a :class:`spaces.Box`."""
         return True
 
-    def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.int8]:
+    def sample(
+        self, mask: MaskNDArray | None = None, probability: MaskNDArray | None = None
+    ) -> NDArray[np.int8]:
         """Generates a single random sample from this space.
 
         A sample is drawn by independent, fair coin tosses (one toss per binary variable of the space).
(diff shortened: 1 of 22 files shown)


pseudo-rnd-thoughts:
@mariojerez Could you quickly review this PR to check that it makes sense

gymnasium/spaces/sequence.py
```diff
@@ -114,59 +120,78 @@ def sample(
...
+                Depending on if mask or probability is used will affect what argument is used.
+            probability: See mask description above, the only difference is on the ``sample_mask`` for the feature space being probability rather than mask.
 
         Returns:
             A tuple of random length with random samples of elements from the :attr:`feature_space`.
         """
+        if mask is not None and probability is not None:
+            raise ValueError("Only one of `mask` or `probability` can be provided.")
+
         if mask is not None:
```
mariojerez, line 137: Not a big deal, but to be consistent with the rest of the file changes, should make this an elif

gymnasium/spaces/multi_binary.py
```diff
@@ -91,8 +100,25 @@ def sample(self, mask: MaskNDArray | None = None) -> NDArray[np.int8]:
...
+            assert np.all(
+                np.logical_and(probability >= 0, probability <= 1)
+            ), f"All values of the sample probability should be between 0 and 1, actual values: {probability}"
 
-        return self.np_random.integers(low=0, high=2, size=self.n, dtype=self.dtype)
+            return (self.np_random.random(size=self.shape) <= probability).astype(
+                self.dtype
+            )
+        else:
+            return self.np_random.integers(low=0, high=2, size=self.n, dtype=self.dtype)
```
mariojerez, line 122: Missing tests for probability sampling MultiBinary

mariojerez (changes requested):
It's looking good! I caught some things that I think should be fixed. I added comments to specific lines in the Files changed tab. I noticed we're missing tests for MultiBinary probability sampling. I also suggested some rephrasing of a doc string to improve clarity.

pseudo-rnd-thoughts:
Thanks for the review @mariojerez, really helpful and if you are interested in helping out with any other PRs or farama projects, you would be very welcome

mariojerez:
Thank you! I'd be interested in helping out over the summer. I'll reach out.

PR #1315, opened 2025-02-21
Add stochastic taxi (rainy+fickle)

### LINKED ISSUE #161: [Proposal] Add transitional probabilities to Taxi and Cliff Walking toy text environments

foreverska:
# Description

Adds rainy transition probabilities and fickle passenger to align environment with paper.

Fixes #161 

## Type of change

# Checklist:


pseudo-rnd-thoughts (changes requested):
Hey @foreverska, thanks for the PR.

Generally the PR looks good. 
Could you change the `np.random` to `self.np_random` and revert the environment version increment (new features that don't affect default behaviour shouldn't require version bumps)

and a couple of questions before approving and merging
1. Is this backward compatible with default parameters? If it does currently, could we make it backward compatible.
2. Is the probability of the behaviour reported in `step` and `reset` correctly?

foreverska:
@pseudo-rnd-thoughts 

Addressed Comments.

> Is this backward compatible with default parameters

Yes, default values for rainy and fickle are both False.  This aligns functionality with pre-commit default behavior.

>Is the probability of the behaviour reported in step and reset correctly?

It matches the other ToyText environments (Cliff/Frozen) with step returning the probability of the taken transition and reset always returning 1.  I added lines in the unit test to guard against regression.

gymnasium/envs/toy_text/taxi.py
```diff
@@ -143,12 +139,19 @@ class TaxiEnv(Env):
...
+    <a id="fickle_passenger"></a>`fickle_passenger=True`: If true the passenger has a 30% chance of changing
+    destinations when the cab has moved one square away from the passenger's source location.
+
     ## References
     <a id="taxi_ref"></a>[1] T. G. Dietterich, “Hierarchical Reinforcement Learning with the MAXQ Value Function Decomposition,”
     Journal of Artificial Intelligence Research, vol. 13, pp. 227–303, Nov. 2000, doi: 10.1613/jair.639.
 
     ## Version History
-    * v3: Map Correction + Cleaner Domain Description, v0.25.0 action masking added to the reset and step information
+    * v3: Map Correction + Cleaner Domain Description, v0.25.0 action masking added to the reset and step information, added stochasticity
```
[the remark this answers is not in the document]
foreverska: Took a stab.  Wasn't totally sure what this would be included in but since a2 has been tagged I bumped the version to a3.

gymnasium/envs/toy_text/taxi.py
```diff
@@ -143,12 +139,20 @@ class TaxiEnv(Env):
...
+    <a id="fickle_passenger"></a>`fickle_passenger=True`: If true the passenger has a 30% chance of changing
+    destinations when the cab has moved one square away from the passenger's source location.
+
     ## References
     <a id="taxi_ref"></a>[1] T. G. Dietterich, “Hierarchical Reinforcement Learning with the MAXQ Value Function Decomposition,”
     Journal of Artificial Intelligence Research, vol. 13, pp. 227–303, Nov. 2000, doi: 10.1613/jair.639.
 
     ## Version History
     * v3: Map Correction + Cleaner Domain Description, v0.25.0 action masking added to the reset and step information
+        - In Gymnasium `1.0.0a3` the `is_rainy` and `fickle_passenger` arguments were added to align with Dietterich paper
```
pseudo-rnd-thoughts, line 155: this should be v1.1.0

gymnasium/envs/toy_text/taxi.py
```diff
@@ -289,9 +355,28 @@ def step(self, a):
...
+        taxi_row, taxi_col, pass_loc, _ = self.decode(s)
+
+        # If we are in the fickle step, the passenger has been in the vehicle for at least a step and this step the
+        # position changed
+        if (
+            self.fickle_step
+            and shadow_pass_loc == 4
+            and (taxi_row != shadow_row or taxi_col != shadow_col)
+        ):
+            self.fickle_step = False
```
pseudo-rnd-thoughts, line 370: So can fickle step only occur once in an episode?

gymnasium/envs/toy_text/taxi.py
```diff
@@ -306,6 +391,7 @@ def reset(
         super().reset(seed=seed)
         self.s = categorical_sample(self.initial_state_distrib, self.np_random)
         self.lastaction = None
+        self.fickle_step = True
```
pseudo-rnd-thoughts, line 394: Shouldn't this be conditional on the `if self.fickle_passenger:`

gymnasium/envs/toy_text/taxi.py
pseudo-rnd-thoughts: Could we make this conditional on `self.fickle_passenger` to make it clear that this can only occur if the user enables it

tests/envs/test_env_implementation.py
```diff
@@ -209,6 +209,60 @@ def test_taxi_encode_decode():
...
+            assert len(transitions) == 1
+            assert transitions[0][0] == 1.0
+
+    state, _ = env.reset()
+    _, _, _, _, info = env.step(0)
+    assert info["prob"] == 1.0
+
+
+def test_taxi_disallowed_transitions():
+    disallowed_transitions = [
```
pseudo-rnd-thoughts, line 239: how were these values generated?

tests/envs/test_env_implementation.py
```diff
@@ -209,6 +209,60 @@ def test_taxi_encode_decode():
         state, _, _, _, _ = env.step(env.action_space.sample())
```
pseudo-rnd-thoughts, line 210: could a test be added for fickle passenger?

pseudo-rnd-thoughts (changes requested):

gymnasium/envs/toy_text/taxi.py
foreverska: > So can fickle step only occur once in an episode

That was my reading of the following from the paper:
The second change is that after the taxi has picked up the passenger and moved one square away from the passenger's source location, the passenger changes his or her destination location with probability 0.3.

There's nothing in that language that suggests to me that it should happen multiple times.  Source location generally denotes the passenger's initial position.  An agent might drop off the passenger at the source location and pick them up again but I'm not sure fickle behavior in that case materially complicates the problem especially for a sufficiently advanced agent.

> Could we make this conditional on self.fickle_passenger to make it clear that this can only occur if the user enables it

Sure, working on a commit that might make this a lot easier to read.

tests/envs/test_env_implementation.py
foreverska: By hand, staring at the map.  A method could be written to build this list from the MAP if that was desired.

gymnasium/envs/toy_text/taxi.py
foreverska: addressed

tests/envs/test_env_implementation.py
foreverska: Yes, I figured out a way to force fickleness and added a test.  Not testing the rate of fickleness but the behavior should be sound.

gymnasium/envs/toy_text/taxi.py
foreverska: Made it conditional on both the fickle setting and whether or not the passenger is fickle that episode.

tests/envs/test_env_implementation.py
pseudo-rnd-thoughts: You could use `env.reset(seed=X)` if you know that fickle will be triggered from a rng

gymnasium/envs/toy_text/taxi.py
pseudo-rnd-thoughts: > The second change is that after the taxi has picked up the passenger and moved one square away from the passenger's source location, the passenger changes his or her destination location with probability 0.3.

Cool, that is what I expected, could we add this to the documentation for the argument

gymnasium/envs/toy_text/taxi.py
```diff
@@ -143,12 +139,20 @@ class TaxiEnv(Env):
...
+    <a id="fickle_passenger"></a>`fickle_passenger=True`: If true the passenger has a 30% chance of changing
+    destinations when the cab has moved one square away from the passenger's source location.
+
     ## References
     <a id="taxi_ref"></a>[1] T. G. Dietterich, “Hierarchical Reinforcement Learning with the MAXQ Value Function Decomposition,”
     Journal of Artificial Intelligence Research, vol. 13, pp. 227–303, Nov. 2000, doi: 10.1613/jair.639.
 
     ## Version History
     * v3: Map Correction + Cleaner Domain Description, v0.25.0 action masking added to the reset and step information
+        - In Gymnasium `1.1.0` the `is_rainy` and `fickle_passenger` arguments were added to align with Dietterich paper
```
pseudo-rnd-thoughts, line 155: Update the paper reference as `Dietterich, 2000` to be slightly more formal.

pseudo-rnd-thoughts (commented):
With the tests and documentation updates, then should be good to merge. 
Thanks for making the changes

tests/envs/test_env_implementation.py
foreverska: I had considered that.  My only qualm was if there is a change in the order stuff is drawn from the PRNG a seed may no longer be fickle.  That issue might be a bit nonobvious to future maintainers.

gymnasium/envs/toy_text/taxi.py
foreverska: Updated

tests/envs/test_env_implementation.py
```diff
@@ -209,6 +209,84 @@ def test_taxi_encode_decode():
...
+                        (end_row, end_col),
+                    ) not in disallowed_transitions
+
+
+def test_taxi_fickle_passenger():
+    env = TaxiEnv(fickle_passenger=True)
+    _, _ = env.reset()
+    # Force passenger being in a fickle state
+    env.fickle_step = True
+    state, reward, done, _, _ = env.step(0)
```
Kallinteris-Andreas, line 271: `terminated` not `done`

gymnasium/envs/toy_text/taxi.py
```diff
@@ -143,12 +139,22 @@ class TaxiEnv(Env):
     gym.make('Taxi-v3')
     ```
 
+    <a id="is_raining"></a>`is_raining=True`: If true the cab will move in intended direction with
```
Kallinteris-Andreas, line 142: it should be "True" not "true"

Kallinteris-Andreas (commented):
Looks good to me,

tests/envs/test_env_implementation.py
foreverska: Wasn't using the parameter anyway, changed to ignored.

gymnasium/envs/toy_text/taxi.py
foreverska: Updated

tests/envs/test_env_implementation.py
pseudo-rnd-thoughts: Fair but I think we can add a comment in the test explaining why the seed value was selected

gymnasium/envs/toy_text/taxi.py
```diff
@@ -143,12 +139,22 @@ class TaxiEnv(Env):
     gym.make('Taxi-v3')
     ```
 
+    <a id="is_raining"></a>`is_raining=True`: If True the cab will move in intended direction with
```
pseudo-rnd-thoughts, line 142: I just realised that this argument defaults to `false` not `true` if `is_raining=True` is meant to represent the default argument

Same for `fickle_passenger`

tests/envs/test_env_implementation.py
foreverska: Luckily the seed was easy to find, 42++.  Udpated.

gymnasium/envs/toy_text/taxi.py
foreverska: Ahhh, those are the default values not an example of how to enable it.  Got it.  Updated.

foreverska:
@pseudo-rnd-thoughts I think this needs one more review since there was a commit after your last one.  Thanks.

pseudo-rnd-thoughts:
Hey @foreverska, sorry I'm on holiday currently. Looking over the PR again, I'm a tad worried about the `is_rainy` change. 
Is there a way of making the code only run if `is_rainy` is true?

foreverska:
> Hey @foreverska, sorry I'm on holiday currently. Looking over the PR again, I'm a tad worried about the `is_rainy` change. Is there a way of making the code only run if `is_rainy` is true?

Please enjoy vacation and ignore this until you're good and rested.

Is the ask to restore the original code and switch to it when it's not rainy and switch to this new code if it is?  Or is there a more pointed change you'd like to see?

pseudo-rnd-thoughts:
hi @foreverska, I'm back. Looking over the PR with a week or two rest, I agree that I would prefer if the new code was "disabled" by default and wouldn't run at some, unlike the current solution. 
I know that this produces arguably less elegant code, I think it will be better maintenance and for people understanding the code.
Could you make that change?

foreverska:
@pseudo-rnd-thoughts Not a problem at all.  Pushed up a change that restores the old code when dry.  Let me know what you think.

gymnasium/envs/toy_text/taxi.py
```diff
@@ -220,11 +212,148 @@ def __init__(self, render_mode: Optional[str] = None):
                             self.P[state][action].append(
                                 (1.0, new_state, reward, terminated)
                             )
+
+    def __build_rainy_transitions(
```
pseudo-rnd-thoughts, line 216: To minimise changes, could this function, take `row, col, pass_idx, dest_idx, action` as arguments that we run the original code unless `is_rainy` then we call this function. 
Then return the data. 
It just means that we don't need to copy and paste the massive for loop and is clear what the differences between the current functions is

pseudo-rnd-thoughts (changes requested):

gymnasium/envs/toy_text/taxi.py
foreverska: I believe I have done a reasonable job at making the code as reusable as possible.  Please let me know if you had something else in mind.

pseudo-rnd-thoughts: Looks good @foreverska.

One last request is to change the function names to `_{name}` rather than double underscore to make the style of the rest of the project. 
Then we should be good to merge

foreverska: Roger, adjusted the function names, ready for final review.

PR #1323, opened 2025-02-28
Add termination condition based on percentage of visited tiles for Car Racing

### LINKED ISSUE #1269: [Bug Report] "CarRacing-v3" Appears to Reset before Completing a Lap causing it to Ignore lap_complete_percent
### Describe the bug

The termination and truncation values being returned from the environment don't seem to line up with what I'd expect from the documentation. The environment appears to ignore the value passed to `lap_complete_percent`, as the episode truncation flag is always `True` unless the agent touches every tile before crossing the 0th tile. It appears that this behavior is occurs because the environment **resets a few update cycles before** the agent re-crosses the 0th tile **if** the agent hasn't visited all the tiles in the track.

### Code example

```shell
import gymnasium as gym

from stable_baselines3 import PPO
from stable_baselines3.common.vec_env import SubprocVecEnv
from stable_baselines3.common.env_util import make_vec_env

from gymnasium.wrappers import GrayscaleObservation

import numpy as np


vec_env = make_vec_env(
    "CarRacing-v3",
    env_kwargs=dict(lap_complete_percent=0.5),
    n_envs=1,
    vec_env_cls=SubprocVecEnv,
    wrapper_class=GrayscaleObservation,
    wrapper_kwargs=dict(keep_dim=True),
)

best_model_string = "ppo_car_gray_2500000"
model = PPO.load(best_model_string, env=vec_env)
vec_env = model.get_env()

# * Running the Agent *
obs = vec_env.reset()

total_reward = 0
race_complete = False
for _ in range(1000):
    action, _states = model.predict(obs)
    obs, rewards, dones, info = vec_env.step(action)

    terminated = bool(dones[0])
    truncated = info[0]["TimeLimit.truncated"]

    total_reward += np.sum(rewards)

    if terminated or truncated:
        if truncated:
            print(f"Race Truncated     : {rewards} | Total: {total_reward}")
        elif terminated:
            print(f"Finished the Race! : {rewards} | Total: {total_reward}")
            race_complete = True
        #
    #
    vec_env.render()

    if race_complete:
        break
    #
#

print("Race Complete! Closing Environment")
vec_env.close()
#
```

### System info

Installation Method: Installed in a conda environment using pip
Gymnasium Version: 1.0.0
Python Version: 3.11.10

## Similar Issues and PRs:
* https://github.com/Farama-Foundation/Gymnasium/pull/813
* https://github.com/Farama-Foundation/Gymnasium/issues/106

## Detailed Description

I'm training an agent I created using the Stable Baselines 3 implementation of PPO (more details below). It is currently able to complete most of the track, but the termination and truncation values being returned don't seem to line up with what I'd expect from the documentation. The environment appears to ignore the value passed to `lap_complete_percent`, as the episode truncation flag is always `True` unless the agent touches every tile before crossing the 0th tile. It appears that this behavior is occurs because the environment resets a few update cycles before the agent re-crosses the 0th tile if the agent hasn't visited all the tiles in the track.

According to my current understanding of the code, `car_racing.py` will only check the percentage of lap completion when the vehicle crosses the starting tile, but after some testing I determined that the environment will reset when the car is approximately 2 few tiles before that point **AND** the car has covered enough tiles to satisfy `lap_complete_percent` **UNLESS** the car has covered all tiles during the first lap, in which case it will work as intended.

I have been able to consistently replicate this behavior by applying my agent and the interactive version of the environment. (The interactive version behaves a little differently. The environment will not reset until all tiles have been covered because it doesn't enforce a time limit. If I start the race as normally, then go off-road to skip at least one tile, then complete the rest of the lap on the track as normal, the race will not be considered complete when I re-cross the 0th tile. It allows me to continue around the track and will only terminate after I cross the tiles I skipped previously. From what I can tell, when I'm driving under manual control the task never terminates "successfully" as `self.env.new_lap` is never set to `True`.)

I have included a few print statements in my "fixed" version of the code to assist with troubleshooting this issue. I only needed to modify one specific section of code, which can quickly be found by searching for the unique string `~~~`.

### Agent Details
This is my first time using Gymnasium and Stable Baselines 3 so I've tried to keep things as simple as possible. I am applying the `GrayscaleObservation` wrapper to my environment to reduce the amount of time it takes the model to converge, but that's the only major modification I've made.

For my training environment I set `lap_complete_percent = 0.9`.

I used these two commands to load and train my agent:
```
model = PPO("CnnPolicy", vec_env, verbose=0, device="cuda")
model.learn(total_timesteps=2_500_000, progress_bar=True)
```

Saved Model: [ppo_car_gray_2500000.tar.gz](https://github.com/user-attachments/files/17998141/ppo_car_gray_2500000.tar.gz)
* I had to convert this file from a .zip to a .tar.gz so it was small enough to attach to this issue

### "Fixed" Code (In Collapsed Section)

I implemented a "quick fix" to verify that the problem is in most likely `car_racing.py` and not the code I've written. It slightly alters the behavior of the environment as the car no longer has to re-cross the 0th tile for the race to be considered complete. Instead it considers the race complete if the car is almost done with a lap (i.e. the index of the current tile is within 1% of the total number of tiles in the track).

<details>
<summary> Expand to View Code </summary>

```
__credits__ = ["Andrea PIERRÉ"]

import math
from typing import Optional, Union

import numpy as np

import gymnasium as gym
from gymnasium import spaces
from gymnasium.envs.box2d.car_dynamics import Car
from gymnasium.error import DependencyNotInstalled, InvalidAction
from gymnasium.utils import EzPickle


try:
    import Box2D
    from Box2D.b2 import contactListener, fixtureDef, polygonShape
except ImportError as e:
    raise DependencyNotInstalled(
        'Box2D is not installed, you can install it by run `pip install swig` followed by `pip install "gymnasium[box2d]"`'
    ) from e

try:
    # As pygame is necessary for using the environment (reset and step) even without a render mode
    #   therefore, pygame is a necessary import for the environment.
    import pygame
    from pygame import gfxdraw
except ImportError as e:
    raise DependencyNotInstalled(
        'pygame is not installed, run `pip install "gymnasium[box2d]"`'
    ) from e


STATE_W = 96  # less than Atari 160x192
STATE_H = 96
VIDEO_W = 600
VIDEO_H = 400
WINDOW_W = 1000
WINDOW_H = 800

SCALE = 6.0  # Track scale
TRACK_RAD = 900 / SCALE  # Track is heavily morphed circle with this radius
PLAYFIELD = 2000 / SCALE  # Game over boundary
FPS = 50  # Frames per second
ZOOM = 2.7  # Camera zoom
ZOOM_FOLLOW = True  # Set to False for fixed view (don't use zoom)


TRACK_DETAIL_STEP = 21 / SCALE
TRACK_TURN_RATE = 0.31
TRACK_WIDTH = 40 / SCALE
BORDER = 8 / SCALE
BORDER_MIN_COUNT = 4
GRASS_DIM = PLAYFIELD / 20.0
MAX_SHAPE_DIM = (
    max(GRASS_DIM, TRACK_WIDTH, TRACK_DETAIL_STEP) * math.sqrt(2) * ZOOM * SCALE
)


class FrictionDetector(contactListener):
    def __init__(self, env, lap_complete_percent):
        contactListener.__init__(self)
        self.env = env
        self.lap_complete_percent = lap_complete_percent

    def BeginContact(self, contact):
        self._contact(contact, True)

    def EndContact(self, contact):
        self._contact(contact, False)

    def _contact(self, contact, begin):
        tile = None
        obj = None
        u1 = contact.fixtureA.body.userData
        u2 = contact.fixtureB.body.userData
        if u1 and "road_friction" in u1.__dict__:
            tile = u1
            obj = u2
        if u2 and "road_friction" in u2.__dict__:
            tile = u2
            obj = u1
        if not tile:
            return

        # inherit tile color from env
        tile.color[:] = self.env.road_color
        if not obj or "tiles" not in obj.__dict__:
            return
        if begin:
            obj.tiles.add(tile)
            if not tile.road_visited:
                tile.road_visited = True
                self.env.reward += 1000.0 / len(self.env.track)
                self.env.tile_visited_count += 1

                print(
                    f"{self.lap_complete_percent=:<3} | {self.env.tile_visited_count / len(self.env.track)=:5f} | {tile.idx=:<3} | {self.env.tile_visited_count=:<3} | {len(self.env.track)=} | {tile.idx / len(self.env.track)=:5f}"
                )

                if tile.idx == 0:
                    print("~~~tile.idx Reset~~~")

                # Lap is considered completed if enough % of the track was covered
                if (
                    tile.idx == 0  # Original Code
                    #                    ((tile.idx / len(self.env.track)) > 0.99) # Quick Fix: Check if the car is close to the end of the track
                    and self.env.tile_visited_count / len(self.env.track)
                    > self.lap_complete_percent
                ):
                    print("Finished!")
                    self.env.new_lap = True
        else:
            obj.tiles.remove(tile)


class CarRacing(gym.Env, EzPickle):
    """
    ## Description
    The easiest control task to learn from pixels - a top-down
    racing environment. The generated track is random every episode.

    Some indicators are shown at the bottom of the window along with the
    state RGB buffer. From left to right: true speed, four ABS sensors,
    steering wheel position, and gyroscope.
    To play yourself (it's rather fast for humans), type:
    ```shell
    python gymnasium/envs/box2d/car_racing.py
    ```
    Remember: it's a powerful rear-wheel drive car - don't press the accelerator
    and turn at the same time.

    ## Action Space
    If continuous there are 3 actions :
    - 0: steering, -1 is full left, +1 is full right
    - 1: gas
    - 2: breaking

    If discrete there are 5 actions:
    - 0: do nothing
    - 1: steer left
    - 2: steer right
    - 3: gas
    - 4: brake

    ## Observation Space

    A top-down 96x96 RGB image of the car and race track.

    ## Rewards
    The reward is -0.1 every frame and +1000/N for every track tile visited, where N is the total number of tiles
     visited in the track. For example, if you have finished in 732 frames, your reward is 1000 - 0.1*732 = 926.8 points.

    ## Starting State
    The car starts at rest in the center of the road.

    ## Episode Termination
    The episode finishes when all the tiles are visited. The car can also go outside the playfield -
     that is, far off the track, in which case it will receive -100 reward and die.

    ## Arguments

    ```python
    >>> import gymnasium as gym
    >>> env = gym.make("CarRacing-v3", render_mode="rgb_array", lap_complete_percent=0.95, domain_randomize=False, continuous=False)
    >>> env
    <TimeLimit<OrderEnforcing<PassiveEnvChecker<CarRacing<CarRacing-v3>>>>>

    ```

    * `lap_complete_percent=0.95` dictates the percentage of tiles that must be visited by
     the agent before a lap is considered complete.

    * `domain_randomize=False` enables the domain randomized variant of the environment.
     In this scenario, the background and track colours are different on every reset.

    * `continuous=True` converts the environment to use discrete action space.
     The discrete action space has 5 actions: [do nothing, left, right, gas, brake].

    ## Reset Arguments

    Passing the option `options["randomize"] = True` will change the current colour of the environment on demand.
    Correspondingly, passing the option `options["randomize"] = False` will not change the current colour of the environment.
    `domain_randomize` must be `True` on init for this argument to work.

    ```python
    >>> import gymnasium as gym
    >>> env = gym.make("CarRacing-v3", domain_randomize=True)

    # normal reset, this changes the colour scheme by default
    >>> obs, _ = env.reset()

    # reset with colour scheme change
    >>> randomize_obs, _ = env.reset(options={"randomize": True})

    # reset with no colour scheme change
    >>> non_random_obs, _ = env.reset(options={"randomize": False})

    ```

    ## Version History
    - v2: Change truncation to termination when finishing the lap (1.0.0)
    - v1: Change track completion logic and add domain randomization (0.24.0)
    - v0: Original version

    ## References
    - Chris Campbell (2014), http://www.iforce2d.net/b2dtut/top-down-car.

    ## Credits
    Created by Oleg Klimov
    """

    metadata = {
        "render_modes": [
            "human",
            "rgb_array",
            "state_pixels",
        ],
        "render_fps": FPS,
    }

    def __init__(
        self,
        render_mode: Optional[str] = None,
        verbose: bool = False,
        lap_complete_percent: float = 0.95,
        domain_randomize: bool = False,
        continuous: bool = True,
    ):
        EzPickle.__init__(
            self,
            render_mode,
            verbose,
            lap_complete_percent,
            domain_randomize,
            continuous,
        )
        self.continuous = continuous
        self.domain_randomize = domain_randomize
        self.lap_complete_percent = lap_complete_percent
        self._init_colors()

        self.contactListener_keepref = FrictionDetector(self, self.lap_complete_percent)
        self.world = Box2D.b2World((0, 0), contactListener=self.contactListener_keepref)
        self.screen: Optional[pygame.Surface] = None
        self.surf = None
        self.clock = None
        self.isopen = True
        self.invisible_state_window = None
        self.invisible_video_window = None
        self.road = None
        self.car: Optional[Car] = None
        self.reward = 0.0
        self.prev_reward = 0.0
        self.verbose = verbose
        self.new_lap = False
        self.fd_tile = fixtureDef(
            shape=polygonShape(vertices=[(0, 0), (1, 0), (1, -1), (0, -1)])
        )

        # This will throw a warning in tests/envs/test_envs in utils/env_checker.py as the space is not symmetric
        #   or normalised however this is not possible here so ignore
        if self.continuous:
            self.action_space = spaces.Box(
                np.array([-1, 0, 0]).astype(np.float32),
                np.array([+1, +1, +1]).astype(np.float32),
            )  # steer, gas, brake
        else:
            self.action_space = spaces.Discrete(5)
            # do nothing, left, right, gas, brake

        self.observation_space = spaces.Box(
            low=0, high=255, shape=(STATE_H, STATE_W, 3), dtype=np.uint8
        )

        self.render_mode = render_mode

    def _destroy(self):
        if not self.road:
            return
        for t in self.road:
            self.world.DestroyBody(t)
        self.road = []
        assert self.car is not None
        self.car.destroy()

    def _init_colors(self):
        if self.domain_randomize:
            # domain randomize the bg and grass colour
            self.road_color = self.np_random.uniform(0, 210, size=3)

            self.bg_color = self.np_random.uniform(0, 210, size=3)

            self.grass_color = np.copy(self.bg_color)
            idx = self.np_random.integers(3)
            self.grass_color[idx] += 20
        else:
            # default colours
            self.road_color = np.array([102, 102, 102])
            self.bg_color = np.array([102, 204, 102])
            self.grass_color = np.array([102, 230, 102])

    def _reinit_colors(self, randomize):
        assert (
            self.domain_randomize
        ), "domain_randomize must be True to use this function."

        if randomize:
            # domain randomize the bg and grass colour
            self.road_color = self.np_random.uniform(0, 210, size=3)

            self.bg_color = self.np_random.uniform(0, 210, size=3)

            self.grass_color = np.copy(self.bg_color)
            idx = self.np_random.integers(3)
            self.grass_color[idx] += 20

    def _create_track(self):
        CHECKPOINTS = 12

        # Create checkpoints
        checkpoints = []
        for c in range(CHECKPOINTS):
            noise = self.np_random.uniform(0, 2 * math.pi * 1 / CHECKPOINTS)
            alpha = 2 * math.pi * c / CHECKPOINTS + noise
            rad = self.np_random.uniform(TRACK_RAD / 3, TRACK_RAD)

            if c == 0:
                alpha = 0
                rad = 1.5 * TRACK_RAD
            if c == CHECKPOINTS - 1:
                alpha = 2 * math.pi * c / CHECKPOINTS
                self.start_alpha = 2 * math.pi * (-0.5) / CHECKPOINTS
                rad = 1.5 * TRACK_RAD

            checkpoints.append((alpha, rad * math.cos(alpha), rad * math.sin(alpha)))
        self.road = []

        # Go from one checkpoint to another to create track
        x, y, beta = 1.5 * TRACK_RAD, 0, 0
        dest_i = 0
        laps = 0
        track = []
        no_freeze = 2500
        visited_other_side = False
        while True:
            alpha = math.atan2(y, x)
            if visited_other_side and alpha > 0:
                laps += 1
                visited_other_side = False
            if alpha < 0:
                visited_other_side = True
                alpha += 2 * math.pi

            while True:  # Find destination from checkpoints
                failed = True

                while True:
                    dest_alpha, dest_x, dest_y = checkpoints[dest_i % len(checkpoints)]
                    if alpha <= dest_alpha:
                        failed = False
                        break
                    dest_i += 1
                    if dest_i % len(checkpoints) == 0:
                        break

                if not failed:
                    break

                alpha -= 2 * math.pi
                continue

            r1x = math.cos(beta)
            r1y = math.sin(beta)
            p1x = -r1y
            p1y = r1x
            dest_dx = dest_x - x  # vector towards destination
            dest_dy = dest_y - y
            # destination vector projected on rad:
            proj = r1x * dest_dx + r1y * dest_dy
            while beta - alpha > 1.5 * math.pi:
                beta -= 2 * math.pi
            while beta - alpha < -1.5 * math.pi:
                beta += 2 * math.pi
            prev_beta = beta
            proj *= SCALE
            if proj > 0.3:
                beta -= min(TRACK_TURN_RATE, abs(0.001 * proj))
            if proj < -0.3:
                beta += min(TRACK_TURN_RATE, abs(0.001 * proj))
            x += p1x * TRACK_DETAIL_STEP
            y += p1y * TRACK_DETAIL_STEP
            track.append((alpha, prev_beta * 0.5 + beta * 0.5, x, y))
            if laps > 4:
                break
            no_freeze -= 1
            if no_freeze == 0:
                break

        # Find closed loop range i1..i2, first loop should be ignored, second is OK
        i1, i2 = -1, -1
        i = len(track)
        while True:
            i -= 1
            if i == 0:
                return False  # Failed
            pass_through_start = (
                track[i][0] > self.start_alpha and track[i - 1][0] <= self.start_alpha
            )
            if pass_through_start and i2 == -1:
                i2 = i
            elif pass_through_start and i1 == -1:
                i1 = i
                break
        if self.verbose:
            print("Track generation: %i..%i -> %i-tiles track" % (i1, i2, i2 - i1))
        assert i1 != -1
        assert i2 != -1

        track = track[i1 : i2 - 1]

        first_beta = track[0][1]
        first_perp_x = math.cos(first_beta)
        first_perp_y = math.sin(first_beta)
        # Length of perpendicular jump to put together head and tail
        well_glued_together = np.sqrt(
            np.square(first_perp_x * (track[0][2] - track[-1][2]))
            + np.square(first_perp_y * (track[0][3] - track[-1][3]))
        )
        if well_glued_together > TRACK_DETAIL_STEP:
            return False

        # Red-white border on hard turns
        border = [False] * len(track)
        for i in range(len(track)):
            good = True
            oneside = 0
            for neg in range(BORDER_MIN_COUNT):
                beta1 = track[i - neg - 0][1]
                beta2 = track[i - neg - 1][1]
                good &= abs(beta1 - beta2) > TRACK_TURN_RATE * 0.2
                oneside += np.sign(beta1 - beta2)
            good &= abs(oneside) == BORDER_MIN_COUNT
            border[i] = good
        for i in range(len(track)):
            for neg in range(BORDER_MIN_COUNT):
                border[i - neg] |= border[i]

        # Create tiles
        for i in range(len(track)):
            alpha1, beta1, x1, y1 = track[i]
            alpha2, beta2, x2, y2 = track[i - 1]
            road1_l = (
                x1 - TRACK_WIDTH * math.cos(beta1),
                y1 - TRACK_WIDTH * math.sin(beta1),
            )
            road1_r = (
                x1 + TRACK_WIDTH * math.cos(beta1),
                y1 + TRACK_WIDTH * math.sin(beta1),
            )
            road2_l = (
                x2 - TRACK_WIDTH * math.cos(beta2),
                y2 - TRACK_WIDTH * math.sin(beta2),
            )
            road2_r = (
                x2 + TRACK_WIDTH * math.cos(beta2),
                y2 + TRACK_WIDTH * math.sin(beta2),
            )
            vertices = [road1_l, road1_r, road2_r, road2_l]
            self.fd_tile.shape.vertices = vertices
            t = self.world.CreateStaticBody(fixtures=self.fd_tile)
            t.userData = t
            c = 0.01 * (i % 3) * 255
            t.color = self.road_color + c
            t.road_visited = False
            t.road_friction = 1.0
            t.idx = i
            t.fixtures[0].sensor = True
            self.road_poly.append(([road1_l, road1_r, road2_r, road2_l], t.color))
            self.road.append(t)
            if border[i]:
                side = np.sign(beta2 - beta1)
                b1_l = (
                    x1 + side * TRACK_WIDTH * math.cos(beta1),
                    y1 + side * TRACK_WIDTH * math.sin(beta1),
                )
                b1_r = (
                    x1 + side * (TRACK_WIDTH + BORDER) * math.cos(beta1),
                    y1 + side * (TRACK_WIDTH + BORDER) * math.sin(beta1),
                )
                b2_l = (
                    x2 + side * TRACK_WIDTH * math.cos(beta2),
                    y2 + side * TRACK_WIDTH * math.sin(beta2),
                )
                b2_r = (
                    x2 + side * (TRACK_WIDTH + BORDER) * math.cos(beta2),
                    y2 + side * (TRACK_WIDTH + BORDER) * math.sin(beta2),
                )
                self.road_poly.append(
                    (
                        [b1_l, b1_r, b2_r, b2_l],
                        (255, 255, 255) if i % 2 == 0 else (255, 0, 0),
                    )
                )
        self.track = track
        return True

    def reset(
        self,
        *,
        seed: Optional[int] = None,
        options: Optional[dict] = None,
    ):
        super().reset(seed=seed)
        self._destroy()
        self.world.contactListener_bug_workaround = FrictionDetector(
            self, self.lap_complete_percent
        )
        self.world.contactListener = self.world.contactListener_bug_workaround
        self.reward = 0.0
        self.prev_reward = 0.0
        self.tile_visited_count = 0
        self.t = 0.0
        self.new_lap = False
        self.road_poly = []

        if self.domain_randomize:
            randomize = True
            if isinstance(options, dict):
                if "randomize" in options:
                    randomize = options["randomize"]

            self._reinit_colors(randomize)

        while True:
            success = self._create_track()
            if success:
                break
            if self.verbose:
                print(
                    "retry to generate track (normal if there are not many"
                    "instances of this message)"
                )
        self.car = Car(self.world, *self.track[0][1:4])

        if self.render_mode == "human":
            self.render()
        return self.step(None)[0], {}

    def step(self, action: Union[np.ndarray, int]):
        assert self.car is not None
        if action is not None:
            action = action.astype(np.float64)
            if self.continuous:
                self.car.steer(-action[0])
                self.car.gas(action[1])
                self.car.brake(action[2])
            else:
                if not self.action_space.contains(action):
                    raise InvalidAction(
                        f"you passed the invalid action `{action}`. "
                        f"The supported action_space is `{self.action_space}`"
                    )
                self.car.steer(-0.6 * (action == 1) + 0.6 * (action == 2))
                self.car.gas(0.2 * (action == 3))
                self.car.brake(0.8 * (action == 4))

        self.car.step(1.0 / FPS)
        self.world.Step(1.0 / FPS, 6 * 30, 2 * 30)
        self.t += 1.0 / FPS

        self.state = self._render("state_pixels")

        step_reward = 0
        terminated = False
        truncated = False
        info = {}
        if action is not None:  # First step without action, called from reset()
            self.reward -= 0.1
            # We actually don't want to count fuel spent, we want car to be faster.
            # self.reward -=  10 * self.car.fuel_spent / ENGINE_POWER
            self.car.fuel_spent = 0.0
            step_reward = self.reward - self.prev_reward
            self.prev_reward = self.reward
            if self.tile_visited_count == len(self.track) or self.new_lap:
                # Termination due to finishing lap
                terminated = True
                info["lap_finished"] = True
            x, y = self.car.hull.position
            if abs(x) > PLAYFIELD or abs(y) > PLAYFIELD:
                terminated = True
                info["lap_finished"] = False
                step_reward = -100

        if self.render_mode == "human":
            self.render()
        return self.state, step_reward, terminated, truncated, info

    def render(self):
        if self.render_mode is None:
            assert self.spec is not None
            gym.logger.warn(
                "You are calling render method without specifying any render mode. "
                "You can specify the render_mode at initialization, "
                f'e.g. gym.make("{self.spec.id}", render_mode="rgb_array")'
            )
            return
        else:
            return self._render(self.render_mode)

    def _render(self, mode: str):
        assert mode in self.metadata["render_modes"]

        pygame.font.init()
        if self.screen is None and mode == "human":
            pygame.init()
            pygame.display.init()
            self.screen = pygame.display.set_mode((WINDOW_W, WINDOW_H))
        if self.clock is None:
            self.clock = pygame.time.Clock()

        if "t" not in self.__dict__:
            return  # reset() not called yet

        self.surf = pygame.Surface((WINDOW_W, WINDOW_H))

        assert self.car is not None
        # computing transformations
        angle = -self.car.hull.angle
        # Animating first second zoom.
        zoom = 0.1 * SCALE * max(1 - self.t, 0) + ZOOM * SCALE * min(self.t, 1)
        scroll_x = -(self.car.hull.position[0]) * zoom
        scroll_y = -(self.car.hull.position[1]) * zoom
        trans = pygame.math.Vector2((scroll_x, scroll_y)).rotate_rad(angle)
        trans = (WINDOW_W / 2 + trans[0], WINDOW_H / 4 + trans[1])

        self._render_road(zoom, trans, angle)
        self.car.draw(
            self.surf,
            zoom,
            trans,
            angle,
            mode not in ["state_pixels_list", "state_pixels"],
        )

        self.surf = pygame.transform.flip(self.surf, False, True)

        # showing stats
        self._render_indicators(WINDOW_W, WINDOW_H)

        font = pygame.font.Font(pygame.font.get_default_font(), 42)
        text = font.render("%04i" % self.reward, True, (255, 255, 255), (0, 0, 0))
        text_rect = text.get_rect()
        text_rect.center = (60, WINDOW_H - WINDOW_H * 2.5 / 40.0)
        self.surf.blit(text, text_rect)

        if mode == "human":
            pygame.event.pump()
            self.clock.tick(self.metadata["render_fps"])
            assert self.screen is not None
            self.screen.fill(0)
            self.screen.blit(self.surf, (0, 0))
            pygame.display.flip()
        elif mode == "rgb_array":
            return self._create_image_array(self.surf, (VIDEO_W, VIDEO_H))
        elif mode == "state_pixels":
            return self._create_image_array(self.surf, (STATE_W, STATE_H))
        else:
            return self.isopen

    def _render_road(self, zoom, translation, angle):
        bounds = PLAYFIELD
        field = [
            (bounds, bounds),
            (bounds, -bounds),
            (-bounds, -bounds),
            (-bounds, bounds),
        ]

        # draw background
        self._draw_colored_polygon(
            self.surf, field, self.bg_color, zoom, translation, angle, clip=False
        )

        # draw grass patches
        grass = []
        for x in range(-20, 20, 2):
            for y in range(-20, 20, 2):
                grass.append(
                    [
                        (GRASS_DIM * x + GRASS_DIM, GRASS_DIM * y + 0),
                        (GRASS_DIM * x + 0, GRASS_DIM * y + 0),
                        (GRASS_DIM * x + 0, GRASS_DIM * y + GRASS_DIM),
                        (GRASS_DIM * x + GRASS_DIM, GRASS_DIM * y + GRASS_DIM),
                    ]
                )
        for poly in grass:
            self._draw_colored_polygon(
                self.surf, poly, self.grass_color, zoom, translation, angle
            )

        # draw road
        for poly, color in self.road_poly:
            # converting to pixel coordinates
            poly = [(p[0], p[1]) for p in poly]
            color = [int(c) for c in color]
            self._draw_colored_polygon(self.surf, poly, color, zoom, translation, angle)

    def _render_indicators(self, W, H):
        s = W / 40.0
        h = H / 40.0
        color = (0, 0, 0)
        polygon = [(W, H), (W, H - 5 * h), (0, H - 5 * h), (0, H)]
        pygame.draw.polygon(self.surf, color=color, points=polygon)

        def vertical_ind(place, val):
            return [
                (place * s, H - (h + h * val)),
                ((place + 1) * s, H - (h + h * val)),
                ((place + 1) * s, H - h),
                ((place + 0) * s, H - h),
            ]

        def horiz_ind(place, val):
            return [
                ((place + 0) * s, H - 4 * h),
                ((place + val) * s, H - 4 * h),
                ((place + val) * s, H - 2 * h),
                ((place + 0) * s, H - 2 * h),
            ]

        assert self.car is not None
        true_speed = np.sqrt(
            np.square(self.car.hull.linearVelocity[0])
            + np.square(self.car.hull.linearVelocity[1])
        )

        # simple wrapper to render if the indicator value is above a threshold
        def render_if_min(value, points, color):
            if abs(value) > 1e-4:
                pygame.draw.polygon(self.surf, points=points, color=color)

        render_if_min(true_speed, vertical_ind(5, 0.02 * true_speed), (255, 255, 255))
        # ABS sensors
        render_if_min(
            self.car.wheels[0].omega,
            vertical_ind(7, 0.01 * self.car.wheels[0].omega),
            (0, 0, 255),
        )
        render_if_min(
            self.car.wheels[1].omega,
            vertical_ind(8, 0.01 * self.car.wheels[1].omega),
            (0, 0, 255),
        )
        render_if_min(
            self.car.wheels[2].omega,
            vertical_ind(9, 0.01 * self.car.wheels[2].omega),
            (51, 0, 255),
        )
        render_if_min(
            self.car.wheels[3].omega,
            vertical_ind(10, 0.01 * self.car.wheels[3].omega),
            (51, 0, 255),
        )

        render_if_min(
            self.car.wheels[0].joint.angle,
            horiz_ind(20, -10.0 * self.car.wheels[0].joint.angle),
            (0, 255, 0),
        )
        render_if_min(
            self.car.hull.angularVelocity,
            horiz_ind(30, -0.8 * self.car.hull.angularVelocity),
            (255, 0, 0),
        )

    def _draw_colored_polygon(
        self, surface, poly, color, zoom, translation, angle, clip=True
    ):
        poly = [pygame.math.Vector2(c).rotate_rad(angle) for c in poly]
        poly = [
            (c[0] * zoom + translation[0], c[1] * zoom + translation[1]) for c in poly
        ]
        # This checks if the polygon is out of bounds of the screen, and we skip drawing if so.
        # Instead of calculating exactly if the polygon and screen overlap,
        # we simply check if the polygon is in a larger bounding box whose dimension
        # is greater than the screen by MAX_SHAPE_DIM, which is the maximum
        # diagonal length of an environment object
        if not clip or any(
            (-MAX_SHAPE_DIM <= coord[0] <= WINDOW_W + MAX_SHAPE_DIM)
            and (-MAX_SHAPE_DIM <= coord[1] <= WINDOW_H + MAX_SHAPE_DIM)
            for coord in poly
        ):
            gfxdraw.aapolygon(self.surf, poly, color)
            gfxdraw.filled_polygon(self.surf, poly, color)

    def _create_image_array(self, screen, size):
        scaled_screen = pygame.transform.smoothscale(screen, size)
        return np.transpose(
            np.array(pygame.surfarray.pixels3d(scaled_screen)), axes=(1, 0, 2)
        )

    def close(self):
        if self.screen is not None:
            pygame.display.quit()
            self.isopen = False
            pygame.quit()

if __name__ == "__main__":
    a = np.array([0.0, 0.0, 0.0])

    def register_input():
        global quit, restart
        for event in pygame.event.get():
            if event.type == pygame.KEYDOWN:
                if event.key == pygame.K_LEFT:
                    a[0] = -1.0
                if event.key == pygame.K_RIGHT:
                    a[0] = +1.0
                if event.key == pygame.K_UP:
                    a[1] = +1.0
                if event.key == pygame.K_DOWN:
                    a[2] = +0.8  # set 1.0 for wheels to block to zero rotation
                if event.key == pygame.K_RETURN:
                    restart = True
                if event.key == pygame.K_ESCAPE:
                    quit = True

            if event.type == pygame.KEYUP:
                if event.key == pygame.K_LEFT:
                    a[0] = 0
                if event.key == pygame.K_RIGHT:
                    a[0] = 0
                if event.key == pygame.K_UP:
                    a[1] = 0
                if event.key == pygame.K_DOWN:
                    a[2] = 0

            if event.type == pygame.QUIT:
                quit = True

    env = CarRacing(render_mode="human")

    quit = False
    while not quit:
        env.reset()
        total_reward = 0.0
        steps = 0
        restart = False
        while True:
            register_input()
            s, r, terminated, truncated, info = env.step(a)
            total_reward += r
            if steps % 200 == 0 or terminated or truncated:
                print("\naction " + str([f"{x:+0.2f}" for x in a]))
                print(f"step {steps} total_reward {total_reward:+0.2f}")
            steps += 1
            if terminated or truncated or restart or quit:
                break
    env.close()
```
</details>

### Checklist

VincenzoPalma:
# Description

The environment will now end with `terminated = True` when the lap is completed after reaching the specified percentage of visited tiles.

Fixes #1269

## Type of change

# Checklist:

### DIFF
--- a/gym/utils/tests/test_env_checker.py
+++ b/gym/utils/tests/test_env_checker.py
@@ -1,0 +1,35 @@
+import gym
+import numpy as np
+import pytest
+from gym.spaces import Box, Dict, Discrete
+
+from gym.utils.env_checker import check_env
+
+
+class ActionDictTestEnv(gym.Env):
+    action_space = Dict({"position": Discrete(1), "velocity": Discrete(1)})
+    observation_space = Box(low=-1.0, high=2.0, shape=(3,), dtype=np.float32)
+
+    def step(self, action):
+        observation = np.array([1.0, 1.5, 0.5])
+        reward = 1
+        done = True
+        return observation, reward, done
+
+    def reset(self):
+        return np.array([1.0, 1.5, 0.5])
+
+    def render(self, mode="human"):
+        pass
+
+
+def test_check_env_dict_action():
+    # Environment.step() only returns 3 values: obs, reward, done. Not info!
+    test_env = ActionDictTestEnv()
+
+    with pytest.raises(AssertionError) as errorinfo:
+        check_env(env=test_env, warn=True)
+        assert (
+            str(errorinfo.value)
+            == "The `step()` method must return four values: obs, reward, done, info"
+        )
--- a/gym/utils/env_checker.py
+++ b/gym/utils/env_checker.py
@@ -1,0 +1,334 @@
+"""
+This file is originally from the Stable Baselines3 repository hosted on GitHub
+(https://github.com/DLR-RM/stable-baselines3/)
+Original Author: Antonin Raffin
+
+It also uses some warnings/assertions from the PettingZoo repository hosted on GitHub
+(https://github.com/PettingZoo-Team/PettingZoo)
+Original Author: Justin Terry
+
+These projects are covered by the MIT License.
+"""
+
+import warnings
+from typing import Union
+
+import gym
+import numpy as np
+from gym import spaces
+
+
+def _is_numpy_array_space(space: spaces.Space) -> bool:
+    """
+    Returns False if provided space is not representable as a single numpy array
+    (e.g. Dict and Tuple spaces return False)
+    """
+    return not isinstance(space, (spaces.Dict, spaces.Tuple))
+
+
+def _check_image_input(observation_space: spaces.Box, key: str = "") -> None:
+    """
+    Check that the input adheres to general standards
+    when the observation is apparently an image.
+    """
+    if observation_space.dtype != np.uint8:
+        warnings.warn(
+            f"It seems that your observation {key} is an image but the `dtype` "
+            "of your observation_space is not `np.uint8`. "
+            "If your observation is not an image, we recommend you to flatten the observation "
+            "to have only a 1D vector"
+        )
+
+    if np.any(observation_space.low != 0) or np.any(observation_space.high != 255):
+        warnings.warn(
+            f"It seems that your observation space {key} is an image but the "
+            "upper and lower bounds are not in [0, 255]. "
+            "Generally, CNN policies assume observations are within that range, "
+            "so you may encounter an issue if the observation values are not."
+        )
+
+
+def _check_nan(env: gym.Env, check_inf: bool = True) -> None:
+    """Check for NaN and Inf."""
+    for _ in range(10):
+        action = env.action_space.sample()
+        observation, reward, _, _ = env.step(action)
+
+        if np.any(np.isnan(observation)):
+            warnings.warn("Encountered NaN value in observations.")
+        if np.any(np.isnan(reward)):
+            warnings.warn("Encountered NaN value in rewards.")
+        if check_inf and np.any(np.isinf(observation)):
+            warnings.warn("Encountered inf value in observations.")
+        if check_inf and np.any(np.isinf(reward)):
+            warnings.warn("Encountered inf value in rewards.")
+
+
+def _check_obs(
+    obs: Union[tuple, dict, np.ndarray, int],
+    observation_space: spaces.Space,
+    method_name: str,
+) -> None:
+    """
+    Check that the observation returned by the environment
+    correspond to the declared one.
+    """
+    if not isinstance(observation_space, spaces.Tuple):
+        assert not isinstance(
+            obs, tuple
+        ), f"The observation returned by the `{method_name}()` method should be a single value, not a tuple"
+
+    # The check for a GoalEnv is done by the base class
+    if isinstance(observation_space, spaces.Discrete):
+        assert isinstance(
+            obs, int
+        ), f"The observation returned by `{method_name}()` method must be an int"
+    elif _is_numpy_array_space(observation_space):
+        assert isinstance(
+            obs, np.ndarray
+        ), f"The observation returned by `{method_name}()` method must be a numpy array"
+
+    assert observation_space.contains(
+        obs
+    ), f"The observation returned by the `{method_name}()` method does not match the given observation space"
+
+
+def _check_box_obs(observation_space: spaces.Box, key: str = "") -> None:
+    """
+    Check that the observation space is correctly formatted
+    when dealing with a ``Box()`` space. In particular, it checks:
+    - that the dimensions are big enough when it is an image, and that the type matches
+    - that the observation has an expected shape (warn the user if not)
+    """
+    # If image, check the low and high values, the type and the number of channels
+    # and the shape (minimal value)
+    if len(observation_space.shape) == 3:
+        _check_image_input(observation_space)
+
+    if len(observation_space.shape) not in [1, 3]:
+        warnings.warn(
+            f"Your observation {key} has an unconventional shape (neither an image, nor a 1D vector). "
+            "We recommend you to flatten the observation "
+            "to have only a 1D vector or use a custom policy to properly process the data."
+        )
+
+    if np.any(np.equal(observation_space.low, -np.inf)):
+        warnings.warn(
+            "Agent's minimum observation space value is -infinity. This is probably too low."
+        )
+    if np.any(np.equal(observation_space.high, np.inf)):
+        warnings.warn(
+            "Agent's maxmimum observation space value is infinity. This is probably too high"
+        )
+    if np.any(np.equal(observation_space.low, observation_space.high)):
+        warnings.warn("Agent's maximum and minimum observation space values are equal")
+    if np.any(np.greater(observation_space.low, observation_space.high)):
+        assert False, "Agent's minimum observation value is greater than it's maximum"
+    if observation_space.low.shape != observation_space.shape:
+        assert (
+            False
+        ), "Agent's observation_space.low and observation_space have different shapes"
+    if observation_space.high.shape != observation_space.shape:
+        assert (
+            False
+        ), "Agent's observation_space.high and observation_space have different shapes"
+
+
+def _check_box_action(action_space: spaces.Box):
+    if np.any(np.equal(action_space.low, -np.inf)):
+        warnings.warn(
+            "Agent's minimum action space value is -infinity. This is probably too low."
+        )
+    if np.any(np.equal(action_space.high, np.inf)):
+        warnings.warn(
+            "Agent's maxmimum action space value is infinity. This is probably too high"
+        )
+    if np.any(np.equal(action_space.low, action_space.high)):
+        warnings.warn("Agent's maximum and minimum action space values are equal")
+    if np.any(np.greater(action_space.low, action_space.high)):
+        assert False, "Agent's minimum action value is greater than it's maximum"
+    if action_space.low.shape != action_space.shape:
+        assert False, "Agent's action_space.low and action_space have different shapes"
+    if action_space.high.shape != action_space.shape:
+        assert False, "Agent's action_space.high and action_space have different shapes"
+
+
+def _check_normalized_action(action_space: spaces.Box):
+    if (
+        np.any(np.abs(action_space.low) != np.abs(action_space.high))
+        or np.any(np.abs(action_space.low) > 1)
+        or np.any(np.abs(action_space.high) > 1)
+    ):
+        warnings.warn(
+            "We recommend you to use a symmetric and normalized Box action space (range=[-1, 1]) "
+            "cf https://stable-baselines3.readthedocs.io/en/master/guide/rl_tips.html"
+        )
+
+
+def _check_returned_values(
+    env: gym.Env, observation_space: spaces.Space, action_space: spaces.Space
+) -> None:
+    """
+    Check the returned values by the env when calling `.reset()` or `.step()` methods.
+    """
+    # because env inherits from gym.Env, we assume that `reset()` and `step()` methods exists
+    obs = env.reset()
+
+    if isinstance(observation_space, spaces.Dict):
+        assert isinstance(
+            obs, dict
+        ), "The observation returned by `reset()` must be a dictionary"
+        for key in observation_space.spaces.keys():
+            try:
+                _check_obs(obs[key], observation_space.spaces[key], "reset")
+            except AssertionError as e:
+                raise AssertionError(f"Error while checking key={key}: " + str(e))
+    else:
+        _check_obs(obs, observation_space, "reset")
+
+    # Sample a random action
+    action = action_space.sample()
+    data = env.step(action)
+
+    assert (
+        len(data) == 4
+    ), "The `step()` method must return four values: obs, reward, done, info"
+
+    # Unpack
+    obs, reward, done, info = data
+
+    if isinstance(observation_space, spaces.Dict):
+        assert isinstance(
+            obs, dict
+        ), "The observation returned by `step()` must be a dictionary"
+        for key in observation_space.spaces.keys():
+            try:
+                _check_obs(obs[key], observation_space.spaces[key], "step")
+            except AssertionError as e:
+                raise AssertionError(f"Error while checking key={key}: " + str(e))
+
+    else:
+        _check_obs(obs, observation_space, "step")
+
+    # We also allow int because the reward will be cast to float
+    assert isinstance(
+        reward, (float, int, np.float32)
+    ), "The reward returned by `step()` must be a float"
+    assert isinstance(done, bool), "The `done` signal must be a boolean"
+    assert isinstance(
+        info, dict
+    ), "The `info` returned by `step()` must be a python dictionary"
+
+    if isinstance(env, gym.GoalEnv):
+        # For a GoalEnv, the keys are checked at reset
+        assert reward == env.compute_reward(
+            obs["achieved_goal"], obs["desired_goal"], info
+        )
+
+
+def _check_spaces(env: gym.Env) -> None:
+    """
+    Check that the observation and action spaces are defined
+    and inherit from gym.spaces.Space.
+    """
+    # Helper to link to the code, because gym has no proper documentation
+    gym_spaces = " cf https://github.com/openai/gym/blob/master/gym/spaces/"
+
+    assert hasattr(env, "observation_space"), (
+        "You must specify an observation space (cf gym.spaces)" + gym_spaces
+    )
+    assert hasattr(env, "action_space"), (
+        "You must specify an action space (cf gym.spaces)" + gym_spaces
+    )
+
+    assert isinstance(env.observation_space, spaces.Space), (
+        "The observation space must inherit from gym.spaces" + gym_spaces
+    )
+    assert isinstance(env.action_space, spaces.Space), (
+        "The action space must inherit from gym.spaces" + gym_spaces
+    )
+
+
+# Check render cannot be covered by CI
+def _check_render(
+    env: gym.Env, warn: bool = True, headless: bool = False
+) -> None:  # pragma: no cover
+    """
+    Check the declared render modes and the `render()`/`close()`
+    method of the environment.
+    :param env: The environment to check
+    :param warn: Whether to output additional warnings
+    :param headless: Whether to disable render modes
+        that require a graphical interface. False by default.
+    """
+    render_modes = env.metadata.get("render.modes")
+    if render_modes is None:
+        if warn:
+            warnings.warn(
+                "No render modes was declared in the environment "
+                " (env.metadata['render.modes'] is None or not defined), "
+                "you may have trouble when calling `.render()`"
+            )
+
+    else:
+        # Don't check render mode that require a
+        # graphical interface (useful for CI)
+        if headless and "human" in render_modes:
+            render_modes.remove("human")
+        # Check all declared render modes
+        for render_mode in render_modes:
+            env.render(mode=render_mode)
+        env.close()
+
+
+def check_env(env: gym.Env, warn: bool = True, skip_render_check: bool = True) -> None:
+    """
+    Check that an environment follows Gym API.
+    This is particularly useful when using a custom environment.
+    Please take a look at https://github.com/openai/gym/blob/master/gym/core.py
+    for more information about the API.
+    It also optionally check that the environment is compatible with Stable-Baselines.
+    :param env: The Gym environment that will be checked
+    :param warn: Whether to output additional warnings
+        mainly related to the interaction with Stable Baselines
+    :param skip_render_check: Whether to skip the checks for the render method.
+        True by default (useful for the CI)
+    """
+    assert isinstance(
+        env, gym.Env
+    ), "Your environment must inherit from the gym.Env class cf https://github.com/openai/gym/blob/master/gym/core.py"
+
+    # ============= Check the spaces (observation and action) ================
+    _check_spaces(env)
+
+    # Define aliases for convenience
+    observation_space = env.observation_space
+    action_space = env.action_space
+
+    # Warn the user if needed.
+    # A warning means that the environment may run but not work properly with popular RL libraries.
+    if warn:
+        obs_spaces = (
+            observation_space.spaces
+            if isinstance(observation_space, spaces.Dict)
+            else {"": observation_space}
+        )
+        for key, space in obs_spaces.items():
+            if isinstance(space, spaces.Box):
+                _check_box_obs(space, key)
+
+        # Check for the action space, it may lead to hard-to-debug issues
+        if isinstance(action_space, spaces.Box):
+            _check_box_action(action_space)
+            _check_normalized_action(action_space)
+
+    # ============ Check the returned values ===============
+    _check_returned_values(env, observation_space, action_space)
+
+    # ==== Check the render method and the declared render modes ====
+    if not skip_render_check:
+        _check_render(env, warn=warn)  # pragma: no cover
+
+    # The check only works with numpy arrays
+    if _is_numpy_array_space(observation_space) and _is_numpy_array_space(action_space):
+        _check_nan(env)
--- a/gym/utils/env_checker.py
+++ b/gym/utils/env_checker.py
@@ -5,7 +5,7 @@ Original Author: Antonin Raffin
 
 It also uses some warnings/assertions from the PettingZoo repository hosted on GitHub
 (https://github.com/PettingZoo-Team/PettingZoo)
-Original Author: Justin Terry
+Original Author: J K Terry
 
 These projects are covered by the MIT License.
 """
--- a/gymnasium/envs/box2d/car_racing.py
+++ b/gymnasium/envs/box2d/car_racing.py
@@ -145,7 +145,7 @@ class CarRacing(gym.Env, EzPickle):
     The car starts at rest in the center of the road.
 
     ## Episode Termination
-    The episode finishes when all the tiles are visited. The car can also go outside the playfield -
+    The episode finishes when the car reaches the end of the circuit or the time limit is reached. The car can also go outside the playfield -
      that is, far off the track, in which case it will receive -100 reward and die.
 
     ## Arguments
@@ -189,6 +189,8 @@ class CarRacing(gym.Env, EzPickle):
     ```
 
     ## Version History
+    - v4: The episode will now terminate upon completing the lap, regardless of the number of tiles visited
+    - v3: Improved RNG handling with seed=-1; fixed bugs in environment checking and Box shape inference; minor fixes, including typos and dependency management.
     - v2: Change truncation to termination when finishing the lap (1.0.0)
     - v1: Change track completion logic and add domain randomization (0.24.0)
     - v0: Original version
--- a/gymnasium/envs/box2d/car_racing.py
+++ b/gymnasium/envs/box2d/car_racing.py
@@ -85,20 +85,21 @@ class FrictionDetector(contactListener):
         tile.color[:] = self.env.road_color
         if not obj or "tiles" not in obj.__dict__:
             return
+
+        # Lap is considered completed if enough % of the track was covered
+        if (
+            tile.idx == 0
+            and self.env.tile_visited_count / len(self.env.track)
+            > self.lap_complete_percent
+        ):
+            self.env.new_lap = True
+
         if begin:
             obj.tiles.add(tile)
             if not tile.road_visited:
                 tile.road_visited = True
                 self.env.reward += 1000.0 / len(self.env.track)
                 self.env.tile_visited_count += 1
-
-                # Lap is considered completed if enough % of the track was covered
-                if (
-                    tile.idx == 0
-                    and self.env.tile_visited_count / len(self.env.track)
-                    > self.lap_complete_percent
-                ):
-                    self.env.new_lap = True
         else:
             obj.tiles.remove(tile)
 


pseudo-rnd-thoughts:
@VincenzoPalma To clarify what was the previous behaviour when the agent had cross that the lap completion percentage and got to the end? 

Could you train an agent using PPO from SB3 and share the training graphes of the old and new versions?

VincenzoPalma:
@pseudo-rnd-thoughts The previous behavior in that scenario was that the environment would not terminate upon completing the lap but would instead continue until reaching the time limit, at which point it would end with `truncated = True`.

I will try and train an agent as you suggest as soon as i can.

pseudo-rnd-thoughts:
@VincenzoPalma Have you had any time to train agents for the different Car Racing versions?

pseudo-rnd-thoughts:
Thanks for doing that @VincenzoPalma, keep me updated here or on discord if you are uncertain how to get some working

VincenzoPalma:
Can i share TensorBoard logs to show you the training graphs?

pseudo-rnd-thoughts:
You can either share images on GitHub or message me on discord and that I can look at the files

VincenzoPalma:
I'll share some images as soon as I obtain the graphs with 75 as minimum percentage of visited tiles. I'll also share the code to see if it's correct for the task and to receive feedback.

VincenzoPalma:
So, here's the code that i used:
```
import torch
from stable_baselines3 import PPO
from stable_baselines3.common.monitor import Monitor
from stable_baselines3.common.vec_env import DummyVecEnv
from stable_baselines3.common.callbacks import EvalCallback, CheckpointCallback
from stable_baselines3.common.logger import configure

import gymnasium as gym

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

n_envs = 16

def make_env():
    env = gym.make("CarRacing-v3", lap_complete_percent=0.25)
    return Monitor(env)

env = DummyVecEnv([make_env for _ in range(n_envs)])

log_dir = "./logs/"
new_logger = configure(log_dir, ["tensorboard"])

new_model = PPO("CnnPolicy", env, verbose=1, tensorboard_log=log_dir, device=device)
new_model.set_logger(new_logger)

eval_callback = EvalCallback(env, best_model_save_path='./logs/best_model',
                             log_path='./logs/results', eval_freq=10000,
                             deterministic=True, render=False)

checkpoint_callback = CheckpointCallback(save_freq=10000, save_path='./logs/',
                                         name_prefix='ppo_model')

new_model.learn(total_timesteps=500_000)

new_model_name = "ppo_car_gray_new252"
new_model.save(new_model_name)
print(f"New model saved as {new_model_name}")
```

So far, I've trained two agents: one before the change and one after, both with 75 lap complete percent.
Here are some graphs of the agent before the change:
image
image

Graphs of the agent after the change:
image
image

What stands out the most to me is the difference in the mean reward per episode. It looks better in the new version of the game, probably because the game now ends correctly before the agent can take any action that would give it a negative reward.

I'll wait for your feedback on the code and the data.

VincenzoPalma:
My initial assumption is that 500k steps might not be sufficient to achieve a positive reward. I initially set 500k steps mainly to compare the agents before and after the change, but I could try increasing it to a value more in line with the numbers used in the benchmarks you shared to see if it yields better results.

pseudo-rnd-thoughts:
Ohh that is way better results. Yeah, if you can do that, use the same hyperparameters and 3 different percentages with both the old and new environments to show the differences. 
We want to demonstrate that the environment changes make sense and don't break something unexpectedly

VincenzoPalma:
Here's the train graphs for 25, 50 and 75 percentages. Looks like old version achieves better rewards at 25 and 50 and similar rewards at 75.
image
image
image
image
image
image

AUnicyclingProgrammer:
I've been following your progress and wanted to pass along something that I noticed when I was training my agent.

I think that v3 of the environment provides a better training environment than our current test version.

When I was training my agent, I considered modifying the environment such that every time the agent crossed the start/finish line the environment checked to see if the agent had covered enough tiles to consider the lap complete, but didn't terminate the race unless the agent had covered enough tiles. My thinking was that this would help the agent learn to stay on the road because the agent could continue learning for the entirety of the time limit. If you started training with a generous time limit (say 5000 steps) an agent could still "complete the race" even if it took 2 or 3 laps to do so.

This would require modifying the truncation and termination conditions we defined in #1269, but I think it'd also speed up training and improve agent performance. My concern is that these changes may be significant enough to create a fundamental difference between the way the current versions of the environment operate compared to future versions were this change to be implemented.

pseudo-rnd-thoughts:
Its just "pseudornd (mark)"

AUnicyclingProgrammer:
I recently found the time to resume working on project I where I first discovered the issue #1269 and remembered why I submitted the issue in the first place. I first noticed that the environment was resetting prematurely because the info dictionary never contained the `lap_finished` key which it should have been returning [according to the code](https://github.com/Farama-Foundation/Gymnasium/blob/675945f1054747e70c5993753b3371a2b7baf6a9/gymnasium/envs/box2d/car_racing.py#L579). I now suspect that this issue may be related to Gymnasium itself as opposed to the Car Racing environment, but I haven't conducted any further tests to confirm this theory.

Additionally, I made some custom modifications to the environment which appear to accelerate/improve the training process.
- Episodes are forcibly terminated if the agent doesn't make any progress (earn a positive reward) during 250 timesteps. No additional penalty is applied
- If the agent remains stationary for 100 consecutive timesteps an additional reward of -0.0125 is applied each timestep until the agent moves again, at which point the 100 timestep timer "resets"
- An additional reward -0.0125 is applied for each timestep that the agent is spinning out as determined by monitoring the ABS sensors.(both front ABS sensors have a reading below 50 while both rear ABS sensors have a value above 150)

I originally implemented these changes to speed up the training process, but they may be unnecessary because I didn't realize I needed to normalize the agent rewards until recently.

AUnicyclingProgrammer:
@VincenzoPalma @pseudo-rnd-thoughts I did some additional testing and no longer think the odd reset behavior we're seeing is unique to Car Racing. I posted my [experimental results](https://github.com/Farama-Foundation/Gymnasium/issues/1269#issuecomment-2923048880) in the original issue thread because that felt like the most appropriate place to put them.

jkterry1:
@VincenzoPalma could you please rebase this PR off main?

jkterry1:
@VincenzoPalma would you be so kind as to rebase your PR off main and get CI to pass so we can review it?

VincenzoPalma:
> would you be so kind as to rebase your PR off main and get CI to pass so we can review it?

Hi! I'm completely swamped right now and don't have time to look into this. It's been a while, and I no longer have the local project environment set up. I'm really sorry about that, but I'll try to take a look at it as soon as I possibly can

gym/utils/tests/test_env_checker.py
```diff
@@ -0,0 +1,35 @@
...
+    def render(self, mode="human"):
+        pass
+
+
+def test_check_env_dict_action():
+    # Environment.step() only returns 3 values: obs, reward, done. Not info!
+    test_env = ActionDictTestEnv()
+
+    with pytest.raises(AssertionError) as errorinfo:
+        check_env(env=test_env, warn=True)
```
Sanjays2402, line 31: this assertion is unreachable: once `check_env` raises, control leaves the `with` block. `pytest.raises(AssertionError, match=...)` would actually verify the diagnostic.

jkterry1:
Closing in favor of #1670

PR #1334, opened 2025-03-20
Update toytext.CliffWalking env version in envs registry

abouelsaadat:
# Description

Update toytext.CliffWalking env version in envs registry,
Motivation:
the example included in the documentation is not working.
https://gymnasium.farama.org/environments/toy_text/cliff_walking/
```python
    import gymnasium as gym
    gym.make('CliffWalking-v1')
```

## Type of change

### DIFF
--- a/gymnasium/envs/__init__.py
+++ b/gymnasium/envs/__init__.py
@@ -147,8 +147,14 @@ register(
 )
 
 register(
-    id="CliffWalking-v0",
+    id="CliffWalking-v1",
     entry_point="gymnasium.envs.toy_text.cliffwalking:CliffWalkingEnv",
+)
+
+register(
+    id="CliffWalkingSlippery-v1",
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Thanks for this PR; this was more of a rollercoaster than I expected. 

You are right that the documentation says that this should be `CliffWalking-v1`; however, when "version" change was made (https://github.com/openai/gym/pull/2619/files and https://github.com/openai/gym/releases/tag/0.19.0), the **actual** version number wasn't updated, which this PR rectifies. 

The problem is that this "v1" change was made 3 years in Gym v0.19, therefore, sadly, we should actually update the documentation to correct inform people of what was happened. 

@abouelsaadat would you be happy to modify your PR for this?

pseudo-rnd-thoughts:
@abouelsaadat Apologies, I must have been much too tired to think straight. 
Yes, your right. I'll update

pseudo-rnd-thoughts (approved):

PR #1382, opened 2025-05-13
Reenable `reportInvalidTypeForm` typing rule

Jammf:
# Description

Reenables the `reportInvalidTypeForm` pyright rule.

Since `mujoco_py_env.py` is no longer being maintained, I chose to silence all typing errors from the file instead of fixing the type annotation (on L322).

## Type of change

# Checklist:

- [x] ~I have commented my code, particularly in hard-to-understand areas~
- [x] ~I have made corresponding changes to the documentation~
- [x] ~I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 import operator as op
 import typing
 from functools import reduce, singledispatch
-from typing import Any, TypeVar, Union, cast
+from typing import Any, TypeVar, Union
 
 import numpy as np
 from numpy.typing import NDArray
@@ -329,7 +329,7 @@ def _unflatten_multidiscrete(
             f"{x} is not a concatenation of one-hot encoded vectors and can not be unflattened to space {space}. "
             "Not all valid samples in a flattened space can be unflattened."
         )
-    (indices,) = cast(type(offsets[:-1]), nonzero)
+    (indices,) = nonzero
     return (
         np.asarray(indices - offsets[:-1], dtype=space.dtype).reshape(space.shape)
         + space.start
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -323,13 +323,12 @@ def _unflatten_multidiscrete(
 ) -> NDArray[np.integer[Any]]:
     offsets = np.zeros((space.nvec.size + 1,), dtype=space.dtype)
     offsets[1:] = np.cumsum(space.nvec.flatten())
-    nonzero = np.nonzero(x)
-    if len(nonzero[0]) == 0:
+    (indices,) = np.nonzero(x)
+    if len(indices) == 0:
         raise ValueError(
             f"{x} is not a concatenation of one-hot encoded vectors and can not be unflattened to space {space}. "
             "Not all valid samples in a flattened space can be unflattened."
         )
-    (indices,) = nonzero
     return (
         np.asarray(indices - offsets[:-1], dtype=space.dtype).reshape(space.shape)
         + space.start
--- a/gymnasium/utils/play.py
+++ b/gymnasium/utils/play.py
@@ -6,6 +6,7 @@ from collections import deque
 from typing import Callable, List
 
 import numpy as np
+from matplotlib.axes import Axes
 
 import gymnasium as gym
 from gymnasium import Env, logger
@@ -369,7 +370,7 @@ class PlayPlot:
         for axis, name in zip(self.ax, plot_names):
             axis.set_title(name)
         self.t = 0
-        self.cur_plot: list[plt.Axes | None] = [None for _ in range(num_plots)]
+        self.cur_plot: list[Axes | None] = [None for _ in range(num_plots)]
         self.data = [deque(maxlen=horizon_timesteps) for _ in range(num_plots)]
 
     def callback(
--- a/gymnasium/utils/play.py
+++ b/gymnasium/utils/play.py
@@ -3,15 +3,18 @@
 from __future__ import annotations
 
 from collections import deque
-from typing import Callable, List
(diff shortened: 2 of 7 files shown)


gymnasium/utils/play.py
```diff
@@ -6,6 +6,7 @@
 from typing import Callable, List
 
 import numpy as np
+from matplotlib.axes import Axes
```
pseudo-rnd-thoughts, line 9: This should be in `if TYPE_CHECKING` as I believe this will require users to install matplotlib as its imported in `import gymnasium`

gymnasium/spaces/utils.py
```diff
@@ -329,7 +329,7 @@ def _unflatten_multidiscrete(
             f"{x} is not a concatenation of one-hot encoded vectors and can not be unflattened to space {space}. "
             "Not all valid samples in a flattened space can be unflattened."
         )
-    (indices,) = cast(type(offsets[:-1]), nonzero)
```
pseudo-rnd-thoughts, line 332: Could you explain this change?

pseudo-rnd-thoughts (changes requested):
Generally looks good. Just two questions

Jammf:
I'm pretty sure CI diverged from my local pre-commit because the CI env doesn't have dependencies installed. Testing it locally, that causes symbols coming from dependencies to be `Unknown`, which doesn't sound ideal. But fixing it should just be a matter of adding `- run: pip install .[all]` to the pre-commit workflow.

gymnasium/spaces/utils.py
pseudo-rnd-thoughts: Could we just go straight from np.nonzero to (indices,)?

pseudo-rnd-thoughts (changes requested):
Thanks for the changes, could we simply the nonzero code. 
For pre-commit, we've had this issue in the past but I can't remember the solution. As we ignore import module then that shouldn't solve it in my head

gymnasium/spaces/utils.py
Jammf: Makes sense, I've added it in the latest commit.

Jammf:
Okay, simplified the nonzero code.

It looks like `reportMissingImports = "none"` silences the warning that imports from missing dependencies were replaced with `Unknown`, but if those dependencies are available pyright will still use them during type checking. So to me, making CI match dev environments would make things more consistent, and a bit safer (since someone could theoretically skip their local pre-commit and commit code that only type checks without dependencies, and CI would miss it).

<details>
<summary>Output comparison for pyright + mypy</summary>

## Output comparison for pyright + mypy

When pyright or mypy are installed in an isolated environment, such as with `uvx`, they'll try to use dependencies from that isolated environment unless they're explicitly told to use a different one (i.e. the dev environment). So it's a good proxy for evaluating how they behave in current CI.

### Pyright
Missing dependencies:
<pre>
> uvx pyright gymnasium/spaces/multi_binary.py 
gymnasium/spaces/multi_binary.py
  gymnasium/spaces/multi_binary.py:62:17 - information: Type of "MaskNDArray" is <b>"Unknown"</b>
  gymnasium/spaces/multi_binary.py:64:21 - error: Variable not allowed in type expression (reportInvalidTypeForm)
  gymnasium/spaces/multi_binary.py:64:61 - error: Variable not allowed in type expression (reportInvalidTypeForm)
2 errors, 0 warnings, 1 information 
</pre>

Dependencies installed:
<pre>
> uvx pyright <b>--pythonpath=.venv/bin/python</b> gymnasium/spaces/multi_binary.py
gymnasium/spaces/multi_binary.py
  gymnasium/spaces/multi_binary.py:62:17 - information: Type of "MaskNDArray" is <b>"type[ndarray[tuple[int, ...], dtype[signedinteger[_8Bit]]]]"</b>
0 errors, 0 warnings, 1 information 
</pre>


### Mypy
Missing dependencies:
<pre>
> uvx mypy --follow-imports=silent --ignore-missing-imports --disable-error-code=assignment gymnasium/spaces/multi_binary.py      
gymnasium/spaces/multi_binary.py:62: note: Revealed type is <b>"Any"</b>
gymnasium/spaces/multi_binary.py:64: error: Variable "gymnasium.spaces.space.MaskNDArray" is not valid as a type  [valid-type]
gymnasium/spaces/multi_binary.py:64: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 1 error in 1 file (checked 1 source file)
</pre>

Dependencies installed:
<pre>
> uvx mypy <b>--python-executable=.venv/bin/python</b> --follow-imports=silent --ignore-missing-imports --disable-error-code=assignment gymnasium/spaces/multi_binary.py
gymnasium/spaces/multi_binary.py:62: note: Revealed type is <b>"*snip* -> numpy.ndarray[builtins.tuple[builtins.int, ...], numpy.dtype[numpy.signedinteger[numpy._typing._nbit_base._8Bit]]]"</b>
Success: no issues found in 1 source file
</pre>

</details>

Jammf:
Found another way to get around `MaskNDArray` not being valid in CI, by explicitly marking it as a `TypeAlias` (thanks to [this comment](https://redirect.github.com/microsoft/pyright/issues/9051#issuecomment-2368944137)). Note the caveat:

> This will prevent pyright (and other type checkers) from assuming this symbol is a variable, but without [access to type stubs], the type alias will have little or no utility because it will represent some type that is unknown at static analysis time.

With access to numpy, type checkers should be able to correctly determine that `MaskNDArray` is a type alias even without the explicit `TypeAlias` annotation. So this fix is really just a band-aid specifically for CI, for the case where pyright (or mypy) is being run without numpy installed.

pseudo-rnd-thoughts (approved):

PR #1385, opened 2025-05-20
Add compatibility note for Python versions in usage documentation

honoursbhaduria:
#Changes Introduced
1. docs/introduction/basic_usage.md

    =>  Added a note block indicating Python 3.10+ is required.

2. docs/introduction/create_custom_env.md

    => Added the same Python compatibility note for consistency.

### DIFF
--- a/docs/introduction/basic_usage.md
+++ b/docs/introduction/basic_usage.md
@@ -3,8 +3,9 @@ layout: "contents"
 title: Basic Usage
 firstpage:
 ---
+<!-- See docs/tutorials/run_python310_example.py for an executable version of this code. -->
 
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for the Pr, I agree with the idea but it would be better if we had some automated testing of this like the tutorials in https://github.com/Farama-Foundation/Gymnasium/pull/1342 
It would be also helpful to note that this code is only guaranteed to work with the latest version.

honoursbhaduria:
Added a tutorial for creating a custom Gymnasium environment, including a tested example script ([run_python310_example.py](https://miniature-space-train-x5vxvqx7p44wcg9p.github.dev/)). Updated documentation to note version requirements and CI integration.

PR #1384, opened 2025-05-17
enum? not in discrete action space unless first converted to 'int' or…

zbenmo:
enum? not in discrete action space unless first converted to 'int' or… used with accessor .value

# Description

I'm trying to play (utils.play) "ALE/MontezumaRevenge-v5"

I got an assertion error: "assert action in env.action_space" 

Apparently it was Action.NOPE or so, that is indeed 0 (if you look at its value), but is not in 0..17

Fixes # (issue)

## Type of change

casting action to int

Please delete options that are not relevant.

- [ ] Documentation only change (no code changed)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Screenshots

Please attach before and after screenshots of the change if applicable.

# Checklist:

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1391, opened 2025-05-24
Re-enable `reportGeneralTypeIssues` typing rule

Jammf:
# Description

Re-enables the `reportGeneralTypeIssues` Pyright rule. 

Generally, type annotations were changed to match the implementation. One exception was `play.play()`, which had the implementation changed to allow the `keys_to_action` dict to also have `int` keys (per https://github.com/Farama-Foundation/Gymnasium/pull/1384#issuecomment-2906722386).

## Type of change

# Checklist:

- [x] ~I have commented my code, particularly in hard-to-understand areas~
- [x] ~I have made corresponding changes to the documentation~

### DIFF
--- a/gymnasium/spaces/multi_binary.py
+++ b/gymnasium/spaces/multi_binary.py
@@ -39,13 +39,17 @@ class MultiBinary(Space[NDArray[np.int8]]):
                 or some sort of sequence (tuple, list or np.ndarray) if there are multiple axes.
             seed: Optionally, you can use this argument to seed the RNG that is used to sample from the space.
         """
-        if isinstance(n, (Sequence, np.ndarray)):
-            self.n = input_n = tuple(int(i) for i in n)
-            assert (np.asarray(input_n) > 0).all()  # n (counts) have to be positive
-        else:
+        if isinstance(n, int):
             self.n = n = int(n)
             input_n = (n,)
             assert (np.asarray(input_n) > 0).all()  # n (counts) have to be positive
+        elif isinstance(n, (Sequence, np.ndarray)):
+            self.n = input_n = tuple(int(i) for i in n)
+            assert (np.asarray(input_n) > 0).all()  # n (counts) have to be positive
+        else:
+            raise ValueError(
+                f"Expected n to be an int or a sequence of ints, actual type: {type(n)}"
+            )
 
         super().__init__(input_n, np.int8, seed)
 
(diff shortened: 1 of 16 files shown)


gymnasium/spaces/multi_binary.py
```diff
@@ -39,13 +39,17 @@ def __init__(
                 or some sort of sequence (tuple, list or np.ndarray) if there are multiple axes.
             seed: Optionally, you can use this argument to seed the RNG that is used to sample from the space.
         """
-        if isinstance(n, (Sequence, np.ndarray)):
-            self.n = input_n = tuple(int(i) for i in n)
-            assert (np.asarray(input_n) > 0).all()  # n (counts) have to be positive
-        else:
```
Jammf, line 45: This was to fix a type error that only occurs in CI: if numpy is not installed then `np.ndarray` becomes `Any`, and pyright assumes `n: int` will match the `if` block instead of the `else` one. (But it goes into the correct block in a dev environment, if it can access annotations from `numpy`)

gymnasium/vector/async_vector_env.py
```diff
@@ -250,7 +250,7 @@ def np_random(self) -> tuple[np.random.Generator, ...]:
     def reset(
         self,
         *,
-        seed: int | list[int] | None = None,
+        seed: int | list[int | None] | None = None,
```
Jammf, line 253: This could also be typed as `int | list[int] | list[None]| None` if we want to soft-disallow specifying seeds for a subset of envs.

Jammf:
Not sure why checks are failing. Seems to be coming from Mujoco, so it might be unrelated?

gymnasium/spaces/multi_binary.py
pseudo-rnd-thoughts: Ohh that makes sense. Would installing NumPy in pre commit help?

gymnasium/vector/async_vector_env.py
pseudo-rnd-thoughts: No I think that's fine

gymnasium/spaces/multi_binary.py
Jammf: I think it'd be a good. It'd make things simpler overall, since we wouldn't need to do CI-specific hacks to get things to type check with numpy. And typing would be a bit safer in CI too, since we wouldn't get things silently replaced with `Any` if the libraries aren't available.

pseudo-rnd-thoughts: @Jammf would you be able to add that in another or

Issue #842, opened 2023-12-13
[Proposal] Mark some functions & attributes of the Env API as optional


Kallinteris-Andreas:
### Proposal

# current API
The current `Env` API contains:
Methods
-    Env.step()
-    Env.reset()
-    Env.render()
-    Env.close()

Attributes
-    Env.action_space
-    Env.observation_space
-    Env.metadata
-    Env.render_mode
-    Env.spec
-    Env.unwrapped
-    Env.np_random

but technically, the only "essential" ("essential" as in minimum every environment must implement to be learning compatible) parts of the API are:

Methods
-    Env.step()
-    Env.reset()

Attributes
-    Env.action_space
-    Env.observation_space
-    Env.spec 
-    Env.unwrapped

# Proposed API
my proposal is simply marked some parts of the API as optional

Methods
-    Env.step()
-    Env.reset()
-    Env.render()  (OPTIONAL: for rendering the environment)
-    Env.close() (OPTIONAL: for rendering the environment)

Attributes
-    Env.action_space
-    Env.observation_space
-    Env.metadata (OPTIONAL: for rendering the environment)
-    Env.render_mode (OPTIONAL: for rendering the environment)
-    Env.spec (TECHNICALLY OPTIONAL: for registering the environment)
-    Env.unwrapped
-    Env.np_random (TECHNICALLY OPTIONAL: only for stochastic environments)

or alternatively presented as:
Core API Methods & Attributes
-    Env.step()
-    Env.reset()
-    Env.action_space
-    Env.observation_space
-    Env.spec
-    Env.unwrapped

Render extension API Methods & Attributes
-    Env.render()
-    Env.close()
-    Env.metadata
-    Env.render_mode

other API Methods & Attributes
-    Env.np_random

# notes
1. I do not care how it is presented, I just propose some parts of the API to be 
2. this does not have to be included in the 1.0 release, it could be in 1.1 or 1.2
3. The implementation should be to simply `raise NotImplementedError` 

# Downside
1) `check_env()` would have to be more complex
2) ???

### Motivation

This would make extending the API a lot easier since, it would not be a requirement for all the environments to implement those extensions.

example https://github.com/Farama-Foundation/Gymnasium/issues/94, https://github.com/Farama-Foundation/Gymnasium/issues/737:

state_clone_restrore extension API Methods & Attributes
-  clone_state()
-  restore_state()

### Alternatives

[...]

PR #1399, opened 2025-06-08
[Array-API] Add support for Box space

amacati:
# Description

This PR is a first shot at making spaces generic over all Array API compatible frameworks (e.g. `numpy`, `jax`, `torch`, ...).

The benefit of this change would be to make gymnasium compatible with the growing list of environments written in `jax` or `torch`. Strictly speaking, the `ArrayConversion` wrappers (e.g. `JaxToTorch`) are already outside of what we can currently express with `gymnasium.spaces`: There is no way to let the action and observation spaces reflect that we are expecting `jax.Array`s as observations and `torch.Tensor`s as inputs (safe for custom spaces which are rarely used).

This PR is also meant as a basis for discussion if things require breaking changes, new core dependencies (i.e. `array_api_compat` would become a core dependency) etc. 

One fundamental change is that each box is now also linked with a `device`. This allows users to express that e.g. the observation is an array on the GPU. The example below shows how this impacts `space.contains`.

## Open Challenges
- [ ] `space.sample`: Sampling will require several changes. It does not make sense to seed a `torch` space with a `numpy` random generator
- [ ] All tests can be rewritten to test against all Array API compatible frameworks. I would like to postpone this until we have a clearer picture of how things should be implemented though

## Type of change

### Example
With the changes in this PR, it is already possible to run the following:

```python
import array_api_strict as xp
import jax.numpy as jp
import numpy as np
import torch

box = Box(low=np.zeros(3, dtype=np.float32), high=np.zeros(3, dtype=np.float32))
print(box)  # Box(0.0, 0.0, (3,), <class 'numpy.float32'>, cpu)

strict_box = Box(low=xp.zeros(3, dtype=xp.float32), high=xp.zeros(3, dtype=xp.float32))
print(strict_box)  # Box(Array(0., dtype=array_api_strict.float32), Array(0., dtype=array_api_strict.float32), (3,), array_api_strict.float32, array_api_strict.Device('CPU_DEVICE'))

torch_box = Box(low=torch.zeros(3), high=torch.ones(3))
print(torch_box)  # Box(tensor(0.), tensor(1.), (3,), torch.float32, cpu)

jax_box = Box(low=jp.zeros(3), high=jp.ones(3))
print(jax_box)  # Box(0.0, 1.0, (3,), <class 'jax.numpy.float32'>, cuda:0)

cpu_device = jax.devices("cpu")[0]
print(jax_box.contains(jp.array([0.5, 0.5, 0.5], device=cpu_device)))  # False, devices do not match
gpu_device = jax.devices("gpu")[0]
print(jax_box.contains(jp.array([0.5, 0.5, 0.5], device=gpu_device)))  # True
```

# Checklist:

### DIFF
--- a/gymnasium/spaces/oneof.py
+++ b/gymnasium/spaces/oneof.py
@@ -48,9 +48,9 @@ class OneOf(Space[Any]):
         self.spaces = tuple(spaces)
         assert len(self.spaces) > 0, "Empty `OneOf` spaces are not supported."
         for space in self.spaces:
-            assert isinstance(space, Space), (
-                f"{space} does not inherit from `gymnasium.Space`. Actual Type: {type(space)}"
-            )
+            assert isinstance(
+                space, Space
+            ), f"{space} does not inherit from `gymnasium.Space`. Actual Type: {type(space)}"
         super().__init__(None, None, seed)
 
     @property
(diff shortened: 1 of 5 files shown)


pseudo-rnd-thoughts:
I'm starting to believe this is a sort of a Gymnasium 2.0 like feature change. 
This would give us more liberties for changing the API where necessary though I wish to make a few of that as possible. 
It would allow us to drop NumPy 1.x for 2.2+
Also we could simplify the Box space if we wished as currently it looks like a mound of technical debt to me.

Thoughts? If we doing this, it would be good to get some more voices involved to check their thoughts as well

amacati:
I would definitely be on board with converting everything. The spaces were just a good module to start with. I also agree with the technical dept. Box could be simplified by a lot if `low` and `high` were always arrays.

This could very well be something for a gymnasium 2.0 release and that it would be great to get some more voices from other people to see where the different interests lie at.

Jammf:
I agree that it'd make more sense as a 2.0 change, and a good opportunity to clean up the tech debt with `Box` with more dramatic changes. I also like the idea of having `low` and `high` be Arrays, constrained to having the same backing library, dtype, and device (and perhaps shape?). Then it'd be impossible to pass in values for `low`/`high` that are outside the bounds of the Box's `dtype`, so all that logic could be removed.

It's also make typing `Box` easier too, since then we could have something like:

```python
Box(Space[T]):
    def __init__(self, low: T, high: T, seed: SeedType): ...
    def contains(x: T) -> bool: ...
    def sample() -> T: ...
```

with `T` having the appropriate bounds. Then, if numpy (or other libraries) ever get something shape typing, that should just work without any changes on our end.

Jammf:
If all the implementation is removed from it, I think making `Env` a Protocol makes sense. That might also be a good time to add a `RenderableEnv` subprotocol to make rendering support optional, along the lines of what #842 was proposing.

amacati:
If we want this to be more of a 2.0 feature I could also draft a version with significantly reduced complexity which would not be backward compatible. But it does not make much sense to iterate more on this PR before we are sure what direction this should take

PR #1402, opened 2025-06-11
Install numpy during pre-commit workflow.

Jammf:
# Description

Some future pyright checks that would pass locally could fail in the GitHub Actions workflow, since the runner environment currently doesn't have access to type information from dependencies. It also leads to weaker type checking overall, since type information from dependencies is silently converted to `Any` when they aren't available.

This PR adds a step to the pre-commit workflow to install all of Gymnasium's dependencies, which is a simpler solution than requiring workarounds in code to make checks pass.

Also fixes a typo with the `python-version` key.

## Type of change

# Checklist:

This is a GitHub Actions workflow change, so local tests aren't applicable.

- ~[ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)~
- ~[ ] I have commented my code, particularly in hard-to-understand areas~
- ~[ ] I have made corresponding changes to the documentation~
- ~[ ] I have added tests that prove my fix is effective or that my feature works~
- ~[ ] New and existing unit tests pass locally with my changes~

### DIFF
--- a/.github/workflows/run-pre-commit.yml
+++ b/.github/workflows/run-pre-commit.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-      - run: pip install .[all]
+      - run: pip install numpy==1.26
       - run: pipx install pre-commit
       - run: pre-commit --version
       - run: pre-commit run --all-files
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts (commented):
Is `.[all]` necessary?

pseudo-rnd-thoughts (approved):

PR #1410, opened 2025-06-27
Fix "pygame is not responding" every time rendering the pygame window  in [ Taxi ] environment

RizkySugiharto:
# Description

I fix the bug that occur every time a program start rendering the pygame window using Taxi environment when I learning DQN.

## Type of change

## Demo Source Code to Test the Bug:
```py
env = gym.make("Taxi-v3", render_mode="human")

state, _ = env.reset()
terminated = False

while not terminated:
    next_state, reward, terminated, truncated, info = env.step(0)
```

## Demo Test Peview

*Sorry if the gif quality is too bad :(*

**Before Fix:** 
Before Fix

**After Fix:**
 After Fix

### DIFF
--- a/gymnasium/envs/toy_text/taxi.py
+++ b/gymnasium/envs/toy_text/taxi.py
@@ -574,6 +574,7 @@ class TaxiEnv(Env):
             )
 
         if mode == "human":
+            pygame.event.pump()
             pygame.display.update()
             self.clock.tick(self.metadata["render_fps"])
         elif mode == "rgb_array":


RizkySugiharto:
> Could you provide more information on how to test confirm the problem and solution?

I have added the demo test source code and the preview for the before and after fix

pseudo-rnd-thoughts (approved):
Thanks, I've checked the solution. Looking at the other environments, I think that most of already include this thankfully

PR #1411, opened 2025-06-28
Wrapper to discretize observations and actions

sparisi:
# Description

Wrappers to discretize continuous states and actions.
There are few environments with discrete states and actions, and it can be useful to benchmark algorithms on more of them -- especially to validate results on tabular algorithms before moving to approximate algorithsm (e.g., Q-Learning before DQN).
For example, MountainCar is a well-know hard-exploration problem, but its observation space is continuous. With the `DiscretizeObservation` wrapper, one can test tablar version of their algorithm on it as well.
The same can be done to environments with continuous actions with `DiscretizeAction`.

## Type of change

# Checklist:

### DIFF
--- a/gymnasium/wrappers/transform_action.py
+++ b/gymnasium/wrappers/transform_action.py
@@ -14,7 +14,7 @@ import numpy as np
 
 import gymnasium as gym
 from gymnasium.core import ActType, ObsType, WrapperActType
-from gymnasium.spaces import Box, Discrete, Space
+from gymnasium.spaces import Box, Discrete, MultiDiscrete, Space
 
 
 __all__ = ["TransformAction", "ClipAction", "RescaleAction"]
@@ -185,7 +185,7 @@ class DiscretizeAction(
     TransformAction[ObsType, WrapperActType, ActType],
     gym.utils.RecordConstructorArgs,
 ):
-    """Discretizes a continuous Box action space into a single Discrete space.
+    """Uniformly discretizes a continuous Box action space into a single Discrete space.
 
     Example 1 - Discretize Pendulum action space:
         >>> env = gym.make("Pendulum-v1")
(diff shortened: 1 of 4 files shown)


gymnasium/wrappers/transform_action.py
```diff
@@ -178,3 +179,107 @@ def __init__(
...
+        if not isinstance(env.action_space, Box):
+            raise TypeError(
+                "DiscretizeAction is only compatible with Box continuous actions."
+            )
+
+        self.low = env.action_space.low
+        self.high = env.action_space.high
+        self.n_dims = self.low.shape[0]
+
+        if not (np.all(np.isfinite(self.low)) and np.all(np.isfinite(self.high))):
```
pseudo-rnd-thoughts, line 245: Shouldn't this be np.any(...) or np.any(...) as we can't allow any low or high to be infinite?

gymnasium/wrappers/transform_action.py
```diff
@@ -178,3 +179,107 @@ def __init__(
...
+        gym.utils.RecordConstructorArgs.__init__(self, bins=bins)
+        gym.ActionWrapper.__init__(self, env)
+
+        if isinstance(bins, int):
+            self.bins = np.array([bins] * self.n_dims)
+        else:
+            assert len(bins) == self.n_dims, "bins must match action dimensions"
+            self.bins = np.array(bins)
+
+        self.bin_centers = [
```
pseudo-rnd-thoughts, line 260: Does this work for 2+ dim box or just 1d boxes?

pseudo-rnd-thoughts (changes requested):
I like the idea but I'm unsure of the implementations. Also could you add tests for it

gymnasium/wrappers/transform_action.py
sparisi: `not np.all(np.isfinite(...))` is equivalent to what you said. I'll switch to yours because it is clearer.

gymnasium/wrappers/transform_action.py
sparisi: Any dimension. I have added an example with LunarLander to the doc.

sparisi:
> I like the idea but I'm unsure of the implementations. Also could you add tests for it

Done. I added functions to revert the discretization (they return a range, since it is impossible to retrieve the exact original continuous obervation), and added tests to see that discretize + revert back stays within the bins range. 
I also checked that the spaces are discretized uniformly, and that errors are raised correctly.

I see that some already existing tests fail (from wrappers/vector) but on my PC everything passes.

Btw, I am on Windows and had to run `python -m pytest` to run the tests locally, [otherwise pytest could not import anything](https://stackoverflow.com/a/49488748/754136). This also prevented me to commit from my PC (had to use the browser). Not sure if others have ever reported this.

pseudo-rnd-thoughts:
> Btw, I am on Windows and had to run python -m pytest to run the tests locally, [otherwise pytest could not import anything](https://stackoverflow.com/a/49488748/754136). This also prevented me to commit from my PC (had to use the browser). Not sure if others have ever reported this.

This surprises me. Also why does windows prevent normal committing more files at the same time?

sparisi:
> > Btw, I am on Windows and had to run python -m pytest to run the tests locally, [otherwise pytest could not import anything](https://stackoverflow.com/a/49488748/754136). This also prevented me to commit from my PC (had to use the browser). Not sure if others have ever reported this.
> 
> This surprises me. Also why does windows prevent normal committing more files at the same time?

No idea. I use GitGui and this is the first time I am encountering this problem. I managed to fix everything in the end with commits from the browser.
Let me know if the PR is ok or if I need to do something else :)

gymnasium/wrappers/transform_action.py
```diff
@@ -178,3 +179,117 @@ def __init__(
...
+    def __init__(
+        self,
+        env: gym.Env[ObsType, ActType],
+        bins: int | tuple[int, ...],
+    ):
+        """Constructor for the discretize action wrapper.
+
+        Args:
+            env: The environment to wrap.
+            bins: int or tuple of ints (number of bins per dimension).
```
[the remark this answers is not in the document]
sparisi: "the bins used to discretrize the space" I would read this as if passing `[0, 5, 10]` means to discretize a continuous space [0, 10]. But `bins` is the number of bins per dimension, so I would have to pass `bins=2` to split the space in half.

gymnasium/wrappers/transform_action.py
```diff
@@ -178,3 +179,117 @@ def __init__(
...
+        self.bin_centers = [
+            0.5
+            * (
+                np.linspace(self.low[i], self.high[i], self.bins[i] + 1)[:-1]
+                + np.linspace(self.low[i], self.high[i], self.bins[i] + 1)[1:]
+            )
+            for i in range(self.n_dims)
+        ]
+
+        self.action_space = Discrete(np.prod(self.bins))
```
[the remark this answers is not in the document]
sparisi: Yes, because of that. Is there a wrapper to flatten MultiDiscrete into Discrete? I can make the discretization wrapper more general with MultiDiscrete, and then one can further flatten it to Discrete. If such a wrapper doesn't exist, I can make it.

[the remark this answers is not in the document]
sparisi: Sure, I think it should be easy by adding an optional flag. I'll look into it!

pseudo-rnd-thoughts (approved):
Looks good, thanks for the PR @sparisi

PR #1419, opened 2025-07-18
[Bugfix] fixed `step()` function in `AsyncVectorEnv` from hanging forever

### LINKED ISSUE #1417: [Bug]`AsyncVectorEnv.step_await` is blocked infinitely if `actions.shape[0] < num_envs`
### Describe the bug

In `AsyncVectorEnv`, the `step()` and `step_async()` methods do not check that the number of actions provided matches the number of environments. If the input actions array has fewer elements than `num_envs`, only that many pipes receive the `step` command. The remaining pipes receive nothing, causing `step_wait()` to block infinitely while waiting for responses from all pipes.

### Code example

```shell
import gymnasium as gym

def make_env():
    return gym.make("CartPole-v1")

env = gym.vector.AsyncVectorEnv([make_env for _ in range(4)])
obs = env.reset()
actions = env.action_space.sample()[:2]  # Intentionally too few actions
env.step(actions)  # This will hang forever
```

### System info

gymnasium version: 1.1.1
Python version: 3.10.17
OS: macOS 14.5 (darwin 24.5.0)

### Additional context

_No response_

### Checklist

matinmoezzi:
# Description

Added action space assertion in `AsyncVectorEnv` `step` method to ensure that the actions passed to the step method are validated against the action space, preventing `step_await` hangs forever.

Fixes #1417 

## Type of change

### Screenshots

image

# Checklist:

### DIFF
--- a/gymnasium/vector/sync_vector_env.py
+++ b/gymnasium/vector/sync_vector_env.py
@@ -243,25 +243,11 @@ class SyncVectorEnv(VectorEnv):
 
         Returns:
             The batched environment step results
-
-        Raises:
-            ValueError: If the number of actions does not match the number of environments.
         """
-        try:
-            actions = np.asarray(actions)
-            if actions.shape[0] != self.num_envs:
-                raise ValueError(
-                    f"Expected {self.num_envs} actions, got {actions.shape[0]}"
-                )
-        except Exception as e:
-            raise ValueError(
-                f"Actions must be convertible to a numpy array with shape ({self.num_envs}, ...), got {type(actions)} with error: {e}"
-            ) from e
-
         actions = iterate(self.action_space, actions)
 
         infos = {}
-        for i, action in enumerate(actions):
+        for i, (action, _) in enumerate(zip(actions, self.envs, strict=True)):
             if self.autoreset_mode == AutoresetMode.NEXT_STEP:
                 if self._autoreset_envs[i]:
                     self._env_obs[i], env_info = self.envs[i].reset()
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts (changes requested):
While this is the simplest solution, this could cause significant slowdowns. 
An alternative solution would be in `step_async` to confirm that every environment have an action. 
Also, could you add a check for the implementation with both Async and Sync vector environments

pseudo-rnd-thoughts (changes requested):
As we support 3.10+ then we can use `zip(..., strict=True)` which will raise an error if the two sequences are different lengths

gymnasium/vector/sync_vector_env.py
```diff
@@ -243,7 +243,21 @@ def step(
 
         Returns:
             The batched environment step results
+
+        Raises:
+            ValueError: If the number of actions does not match the number of environments.
         """
+        try:
```
pseudo-rnd-thoughts, line 250: As we support non array actions, e.g., a dictionary of actions then we can't use this code. 

We should implement the same code as `async_vector_env`, then `for i, (action, _) in enumerate(zip(actions, self.envs, strict=True)):`

Then could you add a test in `tests/vector/test_vector_env.py` with both `SyncVectorEnv` and `AsyncVectorEnv`

pseudo-rnd-thoughts (approved):
Tests look great, thanks for adding them @matinmoezzi

PR #1450, opened 2025-09-15
Standardize gym reward return

### LINKED ISSUE #1445: [Bug Report] Autoreset in AsyncVectorEnv causes crash in 1.2.0

gymnasium==1.2.0

amarkoolk:
When autoreset happens during AsyncVectorEnv step, the pipe returns a non numpy type. This causes a ValueError where the requested array has an inhomogeneous shape, when some of the environments return numpy array types and the environment index that autoreset returns python primitives.

Fixes #1445

I have modified the original PR to now standardize the gym returns to standard python types, rather than modify the asynchronous step return to some arbitrary sized numpy array.

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -43,6 +43,28 @@ from gymnasium.vector.vector_env import ArrayType, AutoresetMode, VectorEnv
 __all__ = ["AsyncVectorEnv", "AsyncState"]
 
 
+def _as_scalar_float(x) -> float:
+    a = np.asarray(x)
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):
@amarkoolk Sorry I'm only now reviewing your PR

However, I think the test is wrong as the reward return is a numpy array rather than a singular value. 
I modified the test to check with numpy values of the correct shape and didn't get an error.

Therefore, I suspect that your environment is returning the wrong type. 
Have you tested
```python
from gymnasium.utils.env_checker import check_env

check_env(your_env)
```

PR #1506, opened 2025-12-29
Add `params` argument to `FuncEnv.__init__` and update JAX wrappers

### LINKED ISSUE #905: [Proposal] Add `params` argument to `FuncEnv`
### Proposal

Currently, the `__init__` function does not take the `params` argument, instead it takes the `options` argument

https://github.com/Farama-Foundation/Gymnasium/blob/f88a61852b50fed3c907e2104019f453b9044c3a/gymnasium/functional.py#L49-L52

But the `params` of the environment are needed to initialize the object (such as to set `observation_space` member variable)

additional changes requirements:
`get_default_params` should be changed to not require the instantiation of the environment to be used as it will be required in `__init__`
https://github.com/Farama-Foundation/Gymnasium/blob/de909da30a1fe40288275da331fa7ad8e219a2bc/gymnasium/functional.py#L123-L125
to
```py
 def get_default_params(**kwargs) -> Params | None: 
     """Get the default params.""" 
     return None
```
and update `FunctionalJaxEnv(gym.Env)` with that change

@RedTachyon 

### Motivation

I have been developing the [MJX based environments](https://github.com/Farama-Foundation/Gymnasium/issues/833) and this is a requirement

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

Here are 2 examples of `params` I need for MJX
1) `xml_file` indicates the file of the robot model to be used and sets up the internal MuJoCo model and everything else that depends on it like action and observation spaces

2) `exclude_current_positions_from_observation` changes, how much of the state space is observable, it is need in `__init__()` to `observation_space` and later in `observation()`

### Checklist

Kallinteris-Andreas:
## Summary

- allow `FuncEnv` callers to supply an explicit params instance
- forward shared params through scalar and vector JAX adapters
- broadcast vector params with `in_axes=None` instead of vmapping them
- preserve scalar `state_info` for nested callbacks such as MJX's `transition_info`
- forward params through rendering initialization, rendering, and cleanup
- add regression tests for parameter storage, JAX adapter forwarding, and nested state-info calls

## Why

Some functional environments need parameters while constructing spaces and
other environment constants. The existing API generated defaults internally,
so callers could not provide the parameter instance used by the environment.

MJX's `transition_info` calls `state_info` for a single state. Keeping the
batched state-info callback on the vector adapter preserves that scalar call
while allowing `reset()` to return info for every environment.

## Validation

- `pytest -q tests/functional tests/envs/functional`
  - 21 passed, 1 skipped
- `ruff check --no-cache gymnasium/envs/functional_jax_env.py tests/envs/functional/test_functional_jax_env.py`
  - passed
- MJX Ant's batched state info and all seven transition-info fields matched scalar results on CPU

Fixes #905


Kallinteris-Andreas:
@pseudo-rnd-thoughts The PR is review ready

PR #1552, opened 2026-03-31
Fix TransformObservation __init__()

solismortis:
Obviously, it was intended for `__init__()` here to not require observation_space to be provided, so I replaced the pipe symbol with "=", as the pipe before None means "or no type", not None as the default.

It broke my old code.


pseudo-rnd-thoughts:
This code was intentionally like this to ensure that users are explicit with the observation space of the transformed observation as it might differ from the wrapped environment's observation space. 
If there is no change to the observation's shape, range, etc, users can easily specify `observation_space=None`

Issue #1510, opened 2026-01-05
[Bug Report] Rainy Taxi is inconsistent in left and right


dm-ackerman:
### Describe the bug

The rainy taxi has an 80% change to move as intended, a 10% chance to go left of course, 10% chance to go right. The left/right should be relative to the direction of motion. From the original paper, "probability 0.1 it instead moves to the right **(of the intended direction)** and with probability 0.1 it moves to the left".

They are not consistent in the code, which reverses them for some directions. More importantly, the test for valid movements seems to assume movements are to the left and right based on a north heading, regardless of actual motion. This blocks some moves that should be allowed.

Example, in the image below with the taxi at (1,3), all movements are valid. However the left and right are not consistent
<img width="326" height="214" alt="Image" src="https://github.com/user-attachments/assets/3ff4ad4e-bb68-4b19-a081-0077d60fb8f6" />

The transitions are:
```
If moving SOUTH, outcomes are
   With probability 0.8 move ahead to (2, 3) 
   With probability 0.1 move left  to (1, 2)  <--- actually right
   With probability 0.1 move right to (1, 4)   <--- actually left
If moving NORTH, outcomes are
   With probability 0.8 move ahead to (0, 3) 
   With probability 0.1 move left  to (1, 2) 
   With probability 0.1 move right to (1, 4) 
If moving EAST, outcomes are
   With probability 0.8 move ahead to (1, 4) 
   With probability 0.1 move left  to (2, 3)  <--- actually right 
   With probability 0.1 move right to (0, 3)   <--- actually left
If moving WEST, outcomes are
   With probability 0.8 move ahead to (1, 2) 
   With probability 0.1 move left  to (2, 3) 
   With probability 0.1 move right to (0, 3) 
```

If the taxi is one step to the east (location 1,4) , the invalid moves should be: EAST (and left/right of east), NORTH right, SOUTH left.  All other moves should be valid
<img width="322" height="214" alt="Image" src="https://github.com/user-attachments/assets/3cf5d8a2-c4a2-4b09-85c2-0a3a9ab1edea" />

The actual transitions are:
```
If moving SOUTH, outcomes are
   With probability 0.8 move ahead to (2, 4) 
   With probability 0.1 move left  to (1, 3) 
   With probability 0.1 move right to (1, 4) (no movement - move blocked)  <-- wrong: left/right swapped
If moving NORTH, outcomes are
   With probability 0.8 move ahead to (0, 4) 
[...]

Issue #1509, opened 2026-01-05
[Bug Report] Rainy Taxi behaves differently in different directions


dm-ackerman:
### Describe the bug

The rainy taxi has an 80% change to move as intended, a 10% chance to go left of course, 10% chance to go right.

When going east or west, the left/right options are only calculated if the east/west move is possible. If there is a wall in the way of the east/west move, it will never go left or right (even if those spots are open)

However, when going north or south, the left/right options are always calculated even if the north/south move is not possible. If there is a wall in the way of the north/south move, it will go left or right with 10% chance (if those spots are open).

These should be consistent. I don't know which behavior is "correct" and looking at the original paper, It's not clear there. I'm leaning towards the east/west behavior: if it can't move to the desired spot, no movement of any kind happens.

Example:
Consider this case, where the taxi is in the blue spot, it cannot move south or west:

<img width="316" height="213" alt="Image" src="https://github.com/user-attachments/assets/7c224579-33b1-45dc-b17e-c5f38c435b52" />

However, a south move gives a chance of moving to a neighboring spot where as a west move does not.
The transition options from the code are:
```
If moving SOUTH, outcomes are
   With probability 0.8 move ahead to (4, 3) (no change)
   With probability 0.1 move left  to (4, 3) (no change)
   With probability 0.1 move right to (4, 4)   <----  Can move even though main move is forbidden
If moving NORTH, outcomes are
   With probability 0.8 move ahead to (3, 3) 
   With probability 0.1 move left  to (4, 3) (no change)
   With probability 0.1 move right to (4, 4) 
If moving EAST, outcomes are
   With probability 0.8 move ahead to (4, 4) 
   With probability 0.1 move left  to (4, 3) (no change)
   With probability 0.1 move right to (4, 3) (no change)
If moving WEST, outcomes are
   With probability 0.8 move ahead to (4, 3) (no change)
   With probability 0.1 move left  to (4, 3) (no change) 
   With probability 0.1 move right to (4, 3) (no change)  <----  Can NOT move because main move is forbidden
```

### Code example

```shell
from gymnasium.envs.toy_text.taxi import TaxiEnv

env = TaxiEnv(is_rainy=True)

taxi_loc = (4, 3)  # blue spot
pax_loc = 0
pax_dest = 1
state = env.encode(taxi_loc[0], taxi_loc[1], pax_loc, pax_dest)

[...]

PR #1561, opened 2026-04-22
Update Taxi to V4 and fix `is_rainy` implementation

pseudo-rnd-thoughts:
# Description

https://github.com/Farama-Foundation/Gymnasium/issues/161 added `is_rainy` and `fickle_passegner` to fully reproduce the original paper. This added stochasticism to the environment through a probability of an action to move laterally (`is_rainy`) and for the passenger destination to change once the passenger is picked up. 
@dm-ackerman noticed that this implementation had two problems: https://github.com/Farama-Foundation/Gymnasium/issues/1509 and https://github.com/Farama-Foundation/Gymnasium/issues/1510
1. It was inconsistent behaviour when moving east or west compared to north or south
2. It was inconsistent behaviour in deciding the direction, sometimes using north or the direction of travel

Therefore, this PR looks to fix both problems, only affecting users using `is_rainy` or `fickle_passenger` while updating the version. 

In addition, this PR adds `rainy_probability` and `fickle_probability` for users to modify the probability that rainy or fickle events occur.


pseudo-rnd-thoughts:
I've confirmed the implementation through
```python
import gymnasium as gym
from gymnasium.utils.play import play

env = gym.make("Taxi-v4", render_mode="rgb_array", is_rainy=False, fickle_passenger=True, fickle_probability=1.0)
play(env, keys_to_action={"s": 0, "w": 1, "d": 2, "a": 3, "j": 4, "k": 5}, wait_on_player=True)
```

wissamblue69-dotcom:
> # Description
> 
> https://github.com/Farama-Foundation/Gymnasium/issues/161 added `is_rainy` and `fickle_passegner` to fully reproduce the original paper. This added stochasticism to the environment through a probability of an action to move laterally (`is_rainy`) and for the passenger destination to change once the passenger is picked up. 
> @dm-ackerman noticed that this implementation had two problems: https://github.com/Farama-Foundation/Gymnasium/issues/1509 and https://github.com/Farama-Foundation/Gymnasium/issues/1510
> 1. It was inconsistent behaviour when moving east or west compared to north or south
> 2. It was inconsistent behaviour in deciding the direction, sometimes using north or the direction of travel
> 
> Therefore, this PR looks to fix both problems, only affecting users using `is_rainy` or `fickle_passenger` while updating the version. 
> 
> In addition, this PR adds `rainy_probability` and `fickle_probability` for users to modify the probability that rainy or fickle events occur. 
>

PR #1567, opened 2026-04-27
Annotate the attributes of the `error` classes

jorenham:
# Description

Same story as #1566, but for `gymnasium.error`.

## Type of change

Static typing improvements

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1566, opened 2026-04-27
Fill in the missing type annotations in `core`

jorenham:
# Description

On the typestats dashboard ([url](https://jorenham.github.io/typestats/dashboard/report/#gymnasium)) I saw that there were a couple of missing annotations in `gymnasium.core`, so I figured I'd fill them in.

Fixes # (issue)

## Type of change

Static typing improvements

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1568, opened 2026-04-27
Return type annotations for the `logger` functions

jorenham:
# Description

The `gymnasium.logger` functions were missing return annotations. And even though you might think that type-checkers would then assume it returns `None`, they actually tend to infer `Any` in cases like these. So it helps to to be explicit about these things and spell out the `-> None`.

Related to #1566 and #1567.

## Type of change

Static typing improvements

# Checklist:


pseudo-rnd-thoughts (approved):
LGTM

PR #1573, opened 2026-04-29
Fix typing errors and add missing annotations in `vector.**`

jorenham:
# Description

Here's yet another batch of static typing improvements and fixes, this time for `gymnasium.vector.*` and `gymnasium.vector.utils.*`.

## Type of change

Static typing fixes and improvements.

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1575, opened 2026-04-29
Fix typing errors and add missing annotations in `wrappers.vector.*`

jorenham:
# Description

A batch of static typing fixes and type coverage improvements targeting the `gymnasium.wrappers.vector.*`  submodules. 

This also several cases of incorrect `TypeVar` usage patterns: unbound use (i.e. as "free" type variables), incorrect variance, and missing upper type parameter bounds. These issues aren't limited to this subpackage, and I wasn't able to fix all cases here because of that. I'll hope to be able to circle back to these though once these issues in the dependent classes (e.g. base classes, wrapped env classes, etc.) have been resolved. I've used PEP 696 type parameter defaults (or a backport thereof via `typing_extensions`) to keep things backwards compatible (type-checkers might complain about missing type arguments otherwise),

I also corrected a minor runtime bug in `vectorize_obsercation.TransformObservation`, where the wrong `_single_observation_space` attribute was written to (spurious underscore). I realize it's a bit out-of-scope, but I figured it was small enough to also fix here. I wouldn't mind moving this to a separate PR though if that what's preferred.

This fixes 28 ty warnings in total.

## Type of change

Static typing fixes and improvements.

# Checklist:

### DIFF
--- a/gymnasium/wrappers/vector/common.py
+++ b/gymnasium/wrappers/vector/common.py
@@ -151,7 +151,7 @@ class RecordEpisodeStatistics(VectorWrapper):
     ) -> tuple[
         np.ndarray,
         npt.NDArray[np.float64],
-        npt.NDArray[np.int_],
+        npt.NDArray[np.bool_],
         npt.NDArray[np.bool_],
         dict[str, Any],
     ]:
--- a/gymnasium/wrappers/vector/common.py
+++ b/gymnasium/wrappers/vector/common.py
@@ -4,13 +4,13 @@ from __future__ import annotations
 
 import time
 from collections import deque
+from typing import Any
 
 import numpy as np
+import numpy.typing as npt
 
-from gymnasium.core import ActType, ObsType
 from gymnasium.logger import warn
 from gymnasium.vector.vector_env import (
-    ArrayType,
     AutoresetMode,
     VectorEnv,
     VectorWrapper,
@@ -64,12 +64,24 @@ class RecordEpisodeStatistics(VectorWrapper):
                      't': array([0.007812, 0.      , 0.      ], dtype=float32)},
     """
 
+    episode_count: int
+
+    # 1-d arrays
+    episode_start_times: np.ndarray[tuple[int], np.dtype[np.float64]]
+    episode_returns: np.ndarray[tuple[int], np.dtype[np.float64]]
+    episode_lengths: np.ndarray[tuple[int], np.dtype[np.int_]]
+    prev_dones: np.ndarray[tuple[int], np.dtype[np.bool_]]
+
+    time_queue: deque[np.float64]
+    return_queue: deque[np.float64]
+    length_queue: deque[np.int_]
+
     def __init__(
         self,
         env: VectorEnv,
         buffer_length: int = 100,
         stats_key: str = "episode",
-    ):
+    ) -> None:
         """This wrapper will keep track of cumulative rewards and episode lengths.
 
         Args:
@@ -90,10 +102,10 @@ class RecordEpisodeStatistics(VectorWrapper):
 
         self.episode_count = 0
 
-        self.episode_start_times: np.ndarray = np.zeros((self.num_envs,))
-        self.episode_returns: np.ndarray = np.zeros((self.num_envs,))
-        self.episode_lengths: np.ndarray = np.zeros((self.num_envs,), dtype=int)
-        self.prev_dones: np.ndarray = np.zeros((self.num_envs,), dtype=bool)
+        self.episode_start_times = np.zeros((self.num_envs,))
+        self.episode_returns = np.zeros((self.num_envs,))
+        self.episode_lengths = np.zeros((self.num_envs,), dtype=int)
+        self.prev_dones = np.zeros((self.num_envs,), dtype=bool)
 
         self.time_queue = deque(maxlen=buffer_length)
         self.return_queue = deque(maxlen=buffer_length)
@@ -102,8 +114,8 @@ class RecordEpisodeStatistics(VectorWrapper):
     def reset(
         self,
         seed: int | list[int] | None = None,
-        options: dict | None = None,
-    ):
+        options: dict[str, Any] | None = None,
+    ) -> tuple[np.ndarray, dict[str, Any]]:
         """Resets the environment using kwargs and resets the episode returns and lengths."""
         obs, info = super().reset(seed=seed, options=options)
 
@@ -135,8 +147,14 @@ class RecordEpisodeStatistics(VectorWrapper):
         return obs, info
 
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, dict]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
+        npt.NDArray[np.float64],
+        npt.NDArray[np.int_],
+        npt.NDArray[np.bool_],
+        dict[str, Any],
+    ]:
         """Steps through the environment, recording the episode statistics."""
         (
             observations,
@@ -180,7 +198,6 @@ class RecordEpisodeStatistics(VectorWrapper):
                 infos[f"_{self._stats_key}"] = dones
 
             self.episode_count += num_dones
-
             for i in np.where(dones):
                 self.time_queue.extend(episode_time_length[i])
                 self.return_queue.extend(self.episode_returns[i])
--- a/gymnasium/wrappers/vector/dict_info_to_list.py
+++ b/gymnasium/wrappers/vector/dict_info_to_list.py
@@ -81,7 +81,7 @@ class DictInfoToList(VectorWrapper):
         np.ndarray,
         npt.NDArray[np.float64],
         npt.NDArray[np.bool_],
-        npt.NDArray[np.bool],
+        npt.NDArray[np.bool_],
         list[dict[str, Any]],
     ]:  # ty:ignore[invalid-method-override]
         """Steps through the environment, convert dict info to list."""
--- a/gymnasium/wrappers/vector/dict_info_to_list.py
+++ b/gymnasium/wrappers/vector/dict_info_to_list.py
@@ -5,9 +5,9 @@ from __future__ import annotations
 from typing import Any
 
 import numpy as np
+import numpy.typing as npt
 
-from gymnasium.core import ActType, ObsType
-from gymnasium.vector.vector_env import ArrayType, VectorEnv, VectorWrapper
+from gymnasium.vector.vector_env import VectorEnv, VectorWrapper
 
 __all__ = ["DictInfoToList"]
 
@@ -66,7 +66,7 @@ class DictInfoToList(VectorWrapper):
      * v1.0.0 - Renamed to ``DictInfoToList``
     """
 
-    def __init__(self, env: VectorEnv):
+    def __init__(self, env: VectorEnv) -> None:
         """This wrapper will convert the info into the list format.
 
         Args:
@@ -74,9 +74,16 @@ class DictInfoToList(VectorWrapper):
         """
         super().__init__(env)
 
+    # ty reports an error because the last return is a `dict` in super but a `list` here
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, list[dict[str, Any]]]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
+        npt.NDArray[np.float64],
+        npt.NDArray[np.bool_],
+        npt.NDArray[np.bool],
+        list[dict[str, Any]],
+    ]:  # ty:ignore[invalid-method-override]
         """Steps through the environment, convert dict info to list."""
         observation, reward, terminated, truncated, infos = self.env.step(actions)
         assert isinstance(infos, dict)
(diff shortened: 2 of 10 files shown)


gymnasium/wrappers/vector/common.py
```diff
@@ -102,8 +114,8 @@ def __init__(
     def reset(
         self,
         seed: int | list[int] | None = None,
-        options: dict | None = None,
-    ):
+        options: dict[str, Any] | None = None,
+    ) -> tuple[np.ndarray, dict[str, Any]]:
```
pseudo-rnd-thoughts, line 118: Should this be `tuple[ObsType, dict[str, Any]]`?

gymnasium/wrappers/vector/common.py
```diff
@@ -135,8 +147,14 @@ def reset(
         return obs, info
 
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, dict]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
+        npt.NDArray[np.float64],
+        npt.NDArray[np.int_],
```
pseudo-rnd-thoughts: This should be `npt.NDArray[np.bool_]`

gymnasium/wrappers/vector/dict_info_to_list.py
```diff
@@ -66,38 +66,45 @@ class DictInfoToList(VectorWrapper):
...
+    # ty reports an error because the last return is a `dict` in super but a `list` here
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, list[dict[str, Any]]]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
+        npt.NDArray[np.float64],
+        npt.NDArray[np.bool_],
+        npt.NDArray[np.bool],
```
pseudo-rnd-thoughts: Why is one `np.bool` and the other `np.bool_`? Shouldn't both be `np.bool_`?

gymnasium/wrappers/vector/dict_info_to_list.py
```diff
@@ -66,38 +66,45 @@ class DictInfoToList(VectorWrapper):
...
 
+    # ty reports an error because the last return is a `dict` in super but a `list` here
     def reset(
-        self,
-        *,
-        seed: int | list[int] | None = None,
-        options: dict[str, Any] | None = None,
-    ) -> tuple[ObsType, list[dict[str, Any]]]:
+        self, *, seed: int | None = None, options: dict[str, Any] | None = None
+    ) -> tuple[np.ndarray, list[dict[str, Any]]]:  # ty:ignore[invalid-method-override]
```
pseudo-rnd-thoughts, line 97: Same question about obs type

gymnasium/wrappers/vector/stateful_observation.py
```diff
@@ -87,7 +92,8 @@ def __init__(self, env: VectorEnv, epsilon: float = 1e-8):
             dtype=np.float32,
         )
         self.single_observation_space = new_single_space
-        self.observation_space = batch_space(new_single_space, self.num_envs)
+        # ty doesn't support `@single_dispatch` (yet?)
```
pseudo-rnd-thoughts: Add `todo` for the future?

gymnasium/wrappers/vector/stateful_reward.py
```diff
@@ -102,15 +108,21 @@ def reset(
         *,
         seed: int | list[int] | None = None,
         options: dict[str, Any] | None = None,
-    ) -> tuple[ObsType, dict[str, Any]]:
+    ) -> tuple[np.ndarray, dict[str, Any]]:
```
pseudo-rnd-thoughts, line 111: Shouldn't this be ObsType?

gymnasium/wrappers/vector/stateful_reward.py
```diff
@@ -102,15 +108,21 @@ def reset(
...
         self.accumulated_reward[:] = 0
         self._prev_dones[:] = 0
         return super().reset(seed=seed, options=options)
 
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, dict[str, Any]]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
```
pseudo-rnd-thoughts, line 120: ObsType?

gymnasium/wrappers/vector/vectorize_action.py
```diff
@@ -153,9 +177,10 @@ def __init__(
         self.action_space = batch_space(self.single_action_space, self.num_envs)
 
         self.same_out = self.action_space == self.env.action_space
-        self.out = create_empty_array(self.env.single_action_space, self.num_envs)
+        # ty doesn't support `@single_dispatch` yet
+        self.out = create_empty_array(self.env.single_action_space, self.num_envs)  # ty:ignore[invalid-assignment]
 
-    def actions(self, actions: ActType) -> ActType:
+    def actions(self, actions: np.ndarray) -> np.ndarray:
```
pseudo-rnd-thoughts, line 183: I think this should be ActType as any action can be transformed by the iterate and concatenate functions later

gymnasium/wrappers/vector/vectorize_observation.py
```diff
@@ -79,7 +95,7 @@ def __init__(
         else:
             self.observation_space = observation_space
             if single_observation_space is not None:
-                self._single_observation_space = single_observation_space
+                self.single_observation_space = single_observation_space
```
pseudo-rnd-thoughts, line 98: I'm surprised this didn't cause an error somewhere in our testing

gymnasium/wrappers/vector/vectorize_observation.py
```diff
@@ -176,11 +201,18 @@ def __init__(
...
-        self.out = create_empty_array(self.single_observation_space, self.num_envs)
+        # ty doesn't support `@single_dispatch` yet
+        self.out = create_empty_array(self.single_observation_space, self.num_envs)  # ty:ignore[invalid-assignment]
 
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, dict[str, Any]]:
+        self, actions: np.ndarray
+    ) -> tuple[
+        np.ndarray,
```
pseudo-rnd-thoughts, line 210: ObsType?

gymnasium/wrappers/vector/vectorize_observation.py
```diff
@@ -176,11 +201,18 @@ def __init__(
...
 
         self.same_out = self.observation_space == self.env.observation_space
-        self.out = create_empty_array(self.single_observation_space, self.num_envs)
+        # ty doesn't support `@single_dispatch` yet
+        self.out = create_empty_array(self.single_observation_space, self.num_envs)  # ty:ignore[invalid-assignment]
 
     def step(
-        self, actions: ActType
-    ) -> tuple[ObsType, ArrayType, ArrayType, ArrayType, dict[str, Any]]:
+        self, actions: np.ndarray
```
pseudo-rnd-thoughts, line 208: ActType?

gymnasium/wrappers/vector/vectorize_observation.py
```diff
@@ -196,10 +228,10 @@ def step(
 
         return obs, rewards, terminations, truncations, infos
 
-    def observations(self, observations: ObsType) -> ObsType:
+    def observations(self, observations: np.ndarray) -> np.ndarray:
```
pseudo-rnd-thoughts, line 231: ObsType?

pseudo-rnd-thoughts (commented):
Got a couple questions about when we can use ObsType and not. This is the weakness and most critical area of our type hinting that we've never really got to grips so your progress has been incredibly helpful

gymnasium/wrappers/vector/common.py
jorenham: `ObsType` is an (invariant) `TypeVar`. This `RecordEpisodeStatistics` isn't a generic class, and `RecordEpisodeStatistics.reset` isn't a generic function. If we were to use `ObsType` here, then it would be a free type variable. So there'd be no way for type-checkers to know what the type is that `ObsType` is supposed to represent.

The runtime analogue would be to a variable that's not defined, which would raise something like a `NameError` or an `UnboundLocalError`.

That's why I replaced these free type variables with actual types. And as far as I could tell, in this case a `np.ndarray` would be most appropriate, but correct me if I'm wrong.

If you prefer, we could also consider turning `RecordEpisodeStatistics` into a generic class using (covariant) type parameters for its observation type.
But if the observation type is always the same (for example because it's a `ndarray` like I've been assuming, and it has a fixed dtype such as `float64`), then there's no reason to parameterize it; it would unnecessarily complicate things without additional benefit.

gymnasium/wrappers/vector/dict_info_to_list.py
jorenham: Ah typo, my bad; good catch.

FYI, since `numpy>=2`, `numpy.bool_` is a direct alias of `numpy.bool`, and its only purpose it for backwards compatibility with `numpy<2`. So once gymnasium drops support for NumPy 1, then we could turn all `np.bool_` into `np.bool`.

gymnasium/wrappers/vector/stateful_observation.py
jorenham: I believe ty complains if there are any unused `# ty:ignore` comments. So if `ty` will implement support `@single_dispatch`, then we'll know right away. 

But either way adding a `TODO` couldn't hurt I suppose

gymnasium/wrappers/vector/vectorize_action.py
jorenham: Using `ActType `for both input and output here means that the input type must always be the same as the return type, even in any of the `TransformAction` subclasses. 

Maybe that's fine. But in case this method will e.g. upcast `np.int64` dtypes to `np.float64`, then it would be incorrect. Because in that case, the signature would be `NDArray[np.int64] -> NDArray[np.float64]`, which wouldn't "fit" the `T -> T` signature that it was. So that's why I played it safe here, and went with the uncorrelated (gradual) `np.ndarray` as input and output types.

But I haven't looked that closely at the implementation to be able to know for certain that it does, so maybe `T -> T` is fine :shrug:

gymnasium/wrappers/vector/vectorize_observation.py
jorenham: Yea same here; it looks kinda important :sweat_smile:

gymnasium/wrappers/vector/common.py
pseudo-rnd-thoughts: Thanks that makes sense. 
`RecordEpisodeStatistics` doesn't modify the observation or action at all, therefore, with `env: Env[ObsType, ActType]` then I thought we might be able to reduce `RecordEpisodeStatistics` to `[ObsType, ActType]`? 
I say this as a guess with limited understanding behind type checkers

jorenham: We could, but we'd have to make both `vector.vector_end.VectorWrapper` and this `RecordEpisodeStatistics` generic types. But I'm a bit hesitant to do that in this PR, because it would overlap with #1573. So although I'm all for it, maybe we should leave that for a follow-up (after this PR and #1573 have been merged)?

pseudo-rnd-thoughts: we've never really got the bottom of typing for environments and wrappers. Env has two and Wrapper has four.
https://github.com/Farama-Foundation/Gymnasium/blob/f417d2e6e18f4a54b93b6eb17f3b6fd9964e45c9/gymnasium/core.py#L25

https://github.com/Farama-Foundation/Gymnasium/blob/f417d2e6e18f4a54b93b6eb17f3b6fd9964e45c9/gymnasium/core.py#L290-L293

This "should" allow us to specify an environment's obs / action types and that the wrapper does or doesn't modify the obs / action types. 

Looking at the vector environment and wrapper, only `VectorEnv` is a generic while `VectorWrapper` doesn't have any.

https://github.com/Farama-Foundation/Gymnasium/blob/f417d2e6e18f4a54b93b6eb17f3b6fd9964e45c9/gymnasium/vector/vector_env.py#L40

https://github.com/Farama-Foundation/Gymnasium/blob/f417d2e6e18f4a54b93b6eb17f3b6fd9964e45c9/gymnasium/vector/vector_env.py#L352

So updating `VectorWrapper` is possibility necessary.

jorenham: > So updating `VectorWrapper` is possibility necessary.

Exactly. So I propose we do that after merging this and #1573 in a separate PR, and then either keep these to-be generic return types as-is or change them to `Any` for the time being. Is that ok with you?

pseudo-rnd-thoughts: Would it be possible to make the PR for `VectorWrapper` generic first then fix the types in `vector.**` and `wrappers.vector.**`? Otherwise wouldn't we be setting the types "incorrect" here and in #1573 then fixing them later?

jorenham: Yea that could work too I suppose. I'm just a bit worried that turning `VectorWrapper` generic will require us to make changes in a lot of other places in the code as well, because it'll likely result in new `ty` errors to be reported if we do that. So that's why I was thinking of doing that all at once in a follow-up, so that we can avoid a bunch of merge conflicts with the other open typing PRs.

Because keep in mind that the current situation is even more incorrect than if we were to change this to `Any`, so this is already an improvement in that sense.

pseudo-rnd-thoughts: Ok thats fair, if you think this is the better way of doing it then lets merge it and take your approach

jorenham: Ok I'll change these ndarray returns to `Any` then real quick

pseudo-rnd-thoughts (approved):
LGTM

gymnasium/wrappers/vector/common.py
jorenham: Haha ok nevermind; this is fine too

PR #1577, opened 2026-05-07
Generic vector env and vector wrapper types

jorenham:
# Description

Ref: https://github.com/Farama-Foundation/Gymnasium/pull/1575#discussion_r3180280831

This fixes the `TypeVar` issues that were discussed in #1575 by turning `VectorEnv` and all subclasses into generic types (with optional type parameters for backwards compatibility). 

I've declared the `TypeVar`s for each module separately. This is because it theoretically makes no sense for type parameters to "live outside" of the generic types they're associated with. They're not much different from function parameters: if you define `def spam(a: int)` you also don't `import a` first. 
And once PEP 695 syntax can be used (requiring Python 3.12+, or 3.13+ to use the type parameter defaults from PEP 696), then `TypeVar` will no longer be needed anyway.

## Type of change

Static typing fixes and improvements.

# Checklist:


pseudo-rnd-thoughts (commented):
Overall looks great, my primary worry is the variable naming. 
As Gymnasium is primarily intended as an API then a user needs to be able to implement their own environments / wrappers or use the built in ones and reading the type hints as `env: VectorEnv[_ObsT, _ActT_contra, _RewardArrT_co, _BoolArrT_co]` I know would scare me without significant understanding of type hinting. 
Therefore, could we update the variables to more easily understandable names? Like `ObsType`, `WrapperObsType`, etc? 
Also, do the variable naming need to be private variables? 

You know that it makes no sense to `from gymnasium.core import ObsType` which I partially agree with but I think about it more like an Interface that I promise to use this variable like this. That way its helpful as an API not just as an code description

Most of the typings require `if TYPE_CHECKING`, could we just use `from typing_extensions import TypeVar` everywhere? Claude suggests we would need to bump typing extensions to 4.4.0

jorenham:
>  As Gymnasium is primarily intended as an API then a user needs to be able to implement their own environments / wrappers or use the built in ones and reading the type hints as `env: VectorEnv[_ObsT, _ActT_contra, _RewardArrT_co, _BoolArrT_co]` I know would scare me without significant understanding of type hinting. Therefore, could we update the variables to more easily understandable names? Like `ObsType`, `WrapperObsType`, etc? Also, do the variable naming need to be private variables?
> 
> You know that it makes no sense to `from gymnasium.core import ObsType` which I partially agree with but I think about it more like an Interface that I promise to use this variable like this. That way its helpful as an API not just as an code description

I can imagine that for those that don't know much about typing this can indeed *look* confusing, but there are several reasons for using this naming scheme:

- Type-theoretically speaking, it makes no sense to treat the type parameters of a generic type of function as a separate entities; they're intrinsically part of the types/functions themselves, and have no meaning outside of that. It's not a very strong argument though, since Python's type system ignored core type-theoretical principles anyhow, but personally I think that's a mistake, and would like to see Python moving more *towards* the mathematically rigorous pillar called type-theory instead away from it.
- Type variables should not be part of the public API, and therefore should start with an `_`. As I explained in the main PR description; that's because they will not be around forever: they'll become obsolute one support for Python <3.12 is dropped. 
- `ObsType` and `ActType` are invariant, which for most use-cases (including the current `Env` for example) is not what you want, and can lead to unexpected typing errors. 
- The current `*Type` suffix implies that they represent *types* rather than the *instances* that they actually **actually** represent. Consider for example the `types` stdlib ([docs](https://docs.python.org/3/library/types.html)), where you have `NoneType` describing `type(None)`, `LambdaType` representing `type(lambda _: _)`, `EllipsisType` representing `type(...)`, etc. etc. So following that widely used convention, it natural to expect that `ObsType` represents `type(observations)`, even though in reality it represents `observations` - the instance, not its type. 
- Instead, for type parameters, the convention is to use `*T` as suffix. Or, in case of covariant or contravariant type parameters, that'll be `*T_co` and `*T_contra`, respectively. TBH, I'm personally not a big fan of this convention (`_co` and `_contra` are way too "jargon-y" IMO), but it is the way it is I'm afraid, and is even enforced by ruff through its stable [`PLC0105`](https://docs.astral.sh/ruff/rules/type-name-incorrect-variance/) rule (docs). Many related libraries also use this convention, including NumPy. That's why I expect that deviating from this will be *more* confusing to users.

> Most of the typings require `if TYPE_CHECKING`, could we just use `from typing_extensions import TypeVar` everywhere? Claude suggests we would need to bump typing extensions to 4.4.0

Ah I didn't realize that you have `typing_extensions` as runtime dependency. That indeed will make things a lot easier. I'll get rid of the nasty `TYPE_CHECKING` stuff. Good catch!

jkterry1:
@jorenham we're starting to review this, would you main fixing the merge conflicts and merging main into your branch?

jorenham:
> @jorenham we're starting to review this, would you main fixing the merge conflicts and merging main into your branch?

Great! I'll rebase tomorrow

jkterry1:
@jorenham could you please rebase off main?

jorenham:
Huh what happened here? There are a bunch of unrelated commits and conflicts now :sweat_smile:

pseudo-rnd-thoughts:
Apologies @jorenham for not returning to this PR earlier. I would love to get this PR merged and hopefully cut a Gymnasium v1.4 soon. 

Looking at the PR again, I still have a problem with the new type hints added as we seem to have two type hinting philosphies of importing centralised type hints in `gymnasium/core.py` and then the individual typevars that this PR adds for `gymnasium/vector`. 
Exploring what to do, I got claude to see what keeping the old import type hints API work look like and produced https://github.com/pseudo-rnd-thoughts/Gymnasium/pull/3 (I've reviewed all the code and made some changes)
What are your thoughts on my changes? I added a `gymnasium/typing.py` to centralise the type hints.

jorenham:
Sorry I seemed to have missed your earlier comment. I'll have a look tomorrow.

jorenham:
> Looking at the PR again, I still have a problem with the new type hints added as we seem to have two type hinting philosphies of importing centralised type hints in `gymnasium/core.py` and then the individual typevars that this PR adds for `gymnasium/vector`.

I wrote a review comment at https://github.com/pseudo-rnd-thoughts/Gymnasium/pull/3#discussion_r3723948675 that's relevant to this

> What are your thoughts on my changes? I added a `gymnasium/typing.py` to centralise the type hints.

I'm having a bit of trouble seeing what actually changed, because it seems to build on this PR, so I can't really see which parts are mine and which are yours.

pseudo-rnd-thoughts:
@jorenham Do you have anymore thoughts on https://github.com/pseudo-rnd-thoughts/Gymnasium/pull/3#discussion_r3740317645, my plan is to merge this PR as is because the trade off between the type hints for now is not great but better for the current version than 7e58a2e5a55fbbf17e4b72fa5348c3fc9346e922

jorenham:
> @jorenham Do you have anymore thoughts on [pseudo-rnd-thoughts#3 (comment)](https://github.com/pseudo-rnd-thoughts/Gymnasium/pull/3#discussion_r3740317645), my plan is to merge this PR as is because the trade off between the type hints for now is not great but better for the current version than [7e58a2e](https://github.com/Farama-Foundation/Gymnasium/commit/7e58a2e5a55fbbf17e4b72fa5348c3fc9346e922)

I finally got around to it (sorry I've been a bit busy recently). I'll rebase once we've come to a decision on how to move on with this.

PR #1586, opened 2026-05-27
perf: use pygame.display.init() instead of pygame.init() in all environments

### LINKED ISSUE #1585: [Proposal] Faster PyGame Init
### Proposal

Per the comment in https://github.com/Farama-Foundation/PettingZoo/issues/1252

If I'm correct that the suggestion is correct, and we aren't doing this in Gymnasium already, then we probably want to do it for performance reasons.

### Motivation

_No response_

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

midhunxavier:
## Description

Replace `pygame.init()` with `pygame.display.init()` (and `pygame.font.init()` where text rendering is used) across all environment and wrapper files that render via pygame.

`pygame.init()` initialises every subsystem — including audio (mixer), joystick, CDROM, etc. — none of which RL environments need. Selective initialisation is:
- **Faster** at startup (the audio-device enumeration in `pygame.init()` can take several seconds on Windows/macOS).
- **Avoids spurious failures/warnings** in headless CI / containerised environments where the audio device is absent.
- Consistent with the recommendation in Farama-Foundation/PettingZoo#1252.

Files changed: CartPole (+ vector), ContinuousMountainCar, MountainCar, Pendulum, Acrobot, LunarLander, BipedalWalker, CarRacing, Blackjack (toy_text + tabular), CliffWalking (toy_text + tabular), FrozenLake, Taxi, phys2d variants, and the scalar + vector `HumanRendering` wrappers.

Adds `test_pygame_selective_init`, a regression test that asserts `pygame.display` is initialised while `pygame.mixer` is **not** after an `rgb_array` render.

Verified in a headless container: every patched env renders `rgb_array` without a display, and `pygame.mixer.get_init()` is `None` after rendering (it was `(44100, -16, 2)` with the old `pygame.init()`).

Fixes #1585

---
*This PR was produced with the assistance of an AI coding assistant.*

## Checklist

### DIFF
--- a/gymnasium/envs/box2d/bipedal_walker.py
+++ b/gymnasium/envs/box2d/bipedal_walker.py
@@ -631,7 +631,6 @@ class BipedalWalker(gym.Env, EzPickle):
             ) from e
 
         if self.screen is None and self.render_mode == "human":
-            pygame.init()
             pygame.display.init()
             self.screen = pygame.display.set_mode((VIEWPORT_W, VIEWPORT_H))
         if self.clock is None:
--- a/gymnasium/envs/box2d/car_racing.py
+++ b/gymnasium/envs/box2d/car_racing.py
@@ -602,7 +602,6 @@ class CarRacing(gym.Env, EzPickle):
 
         pygame.font.init()
         if self.screen is None and mode == "human":
-            pygame.init()
             pygame.display.init()
             self.screen = pygame.display.set_mode((WINDOW_W, WINDOW_H))
         if self.clock is None:
--- a/gymnasium/envs/box2d/lunar_lander.py
+++ b/gymnasium/envs/box2d/lunar_lander.py
@@ -683,7 +683,6 @@ class LunarLander(gym.Env, EzPickle):
             ) from e
 
         if self.screen is None and self.render_mode == "human":
-            pygame.init()
             pygame.display.init()
             self.screen = pygame.display.set_mode((VIEWPORT_W, VIEWPORT_H))
         if self.clock is None:
--- a/gymnasium/envs/classic_control/acrobot.py
+++ b/gymnasium/envs/classic_control/acrobot.py
@@ -297,9 +297,8 @@ class AcrobotEnv(Env):
             ) from e
 
         if self.screen is None:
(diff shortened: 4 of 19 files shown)


pseudo-rnd-thoughts (commented):
Thanks for the PR, have you verified that it works on normal computer? Also, do you know the startup time difference?

midhunxavier:
@pseudo-rnd-thoughts Yes — verified on a real machine (macOS, pygame 2.6 / SDL 2.28), both `human` mode (real window opens) and `rgb_array` render correctly, and the audio mixer is never initialised (`pygame.mixer.get_init()` is `None`, vs `(44100, -16, 2)` with the old `pygame.init()`).

**Startup time difference** (measured natively on macOS, *not* in Docker — Docker has no real audio device, so SDL falls back to a null audio driver and the benefit would be hidden):

Init-call cost (20 cold-process samples, median):

| init call | time |
|---|---|
| `pygame.init()` (old) | 176.7 ms |
| `pygame.display.init()` + `pygame.font.init()` (new, text envs) | 79.8 ms |
| `pygame.display.init()` (new, non-text envs) | 82.7 ms |

→ **~95 ms saved (~2.2× faster)**; the difference is almost entirely the audio mixer opening the audio device.

End-to-end cold `import gymnasium → make(rgb_array) → reset → render()` (two identical clean venvs, only the gymnasium version differs; 12 samples, median):

| env | old | new | saved | speedup |
|---|---|---|---|---|
| CartPole-v1 | 275 ms | 184 ms | 91 ms | 1.49× |
| Blackjack-v1 | 283 ms | 191 ms | 91 ms | 1.48× |
| Acrobot-v1 | 281 ms | 180 ms | 101 ms | 1.56× |

The ~91 ms end-to-end saving matches the ~95 ms init-call delta. The win scales with how slow audio-device enumeration is on the platform, so it's larger on Windows (the original report, "several seconds") than the ~95 ms measured here on macOS.

pseudo-rnd-thoughts (approved):
The changes look good and the test at the end is very helpful

Issue #1610, opened 2026-07-02
Doc Site Gifs Showing Real Policies for Box2D and MuJoCo


jkterry1:
It'd be really nice if people could submit PRs with gifs showing reasonable behaviors rolling out real policies (just something basic from SB3 or something) for the Classic Control, Box2D and MuJoCo environments. Right now they just show random behaviors. Separate PRs per environment or set of environment probably makes the most sense.

The toy text environments we should do this for, but at time of writing a designer is working on overhauling the pixel art for them, so that one should probably wait.

aminehd:
Hi, I would really like to work on this. I can start with simplest env (Classic Control) and pick anything from SB3. 

I was wondering would it be possible for me to create a smaller PR just for Classic Control and assign it to myself so duplicate pr for same env won't happen?

jkterry1:
Just treat this comment/similar ones as self assignment for a subitem. Thanks for taking a crack at this :)

-------- Original Message --------

aminehd:
I created a tool that generates gifs for envs + algos (https://github.com/aminehd/gymbench/blob/master/README.md) and generated the gifs. Would really appreciate a review

https://github.com/Farama-Foundation/Gymnasium/pull/1612

aminehd:
Hi @jkterry1 , 
I have created a pr for two envs in box2d Lunar lander and bipedal walker(pr:  https://github.com/Farama-Foundation/Gymnasium/pull/1627). They look really cool to me as i can see what they learnt:)
I haven't figured what policy gonna work well for Car race yet so I left it out. Based on Claude Code I should use a CNN policy. Will send a pr for that

aminehd:
Hi @jkterry1 , I rendered the carRace too. I used CNNPolicy with 4 Milion steps. however the car goes out of track at the end

natsu0529:
Hi, I'd like to work on the trained-policy GIF for HalfCheetah-v5
as a small MuJoCo PR.

I'll use a deterministic Stable-Baselines3 policy and match the existing
docs/_scripts/gen_gifs.py format.

aminehd:
Added a trained Ant gif . Used SAC policy with 1 Milion steps. Verified the animation that ant will move out of border.

PR #1612, opened 2026-07-04
Use trained-policy GIFs for classic control environments

aminehd:
# Description

Part of #1610. Swaps the random-agent GIFs for the five Classic Control envs with renders of simple trained Stable-Baselines3 policies, so the docs show the envs being solved instead of flailing.

Only the 5 `.gif` files change. Rendering follows `docs/_scripts/gen_gifs.py` (300 frames, `duration=50`, `.unwrapped`) — the one difference is the action comes from `model.predict()` instead of `action_space.sample()`.

Fixes # (part of #1610)

## Type of change

## Reproduce

Scripts + configs live in a small side repo
`gymlab/` → [`packages/gymbench`](https://github.com/aminehd/gymbench) (generator) + `external/Gymnasium` (this fork)

```bash
git clone --recurse-submodules https://github.com/aminehd/gymlab
cd gymlab && pip install -e packages/gymbench
gymbench env-gif run          # trains + renders all 5 from the config
```

-  core logic that generates the gif 
https://github.com/aminehd/gymbench/blob/7c882faf85ce49ddd32cc586268cb0d134faef2e/src/gymbench/core/render.py#L29-L35
- config: https://github.com/aminehd/gymbench/blob/master/configs/env_gif/classic_control.toml
- how it works: https://github.com/aminehd/gymbench/blob/master/README.md

# Checklist


jkterry1:
-Can we get an acrobot gift where the policy actually learns to stand up?
-The cart pole gif is almost less instructive about the environment dymanics than the old one
-Mountain car continuous and pendulum don't appear to have new gifs based on the github diff?

aminehd:
Thanks @jkterry1  for the review.
I looked into the three cases you mentioned. I will update the commit in this pr shortly

- The acrobot goal is to get above the black line. In the Gif it gets there and the next frame gets rendered too fast. I added some waiting to end of gif so the last moment stays for some frame. I attached the success frame for acrobot here to confirm that is what it supposed to learn.  I think acrobot cant stay stationary at top for ever
- The cart pole gif converge too fast, I change the render (begin the pole at a larger tilt , same learnt policy) so that learning would be more visible. Let me know if you have specific setting in mind.
-  For me both car continuous and pendulum new gifs are different from old for me. namely car continuous old dont get to target but new does, and pendulum in old dont get stationary but new does. i will post a screen shot. Although I am not sure if in case of pendulum it is desired to stay at top. 
- Though seems to me the pendulum should settle at top.  its reward -(θ² + 0.1·θ̇² + 0.001·torque²) is maximised upright and stll. 

Acrobat stop:
image

Car continuous and pendulum 
image

aminehd:
Updated the commit

aminehd:
Thanks so much for the review again.
Quick question. Would you prefer one PR per environment (the issue mentioned per-env or small groups), or is this single classic-control PR fine as is?

I haven't dived deep into each env at first time and was more into getting result with a given policy.
I can spend some time playing around with that for each env in classical control as well.

I am pretty new to open source contribution and it would be very nice for me to get my first commit in

jkterry1:
Hey,

Thank you. I'm largely happy with the gifs :)
-Is the current cart pole with with the change you suggested to make the progression more obvious, and if not can you do that?
-I'm a bit confused about pendulum, but does the current gif include the changes you were discussing?

aminehd:
Hey, 
Glad to here that :)
To answer questions , yes to both
  - CartPole in the latest commit pole begins at a larger tilt so you can see the policy actively catch and recover it.
  - Pendulum in the latest commit is the trained policy that swings up and settles upright at the top (the reward-optimal state).

jkterry1:
This works, want to do another set of environents? :)

aminehd:
Thanks @jkterry1 . Yeah, I'd love to work on rest of them. I'll work on Box2D envs next

PR #1467, opened 2025-10-19
Add dtype parameter to Discrete

VadimBim:
# Description

This is a continuation of the work started in #1197. During implementation, the following questions arose:

1. **MultiDiscrete dtype initialization**: In [`gymnasium/spaces/multi_discrete.py` (line 64)](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/spaces/multi_discrete.py#L64), `MultiDiscrete` never triggers `dtype = None` unless explicitly provided. By default, it uses `np.int64`. Is this the intended behavior? Should there be an assertion or warning when `dtype` is not explicitly provided?

2. **`_batch_differing_space_discrete` behavior**: In [`gymnasium/vector/utils/space_utils.py` (line 202)](https://github.com/VadimBim/Gymnasium/blob/issue-1118/gymnasium/vector/utils/space_utils.py#L202), now this function infers the `dtype` from the first space in the list. Should we add a warning when spaces in the list have different integer dtypes and casting to `np.int64` is performed? Or should we raise an assertion if `Discrete` spaces with different dtypes are not allowed as arguments to this function?

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1616, opened 2026-07-08
Fix integer overflow in Discrete.contains for small dtypes (int8/uint8)

teddytennant:
## What

`Discrete.contains` computes the exclusive upper bound of the valid range as `self.start + self.n`. Because `start` and `n` are stored as scalars of the space's own dtype, this addition is performed in that (possibly narrow) dtype. When the largest element of the space sits at the dtype's maximum, `start + n` overflows and wraps around, so the range check `as_np < self.start + self.n` becomes always-false and `contains` rejects every valid element.

This change performs the range comparison in arbitrary-precision Python ints (`int(self.start) <= int(as_np) < int(self.start) + int(self.n)`) so it can no longer overflow the space dtype. The dtype-castability guard (`np.can_cast(as_np.dtype, self.dtype)`) is unchanged.

## Why

The per-space `dtype` argument (added in #1467) makes it easy to construct spaces backed by small integer dtypes such as `np.int8` / `np.uint8`. For example:

```python
import numpy as np

space = Discrete(3, start=125, dtype=np.int8)  # elements {125, 126, 127}
space.contains(127)             # False -- but 127 is a valid element
space.contains(space.sample())  # False for every sample
```

Here `start + n == 128` overflows `int8` to `-128`, so the comparison `x < -128` is never satisfied and the space reports that it does not contain its own samples. The same happens for `Discrete(6, start=250, dtype=np.uint8)` (`250 + 6 == 256` wraps to `0`). Doing the comparison in Python ints fixes this without changing behavior for any non-overflowing case.

## Testing

Added `test_contains_small_dtype_no_overflow` to `tests/spaces/test_discrete.py`, parametrized over `int8`/`uint8` spaces whose largest element is at the dtype max. It asserts that `contains` accepts the boundary elements (as both Python `int` and numpy scalar) and that the space contains its own samples. These assertions fail before the change and pass after.

```
pytest tests/spaces/test_discrete.py -q
34 passed
```

The existing `test_contains` parametrization is unchanged and still passes.

### DIFF
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -210,7 +210,9 @@ class Discrete(Space[_IntegerT_co], Generic[_IntegerT_co]):
         else:
             return False
 
-        value_is_in = bool(self.start <= as_np < self.start + self.n)
+        value_is_in = bool(
+            int(self.start) <= int(as_np) < int(self.start) + int(self.n)
+        )
 
         return value_is_in and np.can_cast(as_np.dtype, self.dtype)
 
--- a/tests/spaces/test_discrete.py
+++ b/tests/spaces/test_discrete.py
@@ -166,6 +166,24 @@ def test_contains(element, dtype, expected_is_in):
 
 
 @pytest.mark.parametrize(
+    "n, start, dtype, element",
+    [
+        (3, 125, np.int8, 125),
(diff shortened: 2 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1621, opened 2026-07-11
Accept any SupportsFloat reward in passive env checker (#1341)

### LINKED ISSUE #1341: [Bug Report] Inconsistent definition of reward
### Describe the bug

In the step function, reward is defined as `SupportsFloat`. However, its type is required to be `np.floating` in
[passive_env_checker](https://github.com/Farama-Foundation/Gymnasium/blob/69471be60182d2046e073aa6723b36a7aa818232/gymnasium/utils/passive_env_checker.py#L243)

### Code example

```shell

```

### System info

_No response_

### Additional context

_No response_

### Checklist

Mohit-Ak:
## Description

`env_step_passive_checker` warns when a reward is returned that is neither an `np.integer` nor an `np.floating` dtype:

```python
if not (
    np.issubdtype(type(reward), np.integer)
    or np.issubdtype(type(reward), np.floating)
):
    logger.warn("The reward returned by `step()` must be a float, int, np.integer or np.floating, ...")
```

But `Env.step` types its reward as `SupportsFloat`, so any object with a `__float__` is a valid reward. The dtype-based check is stricter than the declared contract and produces a false-positive warning for several legitimate reward types:

| reward | `SupportsFloat` | warned before | warned after |
|---|---|---|---|
| `float`, `int` | ✅ | no | no |
| `np.float32`, `np.int64` | ✅ | no | no |
| `bool` | ✅ | **yes** | no |
| `Decimal` | ✅ | **yes** | no |
| custom `__float__` object | ✅ | **yes** | no |
| `str` | ❌ | yes | yes |

This is the case reported in #1341 (a `dataclass` with `__float__` that satisfies `SupportsFloat` but fails `np.issubdtype(type(reward), np.floating)`).

## Fix

Check `isinstance(reward, SupportsFloat)` — matching the type the API already declares — instead of the two `np.issubdtype` checks. A `str` (which is not `SupportsFloat`) is still rejected, so the guard's real purpose is preserved. The NaN/inf checks now cast via `float(reward)` first so they keep working for `Decimal`/custom rewards (previously `np.isnan` was only called after a `cast` that did nothing at runtime).

The `typing.cast` import became unused and was removed.

## How it was tested

Reproduced the false positive against the current code, then confirmed the fix silences it while `str` still warns. Added regression tests to `tests/utils/test_passive_env_checker.py`:

- `test_passive_env_step_checker_supportsfloat_reward` — asserts no reward warning for `float`, `int`, `bool`, `np.float32`, `np.int64`, `Decimal`, and a custom `__float__` object. On the unpatched code this fails for exactly `bool`/`Decimal`/`custom` (RED); passes with the fix (GREEN).
- `test_passive_env_step_checker_nan_reward_still_warns` — proves the NaN check still fires for `SupportsFloat` rewards (`float("nan")`, `Decimal("NaN")`).

```
$ python -m pytest tests/utils/test_passive_env_checker.py -q
55 passed
```

Ran `ruff check` and `ruff format --check` (v0.14.10, the pinned pre-commit version) on both changed files — clean. The two `test_env_checker.py::test_no_error_warnings` failures in my environment are a missing-`pygame` dependency and reproduce identically on `main`, unrelated to this change.

Fixes #1341


jkterry1:
@Mohit-Ak could you please rebase off main?

pseudo-rnd-thoughts:
Thanks for the PR, I agree that the type hint is `SupportsFloat` however I view this a technical debt and the wrong type hint that without a major version update can't be changed. Therefore, I think we're right to restrict the passive env checkers reward types to those expected. I'm happy to continue the conversation, just closing it for now

PR #1623, opened 2026-07-11
Fix vector envs mutating the sub-environment's class-level metadata when setting autoreset_mode

teddytennant:
## What

`SyncVectorEnv` and `AsyncVectorEnv` take a reference to the first sub-environment's `metadata` dict and then write `autoreset_mode` into it. For nearly all environments `metadata` is a class attribute, so this mutates the class-level dict shared by every instance of that environment class:

```python
import gymnasium as gym
from gymnasium.vector import SyncVectorEnv, AutoresetMode

envs = SyncVectorEnv(
    [lambda: gym.make("CartPole-v1")], autoreset_mode=AutoresetMode.SAME_STEP
)
env = gym.make("CartPole-v1")
env.metadata["autoreset_mode"]  # AutoresetMode.SAME_STEP -- leaked from the vector env
```

Beyond leaking a vector-only key into plain environments, two vector envs over the same environment class end up sharing one dict, so the second constructor's `autoreset_mode` silently overwrites the first's `metadata["autoreset_mode"]`.

This change copies the dict (`self.metadata = dict(...)`) before inserting `autoreset_mode`, in both vector env classes.

## Why

`metadata["autoreset_mode"]` is documented as the way to detect a vector env's autoreset behavior (and wrappers/training loops read it), so two vector envs with different modes reporting the same value is actively misleading. The class-level mutation also survives closing the vector env, affecting unrelated code in the same process.

## Testing

Added `test_metadata_sync_vector_env` / `test_metadata_async_vector_env`, which construct two vector envs over `CartPole-v1` with different autoreset modes and assert that each reports its own mode and that the sub-environment's metadata is untouched. Both fail before the change and pass after.

```
pytest tests/vector/test_sync_vector_env.py tests/vector/test_async_vector_env.py -q
55 passed
```

(The one remaining local failure, `test_sync_vector_env_seed`, needs Box2D which isn't installed here -- unrelated to this change.)

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -192,7 +192,8 @@ class AsyncVectorEnv(VectorEnv):
         dummy_env = env_fns[0]()
 
         # As we support `make_vec(spec)` then we can't include a `spec = dummy_env.spec` as this doesn't guarantee we can actual recreate the vector env.
-        self.metadata = dummy_env.metadata
+        # Copy the metadata such that we don't modify the sub-environment's (class-level) metadata
+        self.metadata = dict(dummy_env.metadata)
         self.metadata["autoreset_mode"] = self.autoreset_mode
         self.render_mode = dummy_env.render_mode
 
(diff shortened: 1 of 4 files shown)


jkterry1:

pseudo-rnd-thoughts (approved):
Thanks for PR, looks good to me

PR #1624, opened 2026-07-11
Fix vector RecordEpisodeStatistics reporting wrong stats with SAME_STEP autoreset

teddytennant:
# Description

`vector.RecordEpisodeStatistics` reports wrong returns and lengths when the vector env uses `AutoresetMode.SAME_STEP`. The accumulation logic assumes `NEXT_STEP` semantics: it uses `prev_dones` to skip the step after a termination/truncation (the reset step) and to clear the accumulators. Under `SAME_STEP`, the sub-environment resets within the terminating step itself, so the following step is a real first step of the new episode -- masking it out both drops that step's reward/length and clears the accumulators one step too late.

The effect is that only the first episode of each sub-environment is reported correctly and every subsequent episode is under-reported by one step. Reproducible with a fixed 3-step, reward-1-per-step episode:

```python
envs = gym.wrappers.vector.RecordEpisodeStatistics(
    SyncVectorEnv([make_env] * 2, autoreset_mode=AutoresetMode.SAME_STEP)
)
envs.reset(seed=123)
# first done:      infos["episode"]["r"] == [3. 3.], "l" == [3 3]   (correct)
# every one after: infos["episode"]["r"] == [2. 2.], "l" == [2 2]   (should be 3)
```

The wrapper already reads `env.autoreset_mode` in `reset` to decide when to clear statistics, but `step` ignored it. This change branches on the mode in `step`: for `SAME_STEP`, every step counts towards an episode (no `prev_dones` masking) and the accumulators for done sub-environments are cleared at the end of the same step, right after their statistics are recorded. `NEXT_STEP` and `DISABLED` behavior is unchanged.

## Type of change

## Testing

Added `test_episode_statistics_with_autoreset_mode` to `tests/wrappers/vector/test_record_episode_statistics.py`, parametrized over all three `AutoresetMode`s, using a fixed-length `GenericTestEnv` so the exact expected return/length is known. It runs 25 steps (several episodes per sub-environment) and asserts every reported `r`/`l` equals the true episode length, including everything accumulated in `return_queue`/`length_queue`. The `SAME_STEP` case fails before this change and passes after; `NEXT_STEP` and `DISABLED` pass before and after.

```
pytest tests/wrappers/vector/test_record_episode_statistics.py -q
5 passed
```

# Checklist:

### DIFF
--- a/gymnasium/wrappers/vector/common.py
+++ b/gymnasium/wrappers/vector/common.py
@@ -175,15 +175,24 @@ class RecordEpisodeStatistics(VectorWrapper):
             f"`vector.RecordEpisodeStatistics` requires `info` type to be `dict`, its actual type is {type(infos)}. This may be due to usage of other wrappers in the wrong order."
         )
 
-        self.episode_returns[self.prev_dones] = 0
-        self.episode_returns[np.logical_not(self.prev_dones)] += rewards[
-            np.logical_not(self.prev_dones)
-        ]
+        if self._autoreset_mode == AutoresetMode.SAME_STEP:
+            # Sub-environments reset within the same step as they terminate or
+            # truncate, therefore, every step counts towards an episode.
+            self.episode_returns += rewards
+            self.episode_lengths += 1
+        else:
+            # For `NEXT_STEP` autoreset, the step after a termination or
+            # truncation resets the sub-environment and doesn't count towards
+            # the next episode's statistics.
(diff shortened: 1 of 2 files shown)


jkterry1:

pseudo-rnd-thoughts (approved):
Thanks for the PR and spotting the problem

PR #1625, opened 2026-07-11
Fix create_shared_memory for dtypes without an array module typecode (float16)

teddytennant:
## What

`create_shared_memory` for `Box`/`Discrete`/`MultiDiscrete` spaces passes the numpy dtype's single-character code straight to `ctx.Array`. That only works for dtypes whose character happens to be a valid `array` module typecode. For dtypes without one — most notably `np.float16` (`'e'`) — `multiprocessing` raises `TypeError: this type has no size`, so `AsyncVectorEnv(..., shared_memory=True)` cannot be constructed at all for float16 observation spaces.

This change keeps the existing path for dtypes with a valid typecode and falls back to allocating the equivalent number of bytes (`c_uint8`, `size * itemsize`) otherwise. The fallback is transparent to the read/write side: `read_from_shared_memory` and `write_to_shared_memory` already reinterpret the raw buffer via `np.frombuffer(..., dtype=space.dtype)`, so only the allocation needed fixing.

## Why

float16 observation spaces are a natural choice for memory-heavy image/embedding observations, but vectorizing them with shared memory currently fails on construction:

```python
import numpy as np
from gymnasium.vector.utils import create_shared_memory

create_shared_memory(Box(-1, 1, (3,), dtype=np.float16), n=2)
# TypeError: this type has no size
```

The same error surfaces from `AsyncVectorEnv([...], shared_memory=True)` for any env whose (possibly nested) observation space contains a float16 leaf. The existing special case for bool (`'?'` → `c_bool`) is this same problem solved for one specific dtype; the fallback generalizes it to every dtype the `array` module doesn't cover (float16, complex dtypes, and float128 on platforms that have it).

## Testing

Added `test_shared_memory_no_array_typecode_dtypes` to `tests/vector/utils/test_shared_memory.py`, parametrized over float16 (and float128 where available), asserting that samples written to shared memory round-trip with the correct dtype, shape, and values. Also added `test_float16_async_vector_env_shared_memory` to `tests/vector/test_async_vector_env.py` exercising the end-to-end `AsyncVectorEnv` reset/step path with `shared_memory=True`. Both fail with the `TypeError` above before the change and pass after.

```
pytest tests/vector/utils/test_shared_memory.py tests/vector/test_async_vector_env.py -q
223 passed, 66 skipped
```

`pre-commit run --files gymnasium/vector/utils/shared_memory.py tests/vector/test_async_vector_env.py tests/vector/utils/test_shared_memory.py` passes.


jkterry1:

pseudo-rnd-thoughts:
Closing, please make a new PR with the changes based on main

PR #1626, opened 2026-07-11
Fix non-deterministic Text charset ordering under hash randomization

teddytennant:
# Description

`Text` builds its internal character list and index from raw set iteration order, and the default charset is a `frozenset` whose iteration order depends on `PYTHONHASHSEED`. So identically seeded `Text` spaces behave differently across processes: `sample()` returns different strings for the same seed, `flatten()`/`unflatten()` map characters to different integers, and positional entries in the sample/probability `mask` refer to different characters. Easy to see with the default charset:

```
$ PYTHONHASHSEED=1 python -c "from gymnasium.spaces import Text; print(Text(5, seed=42).sample())"
p
$ PYTHONHASHSEED=2 python -c "from gymnasium.spaces import Text; print(Text(5, seed=42).sample())"
s
```

This silently breaks seeded reproducibility for anything using `Text` across workers or reruns (hash randomization is on by default), and makes flattened `Text` observations inconsistent between processes.

The fix normalizes the ordering when the character list is built: set-based charsets are sorted, while string (or other ordered) charsets keep their given order with duplicate characters dropped, so `character_index` stays consistent with `character_list`. `charset` / `character_set` semantics are unchanged.

## Type of change

## Testing

Added two tests to `tests/spaces/test_text.py`: `test_charset_ordering` checks that set charsets come out sorted and string charsets keep their order with duplicates dropped, and `test_deterministic_across_hash_seeds` runs a seeded `sample()` and a `flatten()` in subprocesses under different `PYTHONHASHSEED` values and asserts identical output -- it fails before this change and passes after.

```
pytest tests/spaces -q
1083 passed
```

# Checklist:

### DIFF
--- a/gymnasium/spaces/text.py
+++ b/gymnasium/spaces/text.py
@@ -68,12 +68,20 @@ class Text(Space[str]):
         self.min_length: int = int(min_length)
         self.max_length: int = int(max_length)
 
-        self._char_set: frozenset[str] = frozenset(charset)
-        self._char_list: tuple[str, ...] = tuple(charset)
+        if isinstance(charset, (set, frozenset)):
+            # Set iteration order depends on hash randomization (PYTHONHASHSEED),
+            # sort to keep sampling, masks and flattening consistent across processes.
+            char_list = sorted(charset)
+        else:
+            # Preserve the given ordering, dropping duplicate characters
+            char_list = list(dict.fromkeys(charset))
+
+        self._char_set: frozenset[str] = frozenset(char_list)
+        self._char_list: tuple[str, ...] = tuple(char_list)
         self._char_index: dict[str, np.int32] = {
-            val: np.int32(i) for i, val in enumerate(tuple(charset))
+            val: np.int32(i) for i, val in enumerate(char_list)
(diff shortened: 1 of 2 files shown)


jkterry1:

gymnasium/spaces/text.py
```diff
@@ -68,12 +68,20 @@ def __init__(
         self.min_length: int = int(min_length)
         self.max_length: int = int(max_length)
 
-        self._char_set: frozenset[str] = frozenset(charset)
-        self._char_list: tuple[str, ...] = tuple(charset)
+        if isinstance(charset, (set, frozenset)):
```
Trenza1ore, line 71:
```suggestion
        if isinstance(charset, collections.abc.Set):
```

pseudo-rnd-thoughts: I think this is fine and `dict.keys()` aren't really support

pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts:
Thanks for the PR

PR #1628, opened 2026-07-11
Fix: Resolve floating-point type violation in Pygame window resizing, cumulative aspect-ratio scaling drift, KEYUP event crash and zoom validation bugs in play.py utility

Ankush0077:
### Summary
This PR resolves three distinct correctness, type-safety, and runtime crash bugs inside the interactive keyboard play utility (`gymnasium/utils/play.py`):
1. A type-safety and cumulative aspect-ratio scaling drift bug during Pygame window resizing.
2. A fatal `ValueError` crash risk and "sticky keys" event loop bug during keyboard input tracking.
3. An unvalidated `zoom <= 0` parameter causing raw C-level display crashes.

---

#### 1. Resizing: Type-Safety & Cumulative Scaling Drift
*   **The Bug:** During window resizing (`pygame.WINDOWRESIZED`), the scaling factor was calculated and applied iteratively against the *already scaled* `self.video_size` rather than the native resolution. This resulted in:
    *   **Type-Safety Violation:** `self.video_size` mutated into a tuple of floats (e.g., `(795.0, 447.9)`). Passing floats to `pygame.transform.scale()`, whose API expects integer `(width, height)` dimensions, violates the Pygame API contract. On modern Python C-API bindings (where implicit integer conversions via `__int__` have been removed in favor of `__index__`), this can actively raise a `TypeError`.
    *   **Cumulative Scaling Drift:** Simply applying `round()` to the existing logic introduces cumulative scaling drift. Because the scale factor is computed iteratively against the *already modified* `self.video_size`, rounding errors compound on every frame during a mouse drag, permanently warping the game's aspect ratio. (A 1,000-frame simulation of a shaky mouse drag on a `160x900` window proves that naively adding `round()` permanently warps the aspect ratio and shrinks the window height by 16px).
*   **The Fix:**
    *   Introduced an immutable `self.original_video_size` initialized in `__init__`.
    *   Refactored `WINDOWRESIZED` to compute scaling relative to `original_video_size` instead of the previous frame's size, completely eliminating compound drift.
    *   Applied `round()` to guarantee dimensions are passed to Pygame as type-safe integers without compounding any rounding errors.

#### 2. Keyboard Inputs: KEYUP Crash & Sticky Keys
*   **The Bug:** `self.pressed_keys` was initialized as a `list` and manipulated via `.append()` and `.remove()`. 
    *   **The Crash:** If the window loses focus during a keypress or if the OS dispatches duplicate key release events, `pygame.KEYUP` can trigger for a key that isn't currently tracked, causing `list.remove()` to raise a fatal `ValueError` and crash the entire play session.
    *   **Sticky Keys:** Multiple sequential `KEYDOWN` events from OS keyboard auto-repeat append duplicate keys to the list, but only a single `KEYUP` event is dispatched on release. This leaves duplicate keys in the list, making the key appear permanently active.
*   **The Fix:**
    *   Migrated `self.pressed_keys` from a `list` to a `set`.
    *   Used `self.pressed_keys.add()` on `KEYDOWN` (handling duplicates natively) and `self.pressed_keys.discard()` on `KEYUP` (safely removing the key if present, and silently doing nothing if missing).

#### 3. Initialization: Unvalidated Zoom Parameter
*   **The Bug:** The `play()` and `PlayableGame` APIs accepted a `zoom` argument without any bounds validation. If a user accidentally passed a non-positive float (e.g., `zoom=0` or `zoom=-1.5`), `PlayableGame._get_video_size` computed zero or negative dimensions and passed them directly to `pygame.display.set_mode()`. This bypassed Python's validation layers and triggered an unhandled C-level display crash inside Pygame's SDL backend (`pygame.error: Cannot set negative sized display mode` or similar).
*   **The Fix:** Added a robust validation check inside `PlayableGame.__init__` to fail fast at the API boundary with an informative `ValueError` if a non-positive `zoom` is provided. Because `play()` instantiates `PlayableGame` internally, this safely protects both public entry points with a single, clean check.

---

### Testing & Verification
*   Added four dedicated regression tests in `tests/utils/test_play.py`:
    *   `test_pygame_window_resize`: Verifies that the scaled `video_size` elements are strictly integers and asserts zero aspect-ratio drift over sequential resize iterations.
    *   `test_keyboard_keyup_unpressed_key_no_crash`: Verifies that receiving a `KEYUP` event for an untracked key does not raise a `ValueError`.
    *   `test_keyboard_duplicate_keydown_event`: Verifies that duplicate `KEYDOWN` events followed by a single `KEYUP` event correctly releases the key.
    *   `test_invalid_zoom_validation`: Verifies that negative and zero zoom inputs safely raise `ValueError` at the API boundary.
*   Updated existing tests in `test_play.py` to align with the new `set` container type.
*   All 3,546 tests in the suite pass successfully.


jkterry1:
@Ankush0077 could you please rebase off main?

pseudo-rnd-thoughts:
Closing due to the number of file changes, please recreate a new PR with main

Ankush0077:
created new pull request #1633

PR #1639, opened 2026-07-15
Fix CarRacing HUD score overflow from float formatting

### LINKED ISSUE #1638: [Bug Report]  CarRacing v3 HUD score overflows the frame due to float decimal precision
### Describe the bug

The reward/score drawn in the bottom-left HUD is a float formatted with `:04`, which sets a minimum width but no decimal precision. So instead of a 4-digit score it renders the float's full representation (e.g. 23.590909090909083), which overflows and is clipped off the left edge of the window.

See https://github.com/Farama-Foundation/Gymnasium/blob/17eaef2cbaedf05f951b5652bf021b272e4a5dd9/gymnasium/envs/box2d/car_racing.py#L640

```py
text = font.render(f"{self.reward:04}", True, (255, 255, 255), (0, 0, 0))
```

To fix it, add .0f so the width applies to a rounded integer:

```py
text = font.render(f"{self.reward:04.0f}", True, (255, 255, 255), (0, 0, 0))
```

23.59… → 0024

Here are the screenshots of current behaviour and after fix. Note that the after screenshot matches format shown on the demo gif on https://gymnasium.farama.org/environments/box2d/car_racing/

current (broken):

Image

fixed: 

Image

I believe this PR introduced the bug:

https://github.com/Farama-Foundation/Gymnasium/pull/1505/changes#diff-b2657ab36b5d46d0f112f0c7d20e3026e4d50c8a0e260e5669002d229e5f027c

and 1.2.3 doesn't have this issue.

### Code example

```shell

```

### System info

- gymnasium 1.3.0
- macOS
- python 3.10.17 and 3.14.0

### Additional context

_No response_

### Checklist

Fstarnb:
## Summary

The CarRacing HUD formatted `self.reward` with `f"{self.reward:04}"`, which only sets a minimum width and still prints full float decimals, overflowing the bottom-left score.

Use `f"{self.reward:04.0f}"` so the score renders as a fixed-width integer field (matching the demo gif).

## Test plan

- [x] Format sanity: `f"{23.59:04.0f}" == "0024"`
- [ ] Manual render with `render_mode="human"` if pygame available

Fixes #1638.

### DIFF
--- a/gymnasium/envs/box2d/car_racing.py
+++ b/gymnasium/envs/box2d/car_racing.py
@@ -637,7 +637,7 @@ class CarRacing(gym.Env, EzPickle):
         self._render_indicators(WINDOW_W, WINDOW_H)
 
         font = pygame.font.Font(pygame.font.get_default_font(), 42)
-        text = font.render(f"{self.reward:04}", True, (255, 255, 255), (0, 0, 0))
+        text = font.render(f"{self.reward:04.0f}", True, (255, 255, 255), (0, 0, 0))
         text_rect = text.get_rect()
         text_rect.center = (60, WINDOW_H - WINDOW_H * 2.5 / 40.0)
         self.surf.blit(text, text_rect)


pseudo-rnd-thoughts (approved):
Thanks for the PR fixing it, looks like the original "04i" would have converted to integer which this PR fixes

PR #1640, opened 2026-07-16
Add vector environment step benchmark

Kallinteris-Andreas:
## Summary

- add `benchmark_step_vector` for measuring individual vector-environment steps per second

## Why

I was using it in MJX env testing

## Validation

- two-environment CartPole smoke run
- `ruff check gymnasium/utils/performance.py`
- `ruff format --check gymnasium/utils/performance.py`


tests/utils/test_performance.py
Kallinteris-Andreas, line 1: this file is a GPTism, it is not incorrect, but i would personally not include it, though it could be expanded if we wanted to keep it

gymnasium/utils/performance.py
```diff
@@ -51,6 +54,53 @@ def benchmark_step(
     return steps_per_time
 
 
+def benchmark_step_vector(
```
pseudo-rnd-thoughts: I think this should be `benchmark_vector_step`

gymnasium/utils/performance.py
```diff
@@ -51,6 +54,53 @@ def benchmark_step(
...
+
+    Returns:
+        The number of individual environment steps per second.
+    """
+    env.action_space.seed(seed)
+    env.reset(seed=seed)
+
+    # Warm up lazy initialization, including JIT compilation, outside the benchmark.
+    env.step(env.action_space.sample())
+    env.reset(seed=seed)
```
pseudo-rnd-thoughts, line 77: This should the line before `while True:`

gymnasium/utils/performance.py
```diff
@@ -51,6 +54,53 @@ def benchmark_step(
...
+        if autoreset_mode == AutoresetMode.DISABLED:
+            done = np.logical_or(terminated, truncated)
+            if np.any(done):
+                env.reset(options={"reset_mask": done})
+
+        if time.time() - start > target_duration:
+            end = time.time()
+            break
+
+    return steps / (end - start)
```
pseudo-rnd-thoughts, line 103: Would the user want to know the number of steps, total time duration, and steps per second?

tests/utils/test_performance.py
pseudo-rnd-thoughts: Do we have a tests for `performance.py`? I agree its a gptism but it would be nice to have some basic performance testing in the CI, for when we do something stupid, then it gets caught. Might be better as a separate PR

gymnasium/utils/performance.py
Kallinteris-Andreas: This reset is intentionally before the warm-up. Some environments perform additional lazy initialization during their first `reset()` after `make()`, which can take substantially longer than subsequent resets. Keeping it here ensures that initialization overhead is excluded from the benchmark. I added a comment to clarify this.

gymnasium/utils/performance.py
Kallinteris-Andreas: the user passes `target_duration` which is the duration in seconds, and the other metrics can be easily derived

pseudo-rnd-thoughts (approved):
LGTM

PR #1642, opened 2026-07-17
Fix MujocoEnv shrinking XML-declared offscreen framebuffer

### LINKED ISSUE #1607: [Question] MujocoEnv ignores `offwidth` and `offheight` parameter from XML file?
### Question

I am currently creating a custom environment by extending the `MujocoEnv` base class. The environment contains a drone which has a front-facing camera attached to it. To access this camera I need to instantiate a custom renderer with the desired width and height of the camera image:
```python
class DroneEnv(MujocoEnv, utils.EzPickle):
    def __init__(self, xml_file, frame_skip, **kwargs):
        ...
        MujocoEnv.__init__(self, xml_file, frame_skip, observation_space=None, **kwargs)
        ...
        self.front_cam_renderer = mujoco.Renderer(self.model, width=960, height=540)
        ...
```
However, this approach raises an error because the image width and height is bigger than the width and height of the frame buffer. In the XML file of my environment I have set the parameters to create a bigger frame buffer:
```xml
<mujoco model="drone_env">
    ...
    <visual>
        <global offwidth="960" offheight="540"/>
    </visual>
    ...
</mujoco>
```
But the error still persists. After digging through the code base I found out that `MujocoEnv.__init__()` overwrites the off screen sizes of the buffer in [this method](https://github.com/Farama-Foundation/Gymnasium/blob/169a247ab478b9f7e36edb96d06574ae60abb03c/gymnasium/envs/mujoco/mujoco_env.py#L115) and sets it to the width and height parameters which come in as function arguments. This arguments simultaneously set the size of the rendering window, which I don't want to.

Is this the intended behavior? What is the best way to set the size of the frame buffer independently to the window size?

midhunxavier:
# Description

`MujocoEnv._initialize_simulation()` unconditionally set the offscreen framebuffer size from the render-window size:

```python
model.vis.global_.offwidth = self.width    # self.width defaults to 480
model.vis.global_.offheight = self.height
```

This silently clobbered a larger buffer that a user had already declared in their model XML via `<visual><global offwidth="960" offheight="540"/></visual>`. A subsequent higher-resolution `mujoco.Renderer(model, width=960, height=540)` then failed with `ValueError: Image width 960 > framebuffer width 480`.

This changes the assignment to **grow, never shrink**, so the larger of the XML-declared size and the requested render window wins in both directions:

```python
model.vis.global_.offwidth = max(model.vis.global_.offwidth, self.width)
model.vis.global_.offheight = max(model.vis.global_.offheight, self.height)
```

- An XML-declared buffer larger than the render window survives (fixes the reported bug).
- A render window larger than the XML declaration still enlarges the buffer, so on-screen rendering continues to fit.

**Note:** For an env whose XML declares no `<global>` element, the offscreen buffer now keeps MuJoCo's compiled default (640×480) instead of being force-shrunk to 480×480. This is harmless — the rendered viewport is governed independently by `width`/`height` (`Ant-v5` still renders `(480, 480, 3)`) — but it is a change from the previous forced-480 behavior, so flagging it for reviewers.

Fixes #1607

## Type of change

### Screenshots

Not applicable (no visual change). Reproduction, headless via `MUJOCO_GL=egl`, using a custom env whose XML declares `offwidth=960 offheight=540`:

| | `model.vis.global_.offwidth/offheight` after `__init__` | `mujoco.Renderer(model, 960, 540)` |
| ------ | ----- | ----- |
| Before | `480 / 480` | raises `ValueError: Image width 960 > framebuffer width 480` |
| After  | `960 / 540` | succeeds → image shape `(540, 960, 3)` |

# Checklist:

- [ ] I have made corresponding changes to the documentation — N/A (no docs reference `offwidth`/`offheight`)
- [x] New and existing unit tests pass locally with my changes (`tests/envs/mujoco/`: 261 passed, 6 skipped)

### DIFF
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -56,8 +56,11 @@ class MujocoEnv(gym.Env):
             frame_skip: Number of MuJoCo simulation steps per gym `step()`.
             observation_space: The observation space of the environment.
             render_mode: The `render_mode` used.
-            width: The width of the render window.
-            height: The height of the render window.
+            width: The width of the render window. The offscreen framebuffer is grown to
+                this size if the model XML declares a smaller one, but never shrunk below
+                the XML-declared size.
+            height: The height of the render window. Grown, but never shrunk, in the same
+                way as `width`.
             camera_id: The camera ID used.
             camera_name: The name of the camera used (can not be used in conjunction with `camera_id`).
             default_camera_config: configuration for rendering camera.
@@ -120,11 +123,7 @@ class MujocoEnv(gym.Env):
         """
         model = mujoco.MjModel.from_xml_path(self.fullpath)
         # MjrContext will copy model.vis.global_.off* to con.off*
-        # Only grow the offscreen framebuffer, never shrink it below the size
-        # the model's XML explicitly requested via `<visual><global offwidth=...
-        # offheight=.../></visual>`. This lets a user allocate a buffer larger
-        # than the render window/`width`/`height` (e.g. for a separate high-res
-        # `mujoco.Renderer`). See https://github.com/Farama-Foundation/Gymnasium/issues/1607
+        # Only grow the offscreen framebuffer, never shrink below the size of the model XML.
         model.vis.global_.offwidth = max(model.vis.global_.offwidth, self.width)
         model.vis.global_.offheight = max(model.vis.global_.offheight, self.height)
         data = mujoco.MjData(model)
--- a/gymnasium/envs/mujoco/mujoco_env.py
+++ b/gymnasium/envs/mujoco/mujoco_env.py
@@ -120,8 +120,13 @@ class MujocoEnv(gym.Env):
         """
         model = mujoco.MjModel.from_xml_path(self.fullpath)
         # MjrContext will copy model.vis.global_.off* to con.off*
-        model.vis.global_.offwidth = self.width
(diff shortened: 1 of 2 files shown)


gymnasium/envs/mujoco/mujoco_env.py
```diff
@@ -120,8 +120,13 @@ def _initialize_simulation(
         """
         model = mujoco.MjModel.from_xml_path(self.fullpath)
         # MjrContext will copy model.vis.global_.off* to con.off*
-        model.vis.global_.offwidth = self.width
-        model.vis.global_.offheight = self.height
+        # Only grow the offscreen framebuffer, never shrink it below the size
```
pseudo-rnd-thoughts: Reduce the comment to
```
# Only grow the offscreen framebuffer, never shrink below the size of the model XML.
```

pseudo-rnd-thoughts (changes requested):
Thanks for the PR, looks good. Could you add a comment to the `__init__(width, height)` about this

gymnasium/envs/mujoco/mujoco_env.py
midhunxavier: Done in 011e305f — reduced the comment to exactly that line, and added a note to the `width`/`height` args in the `__init__` docstring explaining that the offscreen framebuffer is grown to that size but never shrunk below the XML-declared one.

pseudo-rnd-thoughts (approved):
Thanks for the changes

PR #1643, opened 2026-07-18
Fix AsyncVectorEnv NEXT_STEP autoreset reward/done dtypes

### LINKED ISSUE #1445: [Bug Report] Autoreset in AsyncVectorEnv causes crash in 1.2.0

Fstarnb:
## Description

In `AsyncVectorEnv` with `AutoresetMode.NEXT_STEP`, workers that are autoresetting return a **scalar** zero reward while siblings may still return **ndarray** rewards (for example length-1 `float32` vectors from custom envs).

`step_wait` then did `np.array(rewards, dtype=np.float64)`, which raises:

```text
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape ...
```

This is the failure reported in #1445. (Standard CartPole scalar rewards already stack on main; they do **not** reproduce the issue.)

**Fix:** batch rewards through `_batch_rewards()`, which broadcasts autoreset scalar zeros onto the shared non-scalar reward shape before stacking.

## Related Issue

Fixes #1445

## How Has This Been Tested?

- `tests/vector/test_async_vector_env.py::test_async_vector_env_next_step_autoreset_array_reward` (shared_memory True/False)
  - **Fails on current main** with the inhomogeneous-shape `ValueError`
  - **Passes with this PR**
- `tests/vector/test_async_vector_env.py::test_async_vector_env_next_step_autoreset_scalar_reward` — CartPole scalar smoke test
- Local numpy 2.3.x / macOS arm64 / Python 3.13

## Checklist

- [x] I have commented my code where helpful
- [x] I have added tests that prove my fix is effective (red on main, green here)
- [x] New and existing unit tests pass locally with my changes (aside from pre-existing pygame-missing render/call tests in this environment)

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -806,7 +806,11 @@ def _async_worker(
                 if autoreset_mode == AutoresetMode.NEXT_STEP:
                     if autoreset:
                         observation, info = env.reset()
-                        reward, terminated, truncated = 0, False, False
+                        # Use numpy scalars so step_wait can stack rewards / done flags
+                        # when only some workers are in the autoreset path (issue #1445).
+                        reward = np.float64(0.0)
+                        terminated = np.bool_(False)
+                        truncated = np.bool_(False)
                     else:
                         (
                             observation,
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -51,6 +51,39 @@ _StepResult: TypeAlias = tuple[
 ]
 
 
+def _batch_rewards(rewards: list[Any]) -> npt.NDArray[np.float64]:
+    """Stack per-env rewards, broadcasting autoreset scalar zeros onto array rewards.
+
+    ``NEXT_STEP`` autoreset returns a scalar zero reward while other workers may
+    still return ndarray rewards (including length-1 vectors). A plain
+    ``np.array(rewards, dtype=np.float64)`` then fails with an inhomogeneous-shape
+    error (see issue #1445). When every non-scalar reward shares one shape, scalar
+    placeholders are expanded to zeros of that shape before stacking.
+    """
+    target_shape: tuple[int, ...] | None = None
+    for reward in rewards:
+        arr = np.asarray(reward)
+        if arr.ndim >= 1:
+            if target_shape is None:
+                target_shape = arr.shape
+            elif arr.shape != target_shape:
+                # Divergent vector rewards: fall back and let NumPy raise.
+                target_shape = None
+                break
+
+    if target_shape is not None:
+        normalized: list[np.ndarray] = []
+        for reward in rewards:
+            arr = np.asarray(reward, dtype=np.float64)
+            if arr.ndim == 0:
+                normalized.append(np.zeros(target_shape, dtype=np.float64))
+            else:
+                normalized.append(np.asarray(arr, dtype=np.float64))
+        return np.stack(normalized, axis=0)
+
+    return np.asarray(rewards, dtype=np.float64)
+
+
 class AsyncState(Enum):
     """The AsyncVectorEnv possible states given the different actions."""
 
@@ -514,7 +547,7 @@ class AsyncVectorEnv(VectorEnv):
         self._state = AsyncState.DEFAULT
         return (
             deepcopy(self.observations) if self.copy else self.observations,
-            np.array(rewards, dtype=np.float64),
+            _batch_rewards(rewards),
             np.array(terminations, dtype=np.bool_),
             np.array(truncations, dtype=np.bool_),
             infos,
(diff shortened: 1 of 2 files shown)


gymnasium/vector/async_vector_env.py
```diff
@@ -806,7 +806,11 @@ def _async_worker(
                 if autoreset_mode == AutoresetMode.NEXT_STEP:
                     if autoreset:
                         observation, info = env.reset()
-                        reward, terminated, truncated = 0, False, False
+                        # Use numpy scalars so step_wait can stack rewards / done flags
+                        # when only some workers are in the autoreset path (issue #1445).
+                        reward = np.float64(0.0)
```
Sanjays2402, line 845: This doesn't address #1445's array/scalar mismatch: `np.float64(0.0)` is still a scalar, so `np.array([np.array([1.0], dtype=np.float32), np.float64(0.0)], dtype=np.float64)` still raises the same inhomogeneous-shape error. The CartPole test only returns scalar rewards, so it doesn't exercise the reported case.

pseudo-rnd-thoughts (changes requested):
Thanks for the PR, I had missed that issue. I'm running the test on main and can't cause the test to fail, neither on numpy 1.x or 2.x using a macbook. How did you confirm the original issue / update the test to trigger the original issue

gymnasium/vector/async_vector_env.py
Fstarnb: You're right — thanks for calling this out.

`np.float64(0.0)` alone does **not** fix the array/scalar case. I re-checked stacking:

```python
np.array([np.array([1.0], dtype=np.float32), np.float64(0.0)], dtype=np.float64)
# ValueError: inhomogeneous shape
```

So the previous worker-side cast was incomplete for #1445.

**Update:** `step_wait` now goes through `_batch_rewards()`, which detects a shared non-scalar reward shape across workers and broadcasts autoreset scalar zeros to `np.zeros(that_shape)` before stacking. That handles the length-1 (and generally same-shaped) ndarray reward + autoreset mix.

There's a new regression test `test_async_vector_env_next_step_autoreset_array_reward` that uses a custom env returning `np.array([1.0], dtype=np.float32)` with staggered episode lengths. It **fails on current main** with the same inhomogeneous `ValueError` and **passes** with this change (shared_memory True/False).

Fstarnb:
Thanks @pseudo-rnd-thoughts and @Sanjays2402 — both reviews were correct, and I went back and verified the original failure properly.

### Why the CartPole test did not fail on main

You are right that the original test does **not** reproduce #1445. I confirmed on current `main` (numpy 2.3.x, macOS arm64):

| Reward type from sub-env | NEXT_STEP autoreset mix | Result on main |
|---|---|---|
| CartPole scalar (`float` / `np.float32`) | scalar `0` from autoreset workers | **OK** — stacks fine |
| `np.array(1.0, dtype=np.float32)` (0-d) | scalar `0` | **OK** |
| `np.array([1.0], dtype=np.float32)` (1-d) | scalar `0` | **ValueError** inhomogeneous shape at `step_wait` → `np.array(rewards, dtype=np.float64)` |

So #1445 is real, but it is the **array reward + autoreset scalar zero** case, not CartPole. The issue text (“numpy arrays of np.float32 and int”) matches that path; CartPole never hits it. Sorry for the weak first test — that was on me.

### Minimal repro (fails on main)

```python
import numpy as np
import gymnasium as gym
from gymnasium.vector import AsyncVectorEnv, AutoresetMode

class ArrayRewardEnv(gym.Env):
    def __init__(self, ep_len=3):
        super().__init__()
        self.observation_space = gym.spaces.Box(-1, 1, (2,), np.float32)
        self.action_space = gym.spaces.Discrete(2)
        self.ep_len, self.t = ep_len, 0
    def reset(self, *, seed=None, options=None):
        super().reset(seed=seed); self.t = 0
        return np.zeros(2, np.float32), {}
    def step(self, action):
        self.t += 1
        return np.zeros(2, np.float32), np.array([1.0], np.float32), self.t >= self.ep_len, False, {}

envs = AsyncVectorEnv(
    [lambda i=i: ArrayRewardEnv(2 + i) for i in range(4)],
    shared_memory=False,
    autoreset_mode=AutoresetMode.NEXT_STEP,
)
envs.reset(seed=0)
for _ in range(10):
    envs.step(envs.action_space.sample())  # crashes once any worker autoresets
```

### Fix

1. Keep autoreset worker reward as scalar `np.float64(0.0)` (same as “no reward this step”).
2. In `step_wait`, batch via `_batch_rewards()`: if non-scalar rewards share one shape, expand scalar placeholders to `zeros(shape)` then `np.stack`.
3. Regression test `test_async_vector_env_next_step_autoreset_array_reward` — **red on main, green with this PR** (I verified by swapping only `async_vector_env.py` back to main).
4. Kept a CartPole scalar smoke test so normal vector envs stay covered.

Happy to adjust the helper or test style if you prefer a different approach.

pseudo-rnd-thoughts:
This isn't an issue with `AsyncVectorEnv` but with the users environment, therefore, I'm closing this PR

PR #1627, opened 2026-07-11
Use trained-policy GIFs for Box2D environments (LunarLander, BipedalWalker)

aminehd:
# Description

Part of #1610, and a follow-up to #1612 (classic control). Swaps the random-agent GIFs for two of the Box2D envs — LunarLander and BipedalWalker — with renders of trained Stable-Baselines3 policies, so the docs show the envs being solved instead of flailing.

Only the 2 `.gif` files change. Rendering follows `docs/_scripts/gen_gifs.py` (300 frames, `duration=50`, `.unwrapped`) — the one difference is the action comes from `model.predict()` instead of `action_space.sample()`.

- `LunarLander-v3`: DQN, mean greedy return ~212 (descends under thruster control and lands on the pad)
- `BipedalWalker-v3`: SAC, mean greedy return ~307 (steady walking gait across the full course)

CarRacing is left for a later PR — it's pixel-based and needs a CnnPolicy.

Fixes # (part of #1610)

## Type of change

- [x] Documentation only change (no code changed)

## Reproduce

Scripts + configs live in a small side repo
`gymlab/` → [`packages/gymbench`](https://github.com/aminehd/gymbench) (generator) + `external/Gymnasium` (this fork)

```bash
git clone --recurse-submodules https://github.com/aminehd/gymlab
cd gymlab && pip install -e packages/gymbench
pip install "gymnasium[box2d]"
gymbench env-gif run --config box2d          # trains + renders both from the config
```

- core logic that generates the gif
https://github.com/aminehd/gymbench/blob/ac00d5e/src/gymbench/core/render.py#L49-L53
- config: https://github.com/aminehd/gymbench/blob/master/configs/env_gif/box2d.toml
- how it works: https://github.com/aminehd/gymbench/blob/master/README.md

# Checklist

- [x] Documentation only — no code changed, no new warnings

PR #1646, opened 2026-07-19
Use a trained-policy GIF for CarRacing

aminehd:
# Description

Part of #1610, and a follow-up to #1627 (LunarLander + BipedalWalker). Swaps the random-agent GIF for CarRacing with a render of a trained Stable-Baselines3 policy, so the docs show the env being driven instead of spinning in the grass.

Only the 1 `.gif` file changes. Rendering follows `docs/_scripts/gen_gifs.py` (300 frames, `duration=50`, `.unwrapped`) — the one difference is the action comes from `model.predict()` instead of `action_space.sample()`.

- `CarRacing-v3`: PPO + CnnPolicy, mean greedy return ~429 — drives the track through the straights and holds the road through curves.

Unlike the MLP envs, CarRacing is pixel-based, so it needs a CnnPolicy plus two things the others don't: several parallel envs and a stack of recent frames (so the CNN can perceive motion). This uses the rl-baselines3-zoo PPO recipe — 8 envs + a 2-frame stack + 4M steps.

Fixes # (part of #1610)

## Type of change

## Reproduce

Scripts + configs live in a small side repo
`gymlab/` → [`packages/gymbench`](https://github.com/aminehd/gymbench) (generator) + `external/Gymnasium` (this fork)

```bash
git clone --recurse-submodules https://github.com/aminehd/gymlab
cd gymlab && pip install -e packages/gymbench
pip install "gymnasium[box2d]"
gymbench env-gif run --config box2d          # trains + renders all three Box2D envs
```

- core logic that generates the gif
https://github.com/aminehd/gymbench/blob/662e988/src/gymbench/core/render.py#L53-L57
- frame stacking that lets render/evaluate match the trained policy: https://github.com/aminehd/gymbench/blob/662e988/src/gymbench/core/framestack.py
- config: https://github.com/aminehd/gymbench/blob/master/configs/env_gif/box2d.toml
- how it works: https://github.com/aminehd/gymbench/blob/master/README.md

# Checklist


jkterry1:
Hey thank you - We'd like a more optimal policy than this for the demo gif on the site though

aminehd:
Yeah that makes sense. I'm gonna try different setting. Do you have any suggestion?

pseudo-rnd-thoughts (approved):
LGTM, though reminds me that we never really solved car racing's reward function

PR #1648, opened 2026-07-23
Fix Discrete.contains raising OverflowError for out-of-dtype ints

nikolauspschuetz:
## Description

`Discrete.contains(x)` raises `OverflowError` instead of returning `False` when `x` is a Python `int` outside the range representable by the space's dtype:

```python
Discrete(5).contains(10**20)   # OverflowError: Python int too large to convert to C long
```

The normalization cast `np.dtype(self.dtype).type(x)` overflows for such values. Since a value that doesn't fit the space's dtype cannot be a member, `contains` should return `False`. Companion to #1616, which fixed the *comparison* overflow but not this *input-cast* one.

The fix guards the cast (`try/except OverflowError -> return False`) and adds a regression test covering `±10**20`, `2**63`, and `-2**63 - 1`.

Verified: fails on `main` (`OverflowError` at `discrete.py:204`), passes with the fix; full `tests/spaces/test_discrete.py` green (38 passed) on Python 3.12 / numpy 2.5.1.

## Type of change

---

<sub>Disclosure, for the record: Claude Code helped spot this and draft the fix/test; I reproduced and verified it myself.</sub>

### DIFF
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -201,10 +201,10 @@ class Discrete(Space[_IntegerT_co], Generic[_IntegerT_co]):
         Also checks if the given value, when provided as an int, can be cast to the space's dtype.
         """
         if isinstance(x, int):
-            try:
-                as_np = np.dtype(self.dtype).type(x)
-            except OverflowError:
+            dtype_info = np.iinfo(self.dtype)
+            if not (dtype_info.min <= x <= dtype_info.max):
                 return False
+            as_np = np.dtype(self.dtype).type(x)
 
         elif isinstance(x, (np.generic, np.ndarray)) and (
             np.issubdtype(x.dtype, np.integer) and x.shape == ()
--- a/gymnasium/spaces/discrete.py
+++ b/gymnasium/spaces/discrete.py
@@ -201,7 +201,10 @@ class Discrete(Space[_IntegerT_co], Generic[_IntegerT_co]):
         Also checks if the given value, when provided as an int, can be cast to the space's dtype.
         """
         if isinstance(x, int):
-            as_np = np.dtype(self.dtype).type(x)
+            try:
+                as_np = np.dtype(self.dtype).type(x)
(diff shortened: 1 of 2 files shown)


gymnasium/spaces/discrete.py
```diff
@@ -201,7 +201,10 @@ def contains(self, x: Any) -> bool:
         Also checks if the given value, when provided as an int, can be cast to the space's dtype.
         """
         if isinstance(x, int):
-            as_np = np.dtype(self.dtype).type(x)
+            try:
+                as_np = np.dtype(self.dtype).type(x)
+            except OverflowError:
```
nikolauspschuetz: Rationale: casting a Python `int` to the space's dtype (`np.dtype(self.dtype).type(x)`) raises `OverflowError` when `x` is outside the dtype range (e.g. `Discrete(5).contains(10**20)`). A value that does not fit the dtype cannot be a member, so we return `False` rather than let the cast raise. Companion to #1616, which fixed the comparison overflow but not this input-cast one.

Sanjays2402: This only catches values large enough for NumPy to raise. With the supported NumPy 1.x, `Discrete(3, start=125, dtype=np.int8).contains(-129)` casts `-129` to `127` and returns `True`; checking the Python int against `np.iinfo(self.dtype)` before the cast would cover both paths.

nikolauspschuetz: Good catch — you're right. The `OverflowError` guard only covered numpy 2.x; on numpy 1.x `np.int8(-129)` silently wraps to `127` and was wrongly reported as contained. Switched to checking the Python int against `np.iinfo(self.dtype)` before the cast (0419669), which handles both paths, and extended the test with the small-dtype wrap cases. Verified across numpy 1.22 → 2.3 on Python 3.10–3.13.

pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1653, opened 2026-07-29
Fix Text space accepting multi-character charset elements

### LINKED ISSUE #1317: [Bug Report] Text Space has limited support for multi-char symbols
### Describe the bug

I was messing around with categorical actions, where one word out of a predefined word set could be selected (e.g. "sit", "walk", "fly"). My first attempt as to define the charset as the whole words and limit the size to 1

```python
S = frozenset(['sit', 'walk', 'fly'])
space = Text(max_length=1, charset=S)
```

It actually worked, and I was very glad this could be supported. However it seems like the __contains__ implementation for this space has the supposition that all strings in charset have length 1 (which makes sense as its primary use is to store characters). Therefore, `space.contains('sit')` always returns False, even if it should be True.

```python
    ...

    def contains(self, x: Any) -> bool:
        """Return boolean specifying if x is a valid member of this space."""
        if isinstance(x, str):
            if self.min_length <= len(x) <= self.max_length:
                return all(c in self.character_set for c in x)

        return False
```

Should Text allow multi-char symbols?

### Code example

```shell
from gymnasium.spaces import Text

S = frozenset(['sit', 'walk', 'fly'])
space = Text(max_length=1, charset=S)

assert space.contains('sit') # Assertion Error
```

### System info

Gymnasium was installed via pip in a conda environment
Gymnasium version: 1.0.0
OS: Ubuntu 20.04
Python version: 3.11.11

### Additional context

_No response_

### Checklist

adarshsm:
# Description

`Text` validates its length bounds but never checks that each `charset` element is a single character. Because `sample` joins whole elements, a multi-character charset produces strings that are longer than `max_length` and that `contains` then rejects:

```python
>>> from gymnasium.spaces import Text
>>> space = Text(max_length=3, charset=frozenset(["ab", "cd"]))
>>> space.seed(0)
>>> sample = space.sample()
>>> sample
'cdcdcd'
>>> len(sample), space.max_length
(6, 3)
>>> space.contains(sample)
False
```

So `space.contains(space.sample())` is `False` — the invariant `tests/spaces/test_spaces.py::test_sample_contains` asserts for every registered space. It never fired because every `Text` space in `tests/spaces/utils.py` uses a single-character charset. The reporter's original case is the same defect with `max_length=1`, where the space additionally reprs as `Text(1, 1, charset=flysitwalk)`.

This PR raises a `ValueError` at construction rather than adding multi-character support, following your comment on #1317:

> The intention is that each element of the `charset` has length == 1.

That intention wasn't enforced or documented, so the only signal a user got was a sampled value their own space rejected. The empty string is rejected by the same check, since it also can't round-trip through `contains` when `min_length >= 1`. Single-character charsets, given either as a string or as a set, are unaffected — there's an assertion in the test covering that.

Fixes #1317

## Type of change

Strictly speaking this turns three previously-silent constructions into errors: multi-character elements, and the empty string. Anything relying on that was already producing samples outside its own space, so I've treated it as a bug fix rather than a breaking change — happy to be corrected. Nothing in `gymnasium/` or `tests/` constructs a `Text` with a multi-character charset, so nothing in-tree changes behaviour.

# Checklist:

## Testing

`test_multi_character_charset_raises` in `tests/spaces/test_text.py` fails on `main` with `Failed: DID NOT RAISE ValueError` and passes with the change. It covers the set path, the sequence path (which reports elements in the given order), the empty string, and — to confirm the fix isn't over-broad — that single-character charsets still satisfy `contains(sample())` and the `max_length` bound.

`tests/spaces/` is green: 1332 passed. Across the wider suite I get 86 failures both with and without this change (byte-identical lists), all from optional dependencies missing locally — Box2D, pygame, moviepy — so none are related. `pre-commit run --all-files` passes clean, including `ty`.

One aside in case it's useful: running the `ty` hook leaves a `uv.lock` in the repo root, and it isn't in `.gitignore`, so it shows up as untracked after a `pre-commit run --all-files`.

### DIFF
--- a/gymnasium/spaces/text.py
+++ b/gymnasium/spaces/text.py
@@ -46,7 +46,11 @@ class Text(Space[str]):
             min_length (int): Minimum text length (in characters). Defaults to 1 to prevent empty strings.
             max_length (int): Maximum text length (in characters).
             charset (Union[set], str): Character set, defaults to the lower and upper english alphabet plus latin digits.
+                Each element must be a single character.
             seed: The seed for sampling from the space.
+
+        Raises:
+            ValueError: If any element of ``charset`` is not a single character.
         """
         if not np.issubdtype(type(min_length), np.integer):
             raise TypeError(
@@ -75,6 +79,16 @@ class Text(Space[str]):
         else:
             # Preserve the given ordering, dropping duplicate characters
             char_list = list(dict.fromkeys(charset))
+
+        # Each element of the charset must be a single character, otherwise `sample`
+        # concatenates multi-character elements and produces strings that are longer
+        # than `max_length` and that `contains` rejects.
+        invalid_chars = [char for char in char_list if len(char) != 1]
+        if invalid_chars:
+            raise ValueError(
+                "Expects all charset elements to be a single character, actual invalid "
+                f"elements: {invalid_chars}"
+            )
 
         self._char_set: frozenset[str] = frozenset(char_list)
         self._char_list: tuple[str, ...] = tuple(char_list)
--- a/tests/spaces/test_text.py
+++ b/tests/spaces/test_text.py
@@ -94,6 +94,38 @@ def test_charset_ordering():
     assert [space.character_index(c) for c in "dcba"] == [0, 1, 2, 3]
 
 
+def test_multi_character_charset_raises():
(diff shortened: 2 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR and spotting the issue

PR #1654, opened 2026-08-01
Use a trained-policy GIF for HalfCheetah

natsu0529:
# Description

Part of #1610. Replaces the random-action HalfCheetah GIF with a render of a trained Stable-Baselines3 policy, so the docs show a stable forward-running gait.

- Environment: `HalfCheetah-v5`
- Algorithm / policy: SAC / `MlpPolicy`
- Training: 1,000,000 timesteps, seed 0
- Deterministic evaluation over 10 episodes: 11,772.82 +/- 122.48 mean reward
- Visual check: sustained forward motion with a stable gait, no falls or resets (160.68 position units over the capture)

Rendering follows `docs/_scripts/gen_gifs.py`: 301 frames, 50 ms per frame, infinite loop, and `.unwrapped`. Actions come from `model.predict(..., deterministic=True)` instead of `action_space.sample()`. The MuJoCo `track` camera keeps the trained runner in frame. For rendering only, the floor extent and checker repeat count are increased proportionally so the ground remains visible for the full run; environment dynamics and the trained policy are unchanged.

Only `docs/_static/videos/mujoco/half_cheetah.gif` changes.

## Type of change

## Checks

- [x] All 301 frames retain the floor
- [x] `pre-commit run --all-files`
- [x] `make dirhtml`


jkterry1:
Hey this is great, thank you so much :)

You should do another one if you have a moment

PR #1657, opened 2026-08-02
Fix AsyncVectorEnv crash on autoreset with non-scalar rewards

### LINKED ISSUE #1445: [Bug Report] Autoreset in AsyncVectorEnv causes crash in 1.2.0

Danny-397:
Closes #1445.

## The bug

On a `NEXT_STEP` autoreset, `_async_worker` substitutes scalars for the environment's return values:

```python
reward, terminated, truncated = 0, False, False
```

`step_wait` then batches the per-env results with `np.array(rewards, dtype=np.float64)`. If an environment returns a reward of shape `(1,)` rather than a bare scalar, that list contains a shape-`(1,)` array *and* a Python `int`, which NumPy rejects:

```
ValueError: setting an array element with a sequence. The requested array has an
inhomogeneous shape after 1 dimensions.
```

## Why this looks like a bug rather than a restriction

`SyncVectorEnv` does **not** have this problem. It assigns into a preallocated buffer (`self._rewards[i] = ...`), so each value is coerced independently and a shape-`(1,)` reward is accepted. The same environment therefore works under `SyncVectorEnv` and crashes under `AsyncVectorEnv`:

| reward returned by env | `SyncVectorEnv` | `AsyncVectorEnv` (before) |
|---|---|---|
| 0-d array | works | works |
| **shape `(1,)`** | **works** | **`ValueError`** |
| shape `(2,)` | fails | fails |

Reproduced on `main` at 1.3.0, both `shared_memory=True` and `False`.

## The fix

Build the reward, termination and truncation batches the way `SyncVectorEnv` does — preallocate, then assign per sub-environment — so the two backends agree by construction.

## Scope

This deliberately does **not** add support for genuinely multi-element rewards: a shape-`(2,)` reward still raises under both backends, exactly as before. Making that work would be an API change rather than a bug fix, so it seemed out of scope here.

## Testing

Added `test_autoreset_next_step_non_scalar_reward` to `tests/vector/test_autoreset_mode.py`, parametrized over `Sync` / `Async(shared_memory=True)` / `Async(shared_memory=False)` in the style of the surrounding tests.

Without the fix it **fails on both Async variants and passes on Sync**, which is precisely the divergence being corrected.

`tests/vector/` is otherwise unaffected: 1448 passing. Five tests fail both with and without this change — they need Box2D and a rendering backend that aren't installed locally, and three of them are in `test_sync_vector_env.py`, which this change does not touch.

`pre-commit` (ruff check + ruff format) is clean.

---

*Disclosure: this fix was developed with AI assistance. The diagnosis, patch and test were reviewed before submission, and I'm happy to explain or revise any part of it.*

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -512,11 +512,29 @@ class AsyncVectorEnv(VectorEnv):
             )
 
         self._state = AsyncState.DEFAULT
+
+        # Build these the way `SyncVectorEnv` does -- preallocate, then assign
+        # per sub-environment -- rather than `np.array(list_of_values)`.
+        # On an autoreset step the worker substitutes the scalars `0`, `False`,
+        # `False`, so a list-based conversion sees a mix of those scalars and
+        # whatever the environment actually returns. If an environment returns a
+        # single-element array reward (shape `(1,)`), that list is inhomogeneous
+        # and NumPy raises, even though `SyncVectorEnv` accepts the same
+        # environment. Per-index assignment coerces each value identically to
+        # `SyncVectorEnv`, so both vector backends now behave the same.
+        rewards_arr = np.zeros((self.num_envs,), dtype=np.float64)
+        terminations_arr = np.zeros((self.num_envs,), dtype=np.bool_)
+        truncations_arr = np.zeros((self.num_envs,), dtype=np.bool_)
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Closing as this is a bug with the environment and not with the AsyncVectorEnv implementation

PR #1662, opened 2026-08-08
Use trained-policy GIFs for the MuJoCo environments

Amala-Bharadwaj:
# Description

Part of #1610, following #1612 (classic control) and #1627 (Box2D). Replaces the random-agent GIFs for all 11 MuJoCo environments with renders of trained SB3 SAC policies, so the docs show the environments being solved instead of flailing.

Only the 11 `.gif` files change. Rendering follows `docs/_scripts/gen_gifs.py` — 300 frames, `duration=50`, `loop=0` — with the action coming from `model.predict(..., deterministic=True)` instead of `action_space.sample()`. Every gif plays for the same 15.05s as the one it replaces.

Hyperparameters are rl-baselines3-zoo's MuJoCo SAC settings (SAC defaults plus `learning_starts=10000`, Humanoid at 2e6 steps, Swimmer at `gamma=0.999`). Mean return is over 10 deterministic episodes:

| environment | steps | mean return |
|---|---|---|
| HalfCheetah-v5 | 1M | 11386.1 +/- 181.2 |
| InvertedDoublePendulum-v5 | 300k | 9358.9 +/- 0.3 |
| Ant-v5 | 1M | 3997.5 +/- 1415.2 |
| Walker2d-v5 | 1M | 3683.0 +/- 431.6 |
| Hopper-v5 | 1M | 3002.8 +/- 726.8 |
| Humanoid-v5 | 2M | 5723.6 +/- 915.0 |
| HumanoidStandup-v5 | 2M | 160159.8 +/- 67.5 |
| Swimmer-v5 | 1M | 336.1 +/- 0.9 |
| InvertedPendulum-v5 | 100k | 1000.0 +/- 0.0 |
| Reacher-v5 | 300k | -3.6 +/- 1.2 |
| Pusher-v5 | 500k | -27.4 +/- 3.4 |

Three deviations from `gen_gifs.py`, all about keeping the subject visible and moving rather than about the policy. Happy to drop any of them if you'd rather they matched exactly:

- **Reacher and Pusher keep their `TimeLimit`.** Their episodes cap at 50 and 100 steps, so unwrapped they solve the task once and then hold the pose for the remaining ~250 frames. Keeping the wrapper lets the existing reset-on-truncation show several episodes. The other 9 render unwrapped as before.
- **Reset when the agent leaves the rendered floor.** MuJoCo floor planes collide infinitely but only draw out to their geom size (40 for most, 20 for Hopper and the Humanoids). A trained HalfCheetah covers ~165m in 15s, so it ran off the checkerboard around frame 100 and spent the rest of the gif against blank grey. Rendering now resets once the root body passes 90% of the floor half-extent.
- **Reset once the picture stops changing.** A solved InvertedPendulum or InvertedDoublePendulum comes to rest and holds one pose, so unwrapped those two gifs were a still image for most of their length. Rendering now resets after 20 consecutive frames with a mean per-pixel change under 0.3, which shows several recoveries from different random initial states instead. The threshold is measured rather than guessed: settled pendulums sit at 0.01-0.05, and the quietest genuinely-moving environment (HumanoidStandup) averages 1.5, so nothing else triggers it.

Total size goes from 74.2 MB to 68.6 MB.

Fixes # (part of #1610)

## Type of change

## Reproduce

Each environment is trained with SB3 SAC, evaluated over 10 deterministic episodes for the number above, then rendered:

```python
model = SAC("MlpPolicy", gym.make(env_id), learning_starts=10_000, seed=0)
model.learn(total_timesteps=n_timesteps)

env = gym.make(env_id, render_mode="rgb_array")
if env.spec.max_episode_steps >= 300:      # Reacher and Pusher keep their TimeLimit
    env = env.unwrapped

frames, obs = [], env.reset(seed=0)[0]
while len(frames) <= 300:
    frames.append(Image.fromarray(env.render()))
    action, _ = model.predict(obs, deterministic=True)
    obs, _, terminated, truncated, _ = env.step(action)
    if terminated or truncated or left_floor(env) or settled(frames):
        obs, _ = env.reset()

frames[0].save(path, save_all=True, append_images=frames[1:], duration=50, loop=0)
```

`left_floor` compares the root body's xy position against the floor geom's half-extent; `settled` checks whether the last 20 frames changed by less than 0.3 mean per-pixel. Both are described above. Happy to push the full generator script somewhere public if that's useful for review.

# Checklist


jkterry1:
Hey thanks a ton

-Humanoid standup isn't learning the full behavior if you watch the gif
-The new walker2d gif almost looks worse?

Amala-Bharadwaj:
Thanks, I see your points. HumanoidStandup is undertrained, so I'm retraining it with a much larger step budget. For Walker2d I'm rerunning a couple of seeds and it in this PR if it beats the merged #1656 gif, otherwise I'll drop Walker2d and leave that one as is.

pseudo-rnd-thoughts:
@Amala-Bharadwaj Any update?

Amala-Bharadwaj:
I'll drop HumanoidStandup from this PR since it needs more training. I'll open a follow-up for it separately.

pseudo-rnd-thoughts (approved):
Thanks for the changes @Amala-Bharadwaj

PR #1663, opened 2026-08-13
Fix Box.contains raising OverflowError for out-of-dtype ints

nikolauspschuetz:
# Description

`Box.contains(x)` raises `OverflowError` instead of returning `False` when `x` is a Python `int` (or list of ints) outside the range representable by the space's dtype:

```python
import numpy as np
Box(low=0, high=10, shape=(2,), dtype=np.int8).contains([300, 0])
# OverflowError: Python integer 300 out of bounds for int8
```

When `x` is not already an ndarray, `contains` normalizes it via `np.asarray(x, dtype=self.dtype)`. On numpy 2.x that cast raises `OverflowError` for out-of-dtype ints, which propagates instead of being treated as "not a member". Since a value that doesn't fit the space's dtype cannot be a member, `contains` should return `False`.

The fix adds `OverflowError` to the existing `except (ValueError, TypeError)` guard around the cast, and adds a regression test. This mirrors the just-merged #1648, which fixed the same input-cast overflow in `Discrete.contains`.

Verified: fails on `main` (`OverflowError` at `box.py:485`), passes with the fix; full `tests/spaces/test_box.py` green (116 passed) on Python 3.12 / numpy 2.5.2.

## Type of change

# Checklist:

---

<sub>Disclosure, for the record: Claude Code helped spot this and draft the fix/test; I reproduced and verified it myself.</sub>

### DIFF
--- a/gymnasium/spaces/box.py
+++ b/gymnasium/spaces/box.py
@@ -483,7 +483,7 @@ class Box(Space[NDArray[_ScalarT_co]]):
             gym.logger.warn("Casting input x to numpy array.")
             try:
                 x = np.asarray(x, dtype=self.dtype)
-            except (ValueError, TypeError):
+            except (ValueError, TypeError, OverflowError):
                 return False
 
         return bool(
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1664, opened 2026-08-15
Record the constructor arguments missing from five wrappers

mkzung:
### The defect

Five wrappers leave constructor arguments out of `RecordConstructorArgs`. `Wrapper.spec` puts the
recorded arguments into a `WrapperSpec` and `gym.make` splats them back in, so an environment
rebuilt from its own spec comes back with the defaults in their place:

```python
env = DiscretizeAction(base, bins=2, multidiscrete=True)
gym.make(env.spec).action_space   # MultiDiscrete([2]) becomes Discrete(2)

env = RecordEpisodeStatistics(base, stats_key="my_stats")
gym.make(env.spec)                # writes to info["episode"] again
```

Same for `DiscretizeObservation`, `RecordVideo` and `StickyAction`.

Nothing raises. The rebuilt environment is quietly a different environment, which is the same shape
as #1658 in `AddRenderObservation`.

`StickyAction` needs more than one line. Recording the post-widening value would put a tuple in the
spec even where the caller passed an int, and `EnvSpec.from_json` returns a list, which the
constructor rejects. The value the caller gave is kept for the record; the widened pair is used
internally.

### The test

Walks `gymnasium.wrappers.__all__`, parses each `__init__` with `ast`, finds the
`RecordConstructorArgs.__init__` call and compares its keyword names against the signature. It adds
a round trip through `EnvSpec` too, because the existing one in
`tests/envs/registration/test_env_spec.py` builds a stack of two wrappers that both already record
everything, so it could not have caught any of this.

Reverting the five sources and keeping the tests gives 6 failures: one per wrapper, plus the round
trip.

### Left alone

Running that walk over the whole package rather than `__all__` leaves three classes standing, which
invites the obvious question. `ArrayConversion`, in both the plain and vector namespaces, takes four
Array API namespaces, and those are modules, which `RecordConstructorArgs` cannot deepcopy; it sits
in the new test as `xfail(strict=True)` so it fails loudly the day somebody finds a different answer
for it. `gymnasium/wrappers/vector/rendering.py::RecordVideo` has a plain gap of the kind fixed
here, but `VectorWrapper.spec` returns `self.env.spec` unchanged and never builds a `WrapperSpec`,
so nothing reads those arguments back and there is no round trip to break.

### Validation

`tests/wrappers`: 429 passed, 16 skipped, 1 xfailed. The 12 failures are
`test_human_rendering.py::test_num_envs_screen_size` with `Ant-v4`, identical name for name on an
untouched `main`, where the same run gives 393 passed and the same 12.

### DIFF
--- a/gymnasium/wrappers/common.py
+++ b/gymnasium/wrappers/common.py
@@ -496,7 +496,9 @@ class RecordEpisodeStatistics(
             buffer_length: The size of the buffers :attr:`return_queue`, :attr:`length_queue` and :attr:`time_queue`
             stats_key: The info key for the episode statistics
         """
-        gym.utils.RecordConstructorArgs.__init__(self)
+        gym.utils.RecordConstructorArgs.__init__(
+            self, buffer_length=buffer_length, stats_key=stats_key
+        )
         gym.Wrapper.__init__(self, env)
 
         self._stats_key = stats_key
--- a/gymnasium/wrappers/rendering.py
+++ b/gymnasium/wrappers/rendering.py
@@ -271,7 +271,9 @@ class RecordVideo(
             step_trigger=step_trigger,
             video_length=video_length,
             name_prefix=name_prefix,
+            fps=fps,
             disable_logger=disable_logger,
+            gc_trigger=gc_trigger,
         )
         gym.Wrapper.__init__(self, env)
 
--- a/gymnasium/wrappers/stateful_action.py
+++ b/gymnasium/wrappers/stateful_action.py
@@ -64,33 +64,40 @@ class StickyAction(
                 f"`repeat_action_probability` should be in the interval [0,1). Received {repeat_action_probability}"
             )
 
+        # The pair of bounds that the number of repeats is drawn from.
+        # `repeat_action_duration` keeps the value the caller gave, which is what
+        # `RecordConstructorArgs` stores for the env spec.
         if isinstance(repeat_action_duration, int):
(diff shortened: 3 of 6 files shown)


mkzung:
CI caught two things my local run could not.

The three array converters, `JaxToNumpy`, `JaxToTorch` and `NumpyToTorch`, never call `RecordConstructorArgs` at all. Jax and torch are not installed on my machine, so the test skipped them and the assertion was never exercised. Not recording an argument and not recording at all are separate gaps, so the test now skips a wrapper that does not record, and those three are worth their own change.

The other was a stray blank line after the imports, which `ruff` fixes.

pseudo-rnd-thoughts (approved):
Thanks for the PR, appreciate the tests

mkzung:
The three red jobs here are `JaxToNumpy`, `JaxToTorch` and `NumpyToTorch` failing the new test with "does not call RecordConstructorArgs.__init__". All three subclass `ArrayConversion` and never call it in their own constructors, and the test reads each class's own `__init__`, so it raises before the skip below it can fire.

#1667 fixes that by having each converter record its own arguments, and it is green. It stacks on this branch, so merging it clears these three as well.

pseudo-rnd-thoughts:
@mkzung Could you next time stack the changes in this PR or have the other PR only include the new changes as this requires that I merge this broken PR then have the second PR fix the issues

PR #1665, opened 2026-08-15
Reject a min action equal to the max action in RescaleAction

mkzung:
### The defect

`RescaleAction` applies the inverse of the affine map that `rescale_box` builds. A component whose
new min equals its new max is rescaled onto a single point, so the gradient is zero and the inverse
divides by it.

```python
env = RescaleAction(base, 0.0, 0.0)   # builds without complaint
env.step(env.action_space.sample())   # every action is nan
```

Nothing checks the action on the way through, so the `nan` reaches the environment and only surfaces
later as a `nan` observation.

### The change

Raise `InvalidBound` at construction, which is what `ClipReward` raises for bounds it cannot use.

The check is `act_space.low == act_space.high` on the rescaled space, and exact equality is the
right test rather than a tolerance: bounds a denormal apart still have an inverse, an enormous one
but finite, while equal bounds have none at all. Zero gradient is the whole defect.

`RescaleObservation` keeps working with equal bounds. It applies the forward map, and a constant
inside its own observation space is well defined.

### The test

Two cases in `tests/wrappers/test_rescale_action.py`: equal scalar bounds, and one component of a
vector action equal while the others are fine. Reverting the source and keeping them gives one
failure.

### Validation

`tests/wrappers`: 394 passed, 13 skipped. The 12 failures are
`test_human_rendering.py::test_num_envs_screen_size` with `Ant-v4`, identical name for name on an
untouched `main`, where the same run gives 393 passed and the same 12.

### DIFF
--- a/gymnasium/wrappers/transform_action.py
+++ b/gymnasium/wrappers/transform_action.py
@@ -14,6 +14,7 @@ import numpy as np
 
 import gymnasium as gym
 from gymnasium.core import ActType, ObsType, WrapperActType
+from gymnasium.error import InvalidBound
 from gymnasium.spaces import Box, Discrete, MultiDiscrete, Space
 
 __all__ = ["TransformAction", "ClipAction", "RescaleAction"]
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1668, opened 2026-08-16
Fix vector NormalizeReward under SAME_STEP autoreset

teddytennant:
`vector.NormalizeReward` always skips sub-envs whose previous step terminated. That is correct for `NEXT_STEP`, where the following step is a dummy reset. Under `SAME_STEP` the env already reset inside the terminating step, so the next step is a real first step of the new episode.

The wrapper then:

1. Drops that first-step reward from `return_rms`.
2. Leaves the previous episode's return in `accumulated_reward`.

On a 4-step dummy, 12 steps, 2 envs, `return_rms.count` was 20 instead of 24.

When `autoreset_mode` is `SAME_STEP`, treat every step as active and zero `accumulated_reward` on the terminating step, the same way `RecordEpisodeStatistics` already does (#1624). `NEXT_STEP` is unchanged.

`pytest tests/wrappers/vector/test_normalize_reward.py` and the related vector wrapper / autoreset tests pass (86).

### DIFF
--- a/gymnasium/wrappers/vector/stateful_reward.py
+++ b/gymnasium/wrappers/vector/stateful_reward.py
@@ -11,7 +11,7 @@ import numpy as np
 
 import gymnasium as gym
 from gymnasium.error import InvalidBound
-from gymnasium.vector.vector_env import VectorEnv, VectorWrapper
+from gymnasium.vector.vector_env import AutoresetMode, VectorEnv, VectorWrapper
 from gymnasium.wrappers.utils import RunningMeanStd
 
 __all__ = ["NormalizeReward"]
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
@teddytennant Could you fix the merge conflicts then run `pre-commit run --all-files`

teddytennant:
Rebased on main and pushed. The conflicts were both additive, the `InvalidBound` checks that landed in #1669 next to my `AutoresetMode` import and test, so I kept both sides.

`pre-commit run --all-files` is clean now, ruff-format wanted my test signature on one line.

pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1669, opened 2026-08-16
Reject invalid gamma and epsilon in the normalize wrappers

DenisDrobyshev:
## Summary

`NormalizeReward` and `NormalizeObservation` accept any value for their numeric parameters. Every neighbouring wrapper in the package validates its arguments — `StickyAction` and `AddWhiteNoise` raise `InvalidProbability`, `RescaleAction`, `ClipReward` and `RepeatAction` raise `InvalidBound`/`ValueError` — so these four classes are the gap in an otherwise consistent convention.

The failure mode is silence rather than an exception, which is what makes it worth fixing:

* `epsilon` is added under a square root, `np.sqrt(var + epsilon)`. It exists to keep the division away from zero, so a non-positive value defeats its own purpose, and once `|epsilon|` exceeds the running variance the root is taken of a negative number. The wrapper keeps running and returns NaN.
* `gamma` multiplies the discounted accumulator on every step. Above one the accumulator diverges instead of converging; below zero it alternates sign. `gamma=99` — a plausible typo for `0.99` — sends it to infinity.

Measured on `Pendulum-v1` before this change, with nothing raised at construction time:

| argument | result |
|---|---|
| `NormalizeObservation(epsilon=-1.0)` | 67.5% of returned observations are NaN |
| `NormalizeReward(gamma=99)` | accumulator reaches `-inf`, 145 of 300 rewards are NaN |
| `NormalizeReward(gamma=2.0)` | accumulator reaches `-1.7e+90` |
| `wrappers.vector.NormalizeReward(gamma=2.0)` | accumulator reaches `inf` |

A user who hits this sees an agent that will not learn, with no error pointing at the wrapper.

## Changes

Validation in the four affected constructors, in the style already used by `RescaleAction`, with the reason recorded as a comment and a `Raises:` entry added to each docstring:

* `gymnasium/wrappers/stateful_reward.py` — `NormalizeReward`: `0 <= gamma <= 1`, `epsilon > 0`
* `gymnasium/wrappers/stateful_observation.py` — `NormalizeObservation`: `epsilon > 0`
* `gymnasium/wrappers/vector/stateful_reward.py` — vector `NormalizeReward`: both
* `gymnasium/wrappers/vector/stateful_observation.py` — vector `NormalizeObservation`: `epsilon > 0`

Both endpoints of the `gamma` interval stay valid: `0` keeps only the immediate reward and `1` is the undiscounted return, and both are covered by a test so the bound is not tightened by accident later.

Regression tests were added to the four matching test files.

## Verification

`tests/wrappers` and `tests/vector` were run before and after the change and the set of failing tests is identical — 57 failures in both, all pre-existing and caused by `Box2D`/`MuJoCo`/`torch` not being installed in my environment, none of them in the affected files. The four touched test files pass in full (40 passed). `ruff check`, `ruff format` and `codespell` are clean on all eight files.

The behaviour above can be reproduced on the parent commit with:

```python
import numpy as np, gymnasium as gym
from gymnasium.wrappers import NormalizeObservation

env = NormalizeObservation(gym.make("Pendulum-v1"), epsilon=-1.0)  # no error
obs, _ = env.reset(seed=0)
for _ in range(200):
    obs, *_ = env.step(env.action_space.sample())
print(obs)  # [nan nan ...]
```

### DIFF
--- a/gymnasium/wrappers/stateful_observation.py
+++ b/gymnasium/wrappers/stateful_observation.py
@@ -18,6 +18,7 @@ import numpy as np
 import gymnasium as gym
 import gymnasium.spaces as spaces
 from gymnasium.core import ActType, ObsType, WrapperActType, WrapperObsType
+from gymnasium.error import InvalidBound
 from gymnasium.spaces import Box, Dict, Tuple
 from gymnasium.vector.utils import batch_space, concatenate, create_empty_array
 from gymnasium.wrappers.utils import RunningMeanStd, create_zero_array
@@ -508,7 +509,18 @@ class NormalizeObservation(
         Args:
             env (Env): The environment to apply the wrapper
             epsilon: A stability parameter that is used when scaling the observations.
+
+        Raises:
+            InvalidBound: If ``epsilon`` is not strictly positive.
         """
+        # `epsilon` is added under the square root to keep the division away from
+        # zero, so a non-positive value defeats its purpose and, once it exceeds
+        # the variance, silently turns every normalized observation into NaN.
+        if epsilon <= 0:
+            raise InvalidBound(
+                f"`epsilon` should be strictly positive. Received {epsilon}"
+            )
+
         gym.utils.RecordConstructorArgs.__init__(self, epsilon=epsilon)
         gym.ObservationWrapper.__init__(self, env)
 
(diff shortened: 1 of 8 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1671, opened 2026-08-18
Report one error when RecordVideo cannot be constructed, not three

DenisDrobyshev:
## Description

Constructing `RecordVideo` on an environment whose render mode yields no image
reports three errors where one was intended:

```python
>>> gym.wrappers.RecordVideo(gym.make("CartPole-v1"), "videos")
Exception ignored in: <function RecordVideo.__del__ at 0x...>
  File "gymnasium/wrappers/rendering.py", line 438, in __del__
    if len(self.recorded_frames) > 0:
AttributeError: 'RecordVideo' object has no attribute 'recorded_frames'
ValueError: ('Render mode is None, which is incompatible with RecordVideo.', 'Initialize your environment with a render_mode that returns an image, such as rgb_array.')
```

Two separate causes.

The message is split across two arguments of `ValueError` rather than
concatenated, so `str(e)` renders a tuple and the sentence that tells the
caller what to do arrives quoted inside it. The comma was meant to be string
concatenation.

`recorded_frames` is assigned further down `__init__`, after this check, so
`__del__` runs on a half-built wrapper and raises an `AttributeError` of its
own. Python reports that as an ignored exception, printed *before* the error
that actually stopped construction — the first thing the reader sees is
`recorded_frames`, which has nothing to do with what they got wrong.

`__del__` now tolerates a partially initialised wrapper. That covers every
failure path in `__init__` rather than this one alone: any exception raised
before the assignment left the same trailing noise.

Both the plain and the vector wrapper carry the same two lines, so both are
fixed here.

## Type of change

## Checklist

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files`

Added to both `tests/wrappers/test_record_video.py` and
`tests/wrappers/vector/test_record_video.py`: the exception carries a single
argument, the guidance is in the message, and `__del__` on a wrapper that
never reached `recorded_frames` does not raise. Both fail on `main` and pass
with this change; the two suites together go from 26 to 28 passing.

### DIFF
--- a/gymnasium/wrappers/rendering.py
+++ b/gymnasium/wrappers/rendering.py
@@ -279,8 +279,8 @@ class RecordVideo(
 
         if env.render_mode in {None, "human", "ansi"}:
             raise ValueError(
-                f"Render mode is {env.render_mode}, which is incompatible with RecordVideo.",
-                "Initialize your environment with a render_mode that returns an image, such as rgb_array.",
+                f"Render mode is {env.render_mode}, which is incompatible with RecordVideo. "
+                "Initialize your environment with a render_mode that returns an image, such as rgb_array."
             )
 
         if episode_trigger is None and step_trigger is None:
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1672, opened 2026-08-19
Fix registration of wrapper class entry points

### LINKED ISSUE #967: make_vec(...) will raise error "AttributeError: can't set attribute", while vector.make(..) doesn't for customized env
### Describe the bug

Hi, 

I have some customized RL envs, and I want to create asychronized env vector to make them run in parallel. 
Firstly, I used the gymnasium.vector.make(...), and there is some warning saying that it will be deprecated and replaced by the gymnasium.make_vec() function. Thus I decided to use the new one. 

My code looks like:

```python
import fancy_gym
import gymnasium as gym
 
env_id = "metaworld/button-press-v2"
num_envs = 8
render = False

# Buggy
env = gym.make_vec(id=env_id, num_envs=num_envs, vectorization_mode="async",
                       render_mode='human' if render else None)

# Works fine
env = gym.vector.make(id=env_id, num_envs=num_envs,
                          render_mode='human' if render else None)

```  

The former make function will raise an error, as 
```
  File "/home/lige/Software/mamba/envs/tce/lib/python3.8/site-packages/gymnasium/envs/registration.py", line 977, in _create_env
    _env.spec = spec_
AttributeError: can't set attribute
```

I am not sure if the issue comes from my customized env, as it works ok with the old function.
To reproduce the bug easily, you may need to install the two packages:

```shell
conda create -n test_gym python=3.8
conda activate test_gym

pip install git+https://github.com/Farama-Foundation/Metaworld.git@master#egg=metaworld
pip install fancy_gym
``` 

where the metaworld is from your organization and fancy_gym is  our custimozed RL wrappers...

### Code example

```shell
import fancy_gym
import gymnasium as gym
 
env_id = "metaworld/button-press-v2"
num_envs = 8
render = False

# Buggy
env = gym.make_vec(id=env_id, num_envs=num_envs, vectorization_mode="async",
                       render_mode='human' if render else None)

# Works fine
env = gym.vector.make(id=env_id, num_envs=num_envs,
                          render_mode='human' if render else None)
```

### System info

gymnasium: 0.29.1, installed as a dependency of fancy_gym package (via pip)
Ubuntu 22.04
conda 23.11.0
python 3.8

### Additional context

Thank you in advance.

### Checklist

tandede:
# Description

Registered environment entry points can be `gymnasium.Wrapper` subclasses. Their class-level `metadata` attribute is an instance property that delegates to the wrapped environment, but `make()` currently validates that property object as if it were a concrete metadata dictionary. As a result, both `make()` and `make_vec()` reject valid wrapper entry points before constructing them.

This change skips only instance-property metadata during pre-construction inspection. Concrete class metadata is still validated normally. It also corrects the invalid-metadata diagnostic, which currently reports the imported `metadata` module's type instead of the invalid value's type.

Regression tests verify that `make_vec()` can create multiple environments from a registered wrapper class and that a concrete invalid metadata value is still rejected with the correct type.

Fixes #967

## Type of change

# Checklist

- [x] I have run the `pre-commit` checks on the changed files
- [x] I have commented the descriptor handling that is not immediately obvious
- [x] I have added tests that prove the fix is effective and preserve invalid-metadata validation
- [x] New and existing registration unit tests pass locally

# Testing

```
pre-commit run --files gymnasium/envs/registration.py tests/envs/registration/test_make_vec.py
pytest -q tests/envs/registration/test_make_vec.py
pytest -q tests/envs/registration
pytest -q
```

The focused suites report 41 passed and 107 passed, respectively. The complete suite reports 4,390 passed and 146 skipped. Two MuJoCo Humanoid velocity tests fail on exact equality at approximately 1e-16; both failures reproduce on an unmodified `main` worktree in the same environment.

### DIFF
--- a/gymnasium/envs/registration.py
+++ b/gymnasium/envs/registration.py
@@ -7,7 +7,6 @@ import copy
 import dataclasses
 import difflib
 import importlib
-import importlib.metadata as metadata
 import json
 import re
 from collections import defaultdict
@@ -472,7 +471,7 @@ def _check_metadata(testing_metadata: dict[str, Any]) -> None:
     """Check the metadata of an environment."""
     if not isinstance(testing_metadata, dict):
         raise error.InvalidMetadata(
-            f"Expect the environment metadata to be dict, actual type: {type(metadata)}"
+            f"Expect the environment metadata to be dict, actual type: {type(testing_metadata)}"
         )
 
     render_modes = testing_metadata.get("render_modes")
@@ -700,8 +699,11 @@ def make(
     # Determine if to use the rendering
     render_modes: list[str] | None = None
     if hasattr(env_creator, "metadata"):
-        _check_metadata(env_creator.metadata)
-        render_modes = env_creator.metadata.get("render_modes")
+        env_creator_metadata = env_creator.metadata
+        # Wrapper classes expose metadata through an instance property.
+        if not isinstance(env_creator_metadata, property):
+            _check_metadata(env_creator_metadata)
+            render_modes = env_creator_metadata.get("render_modes")
     render_mode = env_spec_kwargs.get("render_mode")
     apply_human_rendering = False
     apply_render_collection = False
--- a/tests/envs/registration/test_make_vec.py
+++ b/tests/envs/registration/test_make_vec.py
@@ -15,6 +15,19 @@ from gymnasium.wrappers import TimeLimit, TransformObservation
 from tests.wrappers.utils import has_wrapper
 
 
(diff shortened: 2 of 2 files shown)


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. All reported build, test, and pre-commit checks have passed, and the wrapper entry-point registration fix is ready for review. I’d be happy to address any feedback. Thank you!

pseudo-rnd-thoughts (approved):
Thanks for the PR, sorry for taking so long to review

PR #1674, opened 2026-08-24
Keep flatten and flatten_space in agreement for OneOf spaces

teddytennant:
`flatten` and `flatten_space` disagree about the dtype of a `OneOf` space, so a flattened sample is not contained in the space that is supposed to describe it. `FlattenObservation` on a `OneOf` observation space emits observations outside its own declared `observation_space`.

```
flatten_space: Box([ 0. -1. -1. -1.], 1.0, (4,), float32) float32
flatten()    : [1.  0.82711583  0.27574444  0.18777488] float64
```

Two causes. `_flatten_oneof` builds the result with `np.concatenate([[idx], flat_sample])`, and the bare list is int64, so it promotes the payload (float32 to float64, int8 to int64). Separately `_flatten_space_oneof` took its dtype from `s.dtype` of the raw subspaces instead of the flattened ones, so any `OneOf` holding a `Text` raised `ValueError: Invalid Box dtype (<U32)` outright.

`_flatten_space_tuple` and `_flatten_space_dict` already get this right: they promote over the *flattened* subspaces. So this is applying the existing rule rather than inventing one. Both sides now go through one helper.

The `_unflatten_oneof` hunk is the non-obvious one and it is required, not defensive. Once `flatten` emits a single promoted dtype, an `OneOf([Text(5), Box(-1, 1, (2,))])` flattens to float64, and `_unflatten_text` does `space.character_list[val]`, which gives `TypeError: tuple indices must be integers or slices, not numpy.float64`. That space round-trips on main today, so without the cast this fix would trade one broken path for another. The cast puts the slice back in the subspace's own flattened dtype.

One thing worth your call. `_oneof_flat_dtype` calls `flatten_space` per subspace, and it runs inside `flatten`, which is per-step under `FlattenObservation`. On my machine, for `OneOf` of two float32 Boxes:

```
flatten          8.0 us -> 94.1 us
unflatten        6.4 us -> 43.6 us
nested flatten  11.7 us -> 297 us
```

That is a real cost and I would rather you decide how to pay it than guess. Caching the dtype on the space is the obvious fix, but `OneOf.spaces` is a public mutable attribute so a naive cache goes stale. Happy to add one with an invalidation story, or to compute the dtype without building a `Box` per subspace, if you would prefer either.

I left the pre-existing `Box low's precision lowered by casting to float32` warning alone. It fires on main too, and the float64 `high` it comes from is a live guard: a `OneOf` of 130 `MultiBinary` raises `Box high is out of bounds of the dtype range` rather than silently wrapping 129 to -127.

`pytest tests/spaces` is 1403 passed and the full sweep matches main exactly, 77 pre-existing failures before and after, all missing optional deps, with 119 more passing cases from the two shared `TESTING_SPACES` entries.

### DIFF
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -266,6 +266,22 @@ def _flatten_sequence(
         return tuple(flatten(space.feature_space, item) for item in x)
 
 
+def _oneof_flat_dtype(space: OneOf) -> np.dtype:
+    """Returns the dtype shared by ``flatten`` and ``flatten_space`` for a ``OneOf`` space.
+
+    The dtype is the promotion of the dtypes of the *flattened* subspaces rather than of
+    the subspaces themselves, as flattening can change the dtype of a space, for example,
+    :class:`gymnasium.spaces.Text` flattens to an ``int32`` :class:`gymnasium.spaces.Box`.
+
+    Args:
+        space: The ``OneOf`` space to compute the flattened dtype of
+
+    Returns:
+        The dtype of the ``Box`` that the space flattens to.
+    """
+    return np.result_type(*(flatten_space(subspace).dtype for subspace in space.spaces))
+
+
 @flatten.register(OneOf)
 def _flatten_oneof(space: OneOf, x: tuple[int, Any]) -> NDArray[Any]:
     idx, sample = x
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR @teddytennant, I've updated the PR with the changes for the dtype as a singledispatch which will preserve the performance testing locally.

PR #1680, opened 2026-08-30
Fix AsyncVectorEnv final transition snapshots

### LINKED ISSUE #1132: [Bug Report] shallow copy of info and observation after termination
### Describe the bug

when using async_vector_env, the info after termination of one episode is wrong, it returns the info from reset() instead of step()

### Code example

```shell
in gymnasium/vector/async_vector_env.py 642:646
if terminated or truncated:
                    old_observation, old_info = observation, info
                    info["final_observation"] = old_observation
                    info["final_info"] = old_info

I think there should be a deepcopy for the old_observation and old_info here.
```

### System info

gymnasium==0.29.1

### Additional context

_No response_

### Checklist

Alphaxiaoteng:
# Description

Fixes #1132.

`AsyncVectorEnv` in `SAME_STEP` autoreset mode currently calls `reset()` before the terminal observation and info are sent to the parent process. Environments that reuse their observation or info buffers can therefore mutate the terminal values into reset values before users receive them.

This snapshots the terminal observation and info before reset using the same `ForkingPickler` serializer that `multiprocessing.Connection.send` already requires. That keeps the existing Async IPC compatibility boundary (including values that are pickleable but intentionally not deepcopyable) while preserving the actual final transition in both shared-memory modes.

## Type of change

## Testing

- With only the regression test applied to unmodified `main`, both Async modes failed because `final_obs` was `[0]` instead of the terminal `[1]`.
- Focused regression: 2 passed (shared memory enabled and disabled).
- Complete `tests/vector/test_autoreset_mode.py`: 76 passed.
- Complete `tests/vector`: 1524 passed, 133 skipped, no failures.
- All applicable changed-file pre-commit hooks passed, including Ruff check/format, codespell, AST validation, private-key detection, and `ty`.
- `git diff --check` passed.

## AI assistance

This PR was implemented with AI assistance and independently reviewed and tested against the reported failure path.

# Checklist:

- [ ] I have run `pre-commit run --all-files` (all applicable hooks were run on the changed files)
- [x] Documentation changes are not required because this restores existing final-transition semantics without changing the public API
- [x] New and existing vector unit tests pass locally

### DIFF
--- a/gymnasium/vector/async_vector_env.py
+++ b/gymnasium/vector/async_vector_env.py
@@ -11,6 +11,7 @@ from copy import deepcopy
 from enum import Enum
 from multiprocessing import Queue
 from multiprocessing.connection import Connection
+from multiprocessing.reduction import ForkingPickler
 from multiprocessing.sharedctypes import SynchronizedArray
 from typing import Any, TypeAlias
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Thanks for the PR, reviewing the issue, I think the problem is actually with environments that use the same object for sequential observations / info. Enforcing this in `check_env` will be the easiest solution rather than this. I'll make a PR to enforce this

PR #1685, opened 2026-09-04
Drop Python 3.10 support

### LINKED ISSUE #1684: Drop Python 3.10 Support At Next Release
Python 3.10 will be EOL in less than a month at the time of writing this issue. Support should be dropped in the next release cut.

teddytennant:
# Description

Python 3.10 reaches end of life this October, so this raises the floor to 3.11 ahead of the next release, as requested in #1684.

Changes: `requires-python` and the classifiers in `pyproject.toml`, the `ty` target version, the CI test matrix and the `build-necessary` job (which now builds on 3.11).

With 3.11 as the minimum, ruff's pyupgrade rule rewrites the `TYPE_CHECKING`-only imports of `Self` and `Never` to come from `typing` instead of `typing_extensions`. `Unpack` and `TypeVar` still come from `typing_extensions` since the forms used here need 3.12 and 3.13.

Rebased on main. The README edits are gone: a25de67c removed that whole line upstream, so there is nothing left to update there.

`ruff check .` and `ruff format --check .` are clean on the rebased tree (262 files). `tests/test_core.py` passed on a 3.11 install before the rebase; the rebase only dropped the README hunks.

Fixes #1684

### DIFF
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
         numpy-version: ['>=1.21,<2.0', '>=2.1']
         exclude:
           - python-version: '3.13'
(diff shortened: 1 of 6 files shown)


jkterry1:
Readme should drop mention of python versions entirely since it has the widget showing supported versions

pseudo-rnd-thoughts:
@jkterry1 The widget is from pypi, it will update when a new release is cut

pseudo-rnd-thoughts:
@teddytennant I'll merge this after the next release is cut this week

teddytennant:
dropped the versions line, the badge covers it.

PR #1688, opened 2026-09-06
Keep Discrete support when DtypeObservation changes dtype

teddytennant:
DtypeObservation builds a Box from Discrete(n, start=s) with high=s+n. Discrete is {s, ..., s+n-1}, so the extra value is both contained and sampled for integer dtypes. MultiDiscrete dropped a non-zero start when the dtype changed.

Repro:

    from gymnasium.wrappers import DtypeObservation

    env = DtypeObservation(GenericTestEnv(observation_space=Discrete(2)), dtype="int64")
    env.observation_space.contains(2)  # True, should be False

Clamped Discrete high to s+n-1 and passed start through on MultiDiscrete.

PYTHONPATH=. python -m pytest tests/wrappers/test_dtype_observation.py tests/wrappers/vector/test_vector_wrappers.py -q -k "dtype or Dtype"

### DIFF
--- a/gymnasium/wrappers/transform_observation.py
+++ b/gymnasium/wrappers/transform_observation.py
@@ -607,13 +607,15 @@ class DtypeObservation(
         elif isinstance(env.observation_space, spaces.Discrete):
             new_observation_space = spaces.Box(
                 low=env.observation_space.start,
-                high=env.observation_space.start + env.observation_space.n,
+                high=env.observation_space.start + env.observation_space.n - 1,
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR @teddytennant, looks good. Could you just run `pre-commit run --all-files` please

teddytennant:
done, ruff-format just wanted a line split in the new test. pushed.

PR #1691, opened 2026-09-10
Flatten n-d Box observations in DiscretizeObservation

teddytennant:
DiscretizeObservation takes n_dims from shape[0], so a finite 2-D Box constructs a Discrete space then crashes digitize on reset with ValueError: object too deep for desired array. The class says it discretizes any finite Box.

Repro:

    from gymnasium.wrappers import DiscretizeObservation

    env = DiscretizeObservation(
        GenericTestEnv(observation_space=Box(0, 1, shape=(2, 2))), bins=2
    )
    env.reset()  # ValueError

Raveled low, high, and the observation so n_dims is the Box size.

PYTHONPATH=. python -m pytest tests/wrappers/test_discretize_observation.py -q

### DIFF
--- a/gymnasium/wrappers/transform_observation.py
+++ b/gymnasium/wrappers/transform_observation.py
@@ -829,9 +829,9 @@ class DiscretizeObservation(
                 "DiscretizeObservation is only compatible with Box continuous observations."
             )
 
-        self.low = env.observation_space.low
-        self.high = env.observation_space.high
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, this should be good. My only worry is `np.ravel` vs `np.flatten` but `ravel` should be good in this case and `np.clip` will return a new array I believe

PR #1692, opened 2026-09-10
Fix MaxAndSkipObservation pooling on early episode endings

Doribelove:
# Description

`MaxAndSkipObservation` returns zero or stale observations if the underlying episode terminates or truncates before the last skipped step. For example, with `skip=4`, an environment returning `[t, 10-t]` and ending at step 2 produces `[0, 0]` instead of the elementwise maximum `[2, 9]`. An early ending after a previous complete call can return that previous call's pooled observation.

The buffer was only updated at `skip-2` and `skip-1`. This change saves each observation in a two-slot circular buffer and initializes both slots on the first internal step. It pools the last two observations actually received during the current call; if only one step occurs, it returns that observation. Reward accumulation, ending flags and the final info remain unchanged.

The added tests cover every ending position for skips 2, 3 and 4, both ending flags, a previous complete call, and reset followed by a one-step ending. They use negative observations with crossing component values and an environment that reuses one ndarray, and check that an earlier returned observation is not mutated. The class documentation now specifies the early-ending behavior.

Each internal step now writes an observation into the buffer, with both slots initialized on the first step. This entails more frame copies than recording only the last two scheduled steps; no performance benchmark or training evaluation was performed.

## Type of change

- [x] Bug fix
- [x] Documentation update

## Validation

Base: `ec4715dda7cda0b36e2cd80af4c6bbff7e8a4fee`.

- Original implementation with the new tests: **16 failed, 8 passed** in `tests/wrappers/test_max_and_skip_observation.py`; failures were observation-value assertions. After the fix: **24 passed**.
- Wrapper/core regression group: **195 passed**, compared with **173 passed** on the unmodified base; both runs report the same eight warnings. This group includes the affected test file, frame stacking, delay, time awareness, observation normalization, action repetition, autoreset, time limits and `tests/test_core.py`.
- The repository's minimal-dependency CI test selection (`tests/test_core.py tests/envs/test_envs.py tests/spaces`): **1662 passed**, with 109 warnings. This overlaps the group above in `tests/test_core.py`; the counts are not additive.
- Doctests in `gymnasium/wrappers/stateful_observation.py`: **5 passed**, with three dependency warnings.
- `pre-commit run --all-files`: all applicable hooks passed, including Ruff and ty.
- Source distribution and wheel builds, and `git diff --check`: passed.

Local environment: Ubuntu 22.04.5 x86_64, Python 3.12.14, NumPy 2.5.3 and pytest 9.1.1, with `testing`, `classic-control`, `toy-text` and `box2d` extras. This does not constitute the full optional-dependency or multi-Python/NumPy CI matrix.

# Checklist

- [x] I have run `pre-commit run --all-files`.
- [x] I have commented the non-obvious buffer initialization.
- [x] I have updated the documentation.
- [x] My changes generate no new warnings in the tested regression group.
- [x] I have added tests that prove the fix is effective.
- [x] New and existing unit tests pass locally in the reported scope.

AI assistance: the implementation, tests and this description were prepared with Codex. A separate agent reviewed the change, and the reported validation commands were executed locally.

### DIFF
--- a/gymnasium/wrappers/stateful_observation.py
+++ b/gymnasium/wrappers/stateful_observation.py
@@ -568,6 +568,10 @@ class MaxAndSkipObservation(
 ):
     """Skips the N-th frame (observation) and return the max values between the two last observations.
 
+    If the episode terminates or is truncated before ``skip`` steps, the maximum
+    is taken over the last two observations received during that call. If only
+    one step is taken, that step's observation is returned unchanged.
+
     No vector version of the wrapper exists.
 
     Note:
@@ -638,9 +642,9 @@ class MaxAndSkipObservation(
         info = {}
         for i in range(self._skip):
             obs, reward, terminated, truncated, info = self.env.step(action)
-            if i == self._skip - 2:
-                self._obs_buffer[0] = obs
-            if i == self._skip - 1:
+            self._obs_buffer[i % 2] = obs
+            if i == 0:
+                # A one-step episode ending must not include a previous frame.
                 self._obs_buffer[1] = obs
             total_reward += float(reward)
             if terminated or truncated:
--- a/tests/wrappers/test_max_and_skip_observation.py
+++ b/tests/wrappers/test_max_and_skip_observation.py
@@ -2,6 +2,7 @@
 
 import re
 
+import numpy as np
 import pytest
 
 import gymnasium as gym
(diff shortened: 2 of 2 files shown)


Doribelove:
I rechecked the early-ending regression against current `main` (`694e912`) and this PR's head (`9a65b0f`) using the same 24-case test file. Current `main` has 16 failing result assertions and 8 passes; the PR head passes all 24. The failures include one-step endings returning zeros and later calls reusing frames from a previous call. The PR's GitHub test/build checks are green, and GitHub reports it as mergeable.

Could a maintainer review the intended early-ending pooling behavior when convenient? I can adjust the implementation or tests if the expected semantics differ.

PR #1693, opened 2026-09-10
Support empty stacked Sequence operations

Doribelove:
# Description

Zero-length sequences are accepted by `Sequence`'s length masks, but `Sequence(..., stack=True).sample(mask=(0, None))` raises `ValueError: need at least one array to stack`. Flattening or unflattening a valid empty stacked sequence fails the same way. For a `Box(shape=(2,))` feature, these operations should produce an empty array of shape `(0, 2)`.

All three paths already allocate a correctly shaped output and pass it to `concatenate`. This change lets the base NumPy handler return that output when the input iterable is empty and the output shape is exactly `(0, *space.shape)`. Other cases retain the existing `np.stack` path, including incorrect empty-output shapes. The provided array's identity and dtype are preserved, and existing Tuple/Dict recursion handles nested features. The public utility docstring describes this behavior.

The regressions cover integer and array zero-length requests through `mask` and `probability`, independent flatten/unflatten calls, scalar and multidimensional feature spaces, discrete dtypes, nested containers, empty iterators, mismatched output shapes, and nonempty zero-width samples.

## Type of change

- [x] Bug fix
- [x] Documentation update

## Validation

Base: `ec4715dda7cda0b36e2cd80af4c6bbff7e8a4fee`.

- Original implementation with the new tests: **36 failed / 4 passed** in `tests/spaces/test_sequence.py`; **12 failed / 1197 passed** in the vector utility file. The failures are empty-input `np.stack` errors.
- After the fix: **1249 passed** across those two files.
- `pytest tests/spaces tests/vector tests/test_core.py -q`: **3455 passed, 133 skipped**, compared with **3401 passed, 133 skipped** on the unmodified base. The skips explicitly exclude dynamic-shape spaces from vector/shared-memory tests. Both versions emit the same 719 warnings; warning blocks match after normalizing filesystem paths and process IDs.
- Doctests in `spaces/sequence.py`, `spaces/utils.py` and `vector/utils/space_utils.py`: **9 passed**.
- `pre-commit run --all-files`, source distribution and wheel builds, and `git diff --check`: passed.

Local environment: Ubuntu 22.04.5 x86_64, Python 3.12.14, NumPy 2.5.3 and pytest 9.1.1, with testing/classic-control/toy-text/box2d extras. The full optional-dependency and multi-Python/NumPy CI matrix was not run. No training or performance result is claimed. Existing Graph/Sequence-feature unflatten limitations are outside this patch.

# Checklist

- [x] I have run `pre-commit run --all-files`.
- [x] The empty-output condition is explicit in the code.
- [x] I have updated the relevant API documentation.
- [x] My changes generate no new warnings in the tested scope.
- [x] I have added tests that prove the fix is effective.
- [x] New and existing unit tests pass locally in the reported scope.

AI assistance: the implementation, tests and this description were prepared with Codex. Another agent independently reviewed the final change. The reported validation commands were executed locally.

### DIFF
--- a/gymnasium/vector/utils/space_utils.py
+++ b/gymnasium/vector/utils/space_utils.py
@@ -392,6 +392,9 @@ def concatenate(
 ) -> tuple[Any, ...] | dict[str, Any] | np.ndarray:
     """Concatenate multiple samples from space into a single object.
 
+    For NumPy-based spaces, empty inputs are supported when ``out`` has shape
+    ``(0, *space.shape)``. The empty output array is returned unchanged.
+
     Args:
         space: Space of each item (e.g. `single_action_space` from vectorized environment)
         items: Samples to be concatenated (e.g. all sample should be an element of the `space`).
@@ -427,7 +430,15 @@ def _concatenate_base(
     items: Iterable,
     out: np.ndarray,
 ) -> np.ndarray:
-    return np.stack(list(items), axis=0, out=out)
+    items = list(items)
+    if (
+        not items
+        and isinstance(out, np.ndarray)
+        and space.shape is not None
+        and out.shape == (0, *space.shape)
+    ):
+        return out
+    return np.stack(items, axis=0, out=out)
 
 
 @concatenate.register(Tuple)
--- a/tests/spaces/test_sequence.py
+++ b/tests/spaces/test_sequence.py
@@ -4,6 +4,8 @@ import numpy as np
 import pytest
 
 import gymnasium as gym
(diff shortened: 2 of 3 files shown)


Doribelove:
The current head passes the GitHub test and pre-commit matrix (the release-only `publish` job is skipped). The PR includes before/after regressions for empty stacked `Sequence` sampling, flattening, and unflattening. Could a maintainer review it when convenient? I can make any requested changes. Thank you.

PR #1694, opened 2026-09-10
Require matching Tuple lengths in space compatibility checks

Doribelove:
## Description

`is_space_dtype_shape_equiv` accepts a shorter Tuple when its components match the longer Tuple's prefix. Reversing the arguments instead raises `IndexError`:

```python
from gymnasium.spaces import Discrete, Tuple
from gymnasium.spaces.utils import is_space_dtype_shape_equiv

short = Tuple((Discrete(2),))
long = Tuple((Discrete(2), Discrete(3)))
is_space_dtype_shape_equiv(short, long)  # Before: True; after: False
is_space_dtype_shape_equiv(long, short)  # Before: IndexError; after: False
```

Require equal lengths before comparing components, following the existing `OneOf` handler. This also lets the existing custom-observation-mode checks reject incompatible Tuple lengths during vector environment construction. With a shorter declared Tuple, `AsyncVectorEnv(shared_memory=False)` could previously omit observation components from `reset` and `step` results.

The patch contains the length guard and one regression test covering both comparison directions plus distinct equal-length Tuples, as requested in review. No new dependencies or public API changes. No separate issue filed.

## Type of change

- [x] Bug fix

## Validation

Latest revision, on base `ec4715dda7cda0b36e2cd80af4c6bbff7e8a4fee`, with Python 3.12.14, NumPy 2.5.3 and pytest 9.1.1:

- The single regression fails against the original implementation with an incorrect `True` result.
- `pytest tests/spaces tests/vector tests/test_core.py -q -rs`: **3402 passed, 133 skipped**, with 719 warnings. The skips and warning count match the previously checked unmodified baseline.
- `pre-commit run --all-files` and `git diff --check`: passed.

Validation is local with the testing/classic-control/toy-text/box2d extras; the full optional-dependency and platform matrix was not run locally.

## Checklist

- [x] Ran all-files pre-commit checks.
- [x] Added a regression and ran the affected spaces/vector/core tests.
- [x] The correction introduces no public API or configuration changes requiring documentation.

AI assistance: the implementation, test, and PR text were prepared with Codex. Validation commands were executed locally.

### DIFF
--- a/gymnasium/spaces/utils.py
+++ b/gymnasium/spaces/utils.py
@@ -731,8 +731,13 @@ def _is_space_dict_dtype_shape_equiv(space_1: Dict, space_2):
 
 @is_space_dtype_shape_equiv.register(Tuple)
 def _is_space_tuple_dtype_shape_equiv(space_1, space_2):
-    return isinstance(space_2, Tuple) and all(
-        is_space_dtype_shape_equiv(space_1[i], space_2[i]) for i in range(len(space_1))
+    return (
+        isinstance(space_2, Tuple)
+        and len(space_1) == len(space_2)
+        and all(
+            is_space_dtype_shape_equiv(space_1[i], space_2[i])
+            for i in range(len(space_1))
+        )
     )
 
 
--- a/tests/spaces/test_utils.py
+++ b/tests/spaces/test_utils.py
@@ -228,6 +228,66 @@ def test_is_space_dtype_shape_equiv(space):
     assert is_space_dtype_shape_equiv(space, space) is True
 
 
+@pytest.mark.parametrize("reverse", [False, True])
+@pytest.mark.parametrize(
+    "space_1, space_2, expected",
+    [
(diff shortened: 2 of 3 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks for the PR, the general fix too good but ironically, there is too much testing needed for a minor bug like this in my opinion. 

Please simplify the tests to single one with your short and long examples plus an expected equal length

Doribelove:
Thanks for the review. I reduced the added tests to one test using the short and long Tuples from the description, checking both comparison directions and an equal-length case. The extra nested-space and vector-environment tests are removed; the length guard is unchanged.

The simplified test still fails against the original implementation. With the fix, the spaces/vector/core suite passes (3,402 passed, 133 skipped), and all-files pre-commit checks pass. I also updated the PR description to reflect the smaller patch.

pseudo-rnd-thoughts (approved):
Thanks for updating

PR #1695, opened 2026-09-10
Preserve integer ranges when batching Discrete spaces

Doribelove:
## Description

`batch_differing_spaces` can change the ranges of accepted Discrete inputs when signed and unsigned dtypes are mixed. For example:

```python
import numpy as np
from gymnasium.vector.utils import batch_differing_spaces

spaces = [Discrete(3, start=-5, dtype=np.int8),
          Discrete(3, start=200, dtype=np.uint8)]
batched = batch_differing_spaces(spaces)
print(batched.dtype, batched.start.tolist())
# Before: int8 [-5, -56]; after: int16 [-5, 200]
```

Reversing those inputs previously produced `uint8 [200, 251]`. A `uint8` count of 200 paired with `int8` could also wrap to a negative count and fail construction. Selecting the first dtype with the largest itemsize does not account for signedness.

Use `np.result_type` to select a common integer dtype, check it before conversion, and construct the count/start arrays with that dtype. Add 24 parameterized regressions for both input orders, range preservation, exact large uint64 starts, and unsupported promotion.

**Compatibility:** NumPy promotes uint64 mixed with any signed integer dtype to float64. Such combinations now raise an explicit `ValueError`, including small-valued inputs that previously happened to work: no native integer dtype covers both full input dtype ranges. The utility's docstring documents this requirement.

Scope is the public batching utility, whose existing tests already cover mixed signed widths. Sync/Async vector constructors continue to require matching observation dtypes; this does not implement the broader vector-environment support discussed in [#1467](https://github.com/Farama-Foundation/Gymnasium/pull/1467#discussion_r2452822528). No separate issue filed or new dependency added.

## Type of change

- [x] Bug fix, with the rejection behavior described above.

## Validation

Base `ec4715dda7cda0b36e2cd80af4c6bbff7e8a4fee`; Ubuntu 22.04.5, Python 3.12.14, NumPy 2.5.3, pytest 9.1.1:

- Final tests against the original implementation: **18 failed, 1197 passed** in `tests/vector/utils/test_space_utils.py`.
- With the fix: **1215 passed** in that file.
- `pytest tests/spaces tests/vector tests/test_core.py -q -rs`: **3425 passed, 133 skipped**, versus **3401 passed, 133 skipped** on the unmodified baseline. Both runs have the same 719 warnings after path/PID normalization; skips are existing dynamic-shape exclusions.
- Vector space-utils doctests: **5 passed**.
- Separate NumPy 1.26.4 environment: all **28 Discrete batching tests passed**.
- Independent actual-source audit: **128 dtype-matrix cases** and **12 composite/RNG cases** passed, including expected rejection cases.
- All-files pre-commit, commit hooks, `git diff --check`, and sdist/wheel builds passed.

Local tests used the testing/classic-control/toy-text/box2d extras. The full optional-dependency and OS/Python/NumPy CI matrix was not run; no training or performance results are claimed.

## Checklist

- [x] Ran `pre-commit run --all-files`.
- [x] Documented the common integer dtype requirement.
- [x] Added regression tests and ran existing tests in the affected scope.
- [x] No new warnings in the tested scope compared with the baseline.

AI assistance: the implementation, tests, and this description were prepared with Codex; a separate Codex agent independently reviewed the change and ran the additional audit. This is not a claim of human or maintainer review.

### DIFF
--- a/gymnasium/vector/utils/space_utils.py
+++ b/gymnasium/vector/utils/space_utils.py
@@ -161,6 +161,10 @@ def batch_differing_spaces(spaces: _PySequence[Space]) -> Space:
     Returns:
         A batched space
 
+    For Discrete spaces, dtypes are promoted to a common integer dtype that can
+    represent the full range of each input dtype. Combinations without such a
+    dtype, including uint64 mixed with a signed integer dtype, raise ValueError.
+
     Example:
         >>> from gymnasium.spaces import Discrete
         >>> spaces = [Discrete(3), Discrete(5), Discrete(4), Discrete(8)]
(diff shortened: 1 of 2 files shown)


Doribelove:
The three workflows for this PR have not started yet; GitHub shows them as awaiting maintainer approval, including the [PyTest run](https://github.com/Farama-Foundation/Gymnasium/actions/runs/34469946763). Could a maintainer approve the pending workflows when convenient so the full matrix can run?

The affected space-utils file passed 1,215 tests locally on Python 3.12 / NumPy 2.5.3, and the 28 Discrete batching cases passed with NumPy 1.26.4. I can follow up on any CI failures once the runs complete.

PR #1697, opened 2026-09-11
Refactor CartPole observation construction into _get_obs

smokingelephants:
# Description

This change refactors observation construction in `CartPoleEnv` into a `_get_obs()` helper method.

Previously, both `step()` and `reset()` constructed the observation directly from `self.state`. They now call `_get_obs()` instead. The default observation returned by `CartPoleEnv` remains unchanged.

The motivation is to separate the environment's internal state from observation construction and provide a simple extension point for subclasses that need to customize observations without overriding both `step()` and `reset()`.

This pattern is also consistent with observation construction used in other Gymnasium environments.

No additional dependencies are required.

Fixes # (issue)

## Type of change

* [ ] Documentation only change (no code changed)
* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] This change requires a documentation update

### Screenshots

Not applicable.

# Checklist:

* [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
* [ ] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation
* [ ] My changes generate no new warnings
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] New and existing unit tests pass locally with my changes

Full pytest collection was attempted locally but was blocked by missing optional dependencies including MuJoCo, Box2D, SciPy, pygame, and moviepy.

### DIFF
--- a/gymnasium/envs/classic_control/cartpole.py
+++ b/gymnasium/envs/classic_control/cartpole.py
@@ -161,6 +161,11 @@ class CartPoleEnv(gym.Env[np.ndarray, int | np.ndarray]):
 
         self.steps_beyond_terminated = None
 
+    def _get_obs(self):
+        """Return the current environment observation."""
+        assert self.state is not None
+        return np.array(self.state, dtype=np.float32)
+
     def step(self, action):
         assert self.action_space.contains(action), (
             f"{action!r} ({type(action)}) invalid"
@@ -223,7 +228,7 @@ class CartPoleEnv(gym.Env[np.ndarray, int | np.ndarray]):
             self.render()
 
         # truncation=False as the time limit is handled by the `TimeLimit` wrapper added during `make`
-        return np.array(self.state, dtype=np.float32), reward, terminated, False, {}
+        return self._get_obs(), reward, terminated, False, {}
 
     def reset(
         self,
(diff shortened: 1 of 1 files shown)


smokingelephants (commented):
The major reason I want to have this, is to be able to sub class this environment the same way I can sub class other environments and override their _get_obs() function.

I noticed that other environments have already been ported to use the _get_obs() way of getting state, but cartpole did not have that yet. I assume it's an oversight, unless it's their intentionally. 

Thanks for checking this. :)

pseudo-rnd-thoughts:
Thanks for your PR @smokingelephants, I think the changes both make sense and I suspect are the wrong solution for you. Why aren't you using a wrapper or why can't you override the function currently anyway?

smokingelephants:
@pseudo-rnd-thoughts That is true. I can override it myself with like a wrapper. Just to add. I absolutely love this library. :)

I intended to teach students about overriding functions in this library, so I picked up cartpole. But I noticed the difference compared to other environments. 

So I just wanted it to be standardized. 

Stable baselines has evolved quite a lot, and it has standardized many things in the code architecture for RL (although imo at the cost of teachability). I found that the cartpole problem not following the envs like the rest. 

Q. Won't it create issues down the line for vectorized environments or further future architectural changes?

PR #1703, opened 2026-09-14
Flatten n-d Box actions in DiscretizeAction

teddytennant:
DiscretizeAction takes n_dims from shape[0], so a finite 2-D Box becomes Discrete(4) instead of Discrete(16). A scalar Box raises IndexError. DiscretizeObservation already ravels this (#1691).

Repro:

    from gymnasium.wrappers import DiscretizeAction

    env = DiscretizeAction(
        GenericTestEnv(action_space=Box(0, 1, shape=(2, 2))), bins=2
    )
    env.action_space  # Discrete(4), should be Discrete(16)

Raveled low and high so n_dims is the Box size, then reshape the reconstructed action back.

PYTHONPATH=. python -m pytest tests/wrappers/test_discretize_action.py -q


pseudo-rnd-thoughts (approved):
Thanks for the PR again @teddytennant, LGTM

PR #1705, opened 2026-09-14
Compare the current RNG state in check_step_determinism

NaCr05:
# Description

check_step_determinism saves the RNG after the first transition, but compares that old object with its own saved copy after the second seeded reset and step. Env.reset(seed=...) can replace the generator, so differences in the current RNG state are missed when the first-step outputs happen to match.

Read the environment's current generator after the second transition and compare its state with the first snapshot. Tests exercise identical outputs with different RNG consumption, a deterministic control, and the existing EnvSpec.nondeterministic opt-out. The fixture passes seeded-reset validation and returns fresh observations and info dictionaries.

## Type of change

- [x] Bug fix

## Validation

Tested locally on Windows with Python 3.11.15.

- NumPy 2.4.6: 88 tests passed across the active and passive environment-checker modules.
- NumPy 1.26.4: the same 88 tests passed.
- The regression with differing RNG consumption failed against unmodified main ec4715dda7cd because no AssertionError was raised.
- pre-commit run --all-files: passed, including Ruff and ty.

Commands:

```text
pytest tests/utils/test_env_checker.py tests/utils/test_passive_env_checker.py -q
pre-commit run --all-files
```

Existing test-collection deprecation warnings were observed. No full repository suite or upstream CI has been run.

# Checklist

- [x] Ran pre-commit checks over all files.
- [x] Added a regression that detects the missed RNG-state mismatch.
- [x] Relevant new and existing tests listed above pass locally.
- [x] Retained the declared-nondeterministic opt-out and deterministic control.


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #1708, opened 2026-09-14
Fix check_environments_match comparing env_b's observation space to itself

### LINKED ISSUE #1707: `check_environments_match` never compares observation spaces
## Description

`gymnasium.utils.env_match.check_environments_match` is intended to verify that
two environments are identical, including their observation spaces. However, the
observation-space assertion compares `env_b` to itself:

```python
assert skip_obs or env_b.observation_space == env_b.observation_space
```

This is a tautology, so any two environments pass the observation-space check
regardless of how different their `observation_space`s are — as long as the
sampled observations happen to be equal (or `num_steps` comparisons don't
diverge). Environments with different `Box` bounds, different shapes, or even
different space types are silently accepted.

## Reproduction

```python
import numpy as np
import gymnasium as gym
from gymnasium import spaces
from gymnasium.utils.env_match import check_environments_match

class Env(gym.Env):
    def __init__(self, obs_space):
        self.observation_space = obs_space
        self.action_space = spaces.Discrete(2)

        return np.zeros(2, dtype=np.float32), {}

    def step(self, action):
        return np.zeros(2, dtype=np.float32), 0.0, False, False, {}

env_a = Env(spaces.Box(0, 1, (2,), dtype=np.float32))
env_b = Env(spaces.Box(0, 2, (2,), dtype=np.float32))

# Expected: AssertionError (observation spaces differ)
# Actual: passes silently
check_environments_match(env_a, env_b, num_steps=0)
```

## Additional issues in the same function

- Two assertion messages are plain strings containing a literal `{step}`
  placeholder (missing the `f` prefix), so failures report `step = {step}`
  instead of the step number.
- The docstrings for `skip_rew`, `skip_terminal`, `skip_truncated`, and
  `skip_info` all incorrectly say they check "equivalence of the observation".

## Environment

- Gymnasium 1.4.0 (main @ ec4715dda7cda0b36e2cd80af4c6bbff7e8a4fee)

betacatsling:
# Description

`check_environments_match` asserted
`skip_obs or env_b.observation_space == env_b.observation_space`, a tautology
that never compares the two environments' observation spaces. Any pair of
environments — even with different `Box` bounds, shapes, or space types —
passed the observation-space check. The assertion now compares
`env_a.observation_space` to `env_b.observation_space`.

The same change also fixes two assertion messages that were missing the `f`
prefix (failures reported the literal text `step = {step}`), and corrects the
`skip_rew` / `skip_terminal` / `skip_truncated` / `skip_info` docstrings, which
all incorrectly described checking "equivalence of the observation".

Fixes #1707

## Type of change

# Checklist:

- [x] I have run the `pre-commit` checks on the changed files
- [x] My changes follow the repository docstring conventions

## Validation

New focused tests in `tests/utils/test_env_match.py` cover:

- matching environments pass
- mismatched observation spaces raise `AssertionError`
- `skip_obs=True` skips the observation-space comparison
- mismatched action spaces raise `AssertionError`

Commands and results:

- `pytest tests/utils/test_env_match.py -q`
  - before: `1 failed, 3 passed` — `test_mismatched_observation_space` did not
    raise because `env_b.observation_space == env_b.observation_space` is
    always true
  - after: `4 passed in 0.16s`
- `pre-commit run --files gymnasium/utils/env_match.py tests/utils/test_env_match.py`
  - all hooks pass (ruff check, ruff format, ty, pydocstyle, etc.)
- Broader suite run: `1696 passed, 1 skipped, 2 failed`; the 2 failures are
  pre-existing `tests/utils/test_env_checker.py` rendering failures caused by
  `pygame` not being installed in this environment, unrelated to this change.


pseudo-rnd-thoughts (approved):
Thanks for spotting those issues

PR #1713, opened 2026-09-15
Use the new Farama pixel art for Cliff Walking

### LINKED ISSUE #1709: Use New Pixel Art for Cliff Walking
We're in the process of reskinning all of Farama's pixel art environments across projects to a single professional format given the amount they're used in schools. If someone could throw up a PR using this initial set for toy text cliff walking I'd be highly grateful

[cliff_walking.zip](https://github.com/user-attachments/files/32199619/cliff_walking.zip)

MannXo:
# Description

Reskins Cliff Walking with the new pixel art tile set attached to the issue.

The new set is structured differently from the old one, so this is not a file swap. It has a single agent sprite instead of four directional ones, one ground tile instead of a two-tile checkerboard, no near-cliff tile, and a cliff row capped at both ends with alternating middle tiles. Start and goal are letter glyphs now, with a flag on the goal.

I reconstructed the intended layout by matching every cell of the `master.png` reference against the individual sprites rather than eyeballing it. All four overlay positions resolve at zero pixel residual, so the placement below is measured:

| element | placement | offset in a 64px cell |
| ------- | --------- | --------------------- |
| `tile_ground` | every non-cliff cell | full cell |
| `tile_cliff_left` | first cliff column | full cell |
| `tile_cliff_mid_a` / `_mid_b` | between the caps, alternating by column parity | full cell |
| `tile_cliff_right` | last cliff column | full cell |
| `letter_s` / `letter_g` | start / goal cell | (8, 8) |
| `goal_flag` | goal cell | (28, 16) |
| `agent` | player cell | (20, 20) |

`cell_size` goes from 60 to 64 so the art renders at its authored scale with no resampling. A 60px cell would need a 0.9375 scale factor and would break the pixel grid. The visible consequence is that `rgb_array` output goes from 720x240 to 768x256.

`gymnasium/envs/tabular/cliffwalking.py` renders the same environment and loaded the same assets from `../toy_text/img/`, so it is updated alongside. Leaving it behind would have meant two different art styles for one environment, and would have forced keeping the old assets alive purely to feed it. Happy to split that into its own PR if you would rather keep this to toy text.

`cookie.png` and the five `mountain_*.png` are now unused and removed. `elf_*.png` and `stool.png` stay, because `frozen_lake.py` uses them.

Two things I could not decide on my own:

1. The credit comment at the bottom of `cliffwalking.py` read `# All other assets by ____`, with the blank already there. I left it as `# Pixel art assets by ____` rather than invent an attribution. Tell me what it should say and I will fill it in.
2. `docs/_static/videos/toy_text/cliff_walking.gif` is regenerated with the repo's own `gen_gifs.py` parameters (300 frames, 50ms, random actions). Say the word if you would rather regenerate all the env gifs in one pass separately.

Fixes #1709

## Type of change

I left "breaking change" unchecked because the observation space, action space, reward and dynamics are all untouched. The one thing that does change for an existing user is the size of the `rgb_array` frame, 720x240 to 768x256. Move the tick if you count that as breaking.

### Screenshots

| Before | After Cliffwalking | After Tabular Clifwalking |
| ------ | ----- | ----- |
| image | image| image|

The `tabular/CliffWalking-v0` render is pixel-identical to the toy text one above.

# Checklist:

On the two unchecked boxes. I added no tests because the repo has no render-geometry tests for the toy text envs and adding the first one felt like scope creep on an art change. Instead I verified the render against the reference directly: the rendered frame is pixel-identical to `master.png` except for 3984 pixels that are all on the outermost 4px of the right and bottom edges and all one colour, which is the decorative outer stroke the mockup drew and for which no sprite was shipped.

Test results, run against `main` and this branch in the same environment for comparison:

```
main:   105 failed, 4180 passed, 147 skipped, 7 errors in 48.74s
branch: 105 failed, 4180 passed, 147 skipped, 7 errors in 48.65s
```

Identical. The 105 failures and 7 collection errors are pre-existing and all come from optional dependencies I do not have installed (Box2D, MuJoCo, torch, flax, moviepy). None mentions cliff walking. The 9 `CliffWalking` tests that do run pass on both; the 3 `tabular/CliffWalking-v0` failures are pre-existing and also fail on `main`.


pseudo-rnd-thoughts:
@jkterry1 Should we consider incrementing the version number for rendering changes?

jkterry1:
Historically at least we’ve only ever dined that when the rendering change impacts the observation

-------- Original Message --------

PR #1715, opened 2026-09-16
Preserve final observation and info in async SAME_STEP autoreset

### LINKED ISSUE #1132: [Bug Report] shallow copy of info and observation after termination

lindicaphxag-tech:
Fixes #1132\n\n### Summary\n\nWhen an environment reuses and mutates the same observation and info objects across step() and reset(), AsyncVectorEnv with AutoresetMode.SAME_STEP could expose reset values through final_obs and final_info.\n\nCopy the terminal observation and info before calling reset(), then use those snapshots for final_obs and final_info.\n\n### Tests\n\n- Added a regression test with a reusing environment.\n- Covers shared_memory=True and shared_memory=False.\n- Local test suite: 85 passed.\n- pre-commit checks: passed.\n\nAI assistance disclosure: This change was developed with AI assistance. The issue reproduction, implementation, tests, and local verification were reviewed by the contributor.


pseudo-rnd-thoughts:
Thanks for the PR but I'm closing this issue for the same reason https://github.com/Farama-Foundation/Gymnasium/pull/1680  The problem is actually with environments that use the same object for sequential observations / info.

